PR #7, opened 2017-06-13
Main pages are already translated

rizzopablo:
Hi, would you like to merge and publish the spanish translation? I'll finish remaining pages en a few days.
Regards

Pablo


r10s:
Thank you very much. I've just commited the request and I also will give you direct access to the repository, so that things get easier. 
Feel free still to created pull requests then, but you can also change things quicker if I'm not around.

PR #10, opened 2017-07-12
Portuguese features.md

luisfsr:
Translation of features.md into Portuguese.


r10s:
Added to `pt` - thank you very much :-) :+1:

PR #13, opened 2017-07-12
Portuguese translation to support.md

luisfsr:
As follows...


r10s:
Thanks, I add to to the `pt` directory in a few minutes.

PR #14, opened 2017-07-12
Portuguese translation of download.md

luisfsr:
As follows...


r10s:
Thanks. I add the translated file to the `pt` directory in a few minutes.

PR #19, opened 2017-07-22
Add initial italian language support

Claudinux:


r10s:
Grazie - thank you very much. I'll merge the pull request soon, I think the Changelog and delta-makes-chatting-better.md should not be translated and just be linked to the original pages.

PR #21, opened 2017-07-28
Changed defalut layout from DE to PT

luisfsr:


r10s:
Perfect - sorry, I forgot ...

PR #31, opened 2017-10-02
Add a new page by renaming an existing

haeckser:
I hope that works fine to add a new post.
Content is the hint to ANBOx for Linux users to use Delta Chat on Desktop


r10s:
Thank you very much - I've added you post by a [another commit](https://github.com/deltachat/deltachat-pages/commit/6ee94edb3d6f6cbba6208823d26e03086dcbeb5a) (renaming is no good idea as this would delete the older post)

PR #37, opened 2017-10-08
exchanged negatives, added autocrypt

testbird:


r10s:
Thank you very much :)

PR #45, opened 2017-10-21
Updated to follow the english version changes...

luisfsr:


Ampli-fier:
Oi! Desculpa aí!
I changed the german and the english layout of the featured.md file. I wanted to get closer to the slider content of the index.md files and to the slider inside the app. Unfortunately this did collide with your changes, it seems :-/

I might try to get help from a native speaker (english) to have a look at the feature.md that I just wrote. Or can you do that as well?

Muito obrigado!

luisfsr:
Yes, I can do it, but I'm not an English native speaker.
Personally, I think the text is good, I just wouldn't repeat the signal ":" twice in the same sentence and would use more connective words instead of the "-" signal.
One more thing: in the top side of the page we have a "why" and the first word in the body was also a "why". Now it's a "what" which is a bit strange. In my opinion we should maintain the use of "why" in the first sentence of the body...
Lastly, add me to your Delta Chat: luisfsr@gmai.com :)

Ampli-fier:
All valid points :)
I wanted to get rid of the question mark in the header line: "Why?" did not look appealing to me. So, as a 1st step I removed the "?" and kept the word in the page title (shown in browser tab). As a replacement for "Why" I came up with these ideas: "About" / "Features" / "Idea" / "Vision" / "Mission". I do not like to use "questions" in the header.

The _Contribute_ page also has a deviating sentence on top of the page: "How to support us". We could try to keep this all aligned.

In the bullet list I wanted to avoid dots to separate sentences, so I used "-". Actually the first bullet is not following the schema of all others:
- Feature name\: Sentence...

Should be changed as well ...

luisfsr:
I see the section "why" (if I'm not mistaken) as a deeper explanation of the "vision" and "features", with an even deeper explanation, at the bottom of the page (Delta makes chatting better), agretating one of the "missions" of DC - which would be, apart from easing the use of e-mail wiht cellphones, to bring real end-to-end encryption to everybody, thus making it possible to the average user to resist from bulk vigilance (right?). Maybe the word "features" would be the most comprehensive word, which would convene all of it. So I vote for "features". I agree with your observation about the section "contribute". I'd like to add that I see in several sites of chat apps an appeal to youthness and joy in chatting. Maybe that could be explored in the section "why / features"... What do you think?

Ampli-fier:
I fully agree.
Some weeks ago I created [this version](https://ampli-fier.github.io/deltachat-pages/en/features) with the _Header line_ as we both think makes the most sense (hopefully) :) Even the markdown file is called features.md.
In todays version of the "live" pages I tried to "sound straight by using short sentences and coming to the point" in the Features page. 
If you have an idea how to express the "joy in chatting", that would be great.

luisfsr:
Ideas come really easily, but when it comes into putting them in motion things get harder. I searched the web looking for two chat app sites (telegram and ICQ) and both have some similar approaches, but each one has its own essence. Telegram appeals to funny drawings filled with humor (e.g., a man sending a message by the bathroom, or the other one hiding his face with an anonymous mask, or the other using dynamite to destroy selfdestruct messages...); ICQ appeals to people talking about social occasions and so (which would be easier to recreate, apart from that bad tasting young ladies pics, which could be substituted by other images - like public domain pictures from old paintings or so)... I pretty much prefer the telegram approach, but for that it'd be necessary to hire a designer.
About the new proposed header, do you intend to change the texts by icons?

luisfsr:
P.S: Whatsapp site also appeals to social ocasions, they show a group conversation called "it's the weekend" in which someone sent a video of a child and people chattering around whit several emoticons and so... Well, those things pass the idea of joy. Maybe we could do something like that, but preserving the essence of Delta Chat.

Ampli-fier:
Björn is looking for a new design, see [this issue](https://github.com/deltachat/deltachat-android/issues/176).

So, based on the (possible) outcome, we could build newly designed pages.

I forgot to mention that the _Header_ I wanted to show you, was only visible when **not** using a smartphone.
The Icon-test was done to show Björn how it would look like on small devices (smartphones) - but he prefers to keep text version only.

So, at the moment the question is more how can we "pimp" the text.

r10s:
My idea is not to redesign the whole layout, pages or sth. like that, but only to replace the six graphics.

(however, of course, if there are issues with the current layout, we should fix it)

luisfsr:
I've visualized the desktop version of the headers and I liked.
About the pimping issue, maybe what could be done was to insert a cool image in the screenshot conversarion shown in DC. Maybe something related to beach, once the conversation is about it... Maybe change the person to person conversation and turn it into a group conversation. Other pimpings could be postponed to when Björn manages to hire the designer.

r10s:
Ahh, I see. Yes, the screenshot should definitely be updated. I just did not found the time up to now.

For anyone how's interested in experiments, here is the logo as a vector graphic (if it should be integrated to the screenshot again): [delta-v7-pathed.svg.zip](https://github.com/deltachat/deltachat-pages/files/1415306/delta-v7-pathed.svg.zip)

PR #48, opened 2017-11-12
Update help.md

haeckser:
Add installation via F-Droid store to migrate data instead of removing to linux on desktop


Ampli-fier:
Update help.md (en) - Use F-Droid in Anbox

PR #60, opened 2018-02-24
A few typo and consistency fixes

ujdhesa:


ujdhesa:
Here you are sq translations of the web pages. There are a few words (a couple of sentences in the changelog.md) untranslated and I’d be grateful if someone help a bit with them.

r10s:
thank you very much, i will add the language to the footer and to the automatic language selection in a subsequent commit. I think the standards.md and changes.md should not be translated - the latter to avoid new versions to be blocked and the former is just fine in english.

r10s:
@ujdhesa I've added an Albanian layout for the main menu and for the footer - can you please check/modify https://github.com/deltachat/deltachat-pages/blob/master/_layouts/default-sq.html

If you think that is okay, I will proudly add the new language to the footer and to the automatic selection.

Have a nice weekend :)

r10s:
NB: in between the pages are already available via https://delta.chat/sq/

ujdhesa:
Go ahead and push it in production. I’ll see it live and perhaps might need to translate the bits in the footer (Language and Copyright note).

Thank you!!!

r10s:
@ujdhesa okay, Albanian is live now :) the language should be selected based upon the browser language and can be selected manually in the footer (wondering what "sq" stands for).

One remark: Maybe the term "Imprint" in the footer and as the headline on the linked pages should be translated, too?

ujdhesa:
Ideally yes. But I haven’t seen a proper sq equivalent anywhere and I don not have any idea right now of what that could be in sq… That doesn’t influence the overall working of the site anyway… So I’m happy leaving it untranslated for now.

sq stands to Albanian as de stands to German. _Shqipëri_ is how albanians call their country, hence _shqip_ for the language => sq for a more consistent (IMO) locale name than al.

r10s:
okay  - and thanks for the explanation :)

Ampli-fier:
Hi folks, there is an "official" list of country codes on [Wikipedia](https://en.wikipedia.org/wiki/ISO_3166-1). According to this list, the abbreviation for Albania is **al**. Would you agree to change from sq to al?

ujdhesa:
sq is the correct code. Besides, it’s well established and largely used. So I don’t see why we should refrain using it.

r10s:
@Ampli-fier we're not using a _country_ code for the language subdirectory or for the list but the [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) _language_ code, which is `sq` for Albanian. So, everything is correct :)

Ampli-fier:
**Of course** - No idea why I mixed this up ... :-/

r10s:
@Ampli-fier All these things are complicated enough to not worry about mixing things up here and there :)

PR #62, opened 2018-03-16
FR translation of the mailing-list bullet item

kdltr:


r10s:
perfect, thank you very much. 

how did you recognise the change? i am wondering if we need any additional method to inform that the english page has changed and the translations should be checked.

kdltr:
I was on the IRC channel when the change got announced. It might be interesting to add a notification tool… but just watching the commits might just work well enough.

r10s:
IRC works :)
and: of course - watching the commits may be sufficient.

PR #66, opened 2018-04-01
Update help.md

Sebek05:


r10s:
ah, now i understand the bug you reported via mail some days ago :)

in fact, the part after "Yes" is not translated to any _language_. I think we should figure out the reason for this which would make this pr superfluous then. will have a look at this soon. 

apart from that, the _recommended_ way to export/import keys should be the Autocrypt Setup Message now, we should mention this somewhere (this feature was not available the help was checked the last time) ...

r10s:
i have update the source strings to add some Autocrypt hints, however, the _first_ string stays untranslatable - not only in polish, i've created an issue about this, see https://github.com/deltachat/deltachat-pages/issues/67

we can close this pr now as it won't fix the problem and as the sources already are updated.

feel free to translate the string in polish, i think sooner or later the bug will be fixed :)

PR #77, opened 2018-05-12
Update

luisfsr:


r10s:
@hi luisfsr thank you for the translation. i've merged the request and copied the translated strings to Transifex - otherwise the markdown files would be overwritten by the next adaption of the english source texts.

For further translations, it would save some work using Transifex for translation directly :)
here is the link: https://www.transifex.com/delta-chat/

however, great you push this forward :)

PR #85, opened 2018-05-24
Change imprint

r10s:
- change data to the one from merlinux.eu
- remove phone number (i think, this is not needed)
- add a line that delta chat is a community project
- use delta at merlinux dot eu as address (suggestion, should be created)


en/imprint.md
```diff
@@ -5,14 +5,18 @@ layout: default-en
...
+    Merlinux GmbH
+    Reichgrafen Str. 20
+    79102 Freiburg, Germany
+
+E-mail: {% include email-link %}
+
+Amtsgericht Freiburg: HRB709589  
+VAT: DE814082730  
+Managing Director: Holger Krekel  
+Prokura: Lene Wagner
```
hpk42: let's not list procura here.

hpk42:
other than the one comment looks good to merge

PR #89, opened 2018-06-26
attempt to sort the bullets into more easily comprehensible points

testbird:


testbird:
@hpk42 By going to "files changed" and using the button to "view" the whole file, the comparison with your initial version seems easiest. Tried to keep all points untouched as much as possible, just sorting and making the points clearer by having them begin with what actually relates to the listings.

You could move the "current status" to the bottom afterwards, to be even less in the middle.

PR #124, opened 2019-01-28
change in social media toolkit

ghost:
@r10s I changed the description. What do you think?


r10s:
i do not use mastodon no friendica currently, but, yes a ping on irc is fine :)

PR #125, opened 2019-01-28
make multi-line headlines look nicer

r10s:
the line-height of h1-elements spanning multiple-lines was a bit too small. and the spacing between characters was a bit too much, esp. in combination with multiple-lines.

this pr tries to mitigate both.


hpk42:
thanks for fixing this.

PR #131, opened 2019-02-08
Update help.md

adbenitez:
this seems to be wrong...

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -132,8 +132,8 @@ Other messages do not appear automatically. You can see the other messages in th
 
 The user's profile shows some additional information:
 
-- You can tap "QR Invite code" on Android and then use the "scan QR"
-  code button on another device to scan this code.  If both devices are online
+- You can tap "QR Invite code" on Android and then use the "scan QR code"
+  button on another device to scan this code.  If both devices are online
   they will introduce a chat channel with each other (if it doesn't exist already) 
   and also the encryption keys will be verified. 
 
--- a/en/help.md
+++ b/en/help.md
@@ -125,7 +125,7 @@ Other messages do not appear automatically. You can see the other messages in th
 
 - If a little **padlock** is shown beside a message, this implies that the message is end-to-end-encrypted _and_ is sent from the given sender _and_ your answer will be end-to-end-encrypted as well.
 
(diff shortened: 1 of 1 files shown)


en/help.md
```diff
@@ -125,15 +125,15 @@ Other messages do not appear automatically. You can see the other messages in th
 
 - If a little **padlock** is shown beside a message, this implies that the message is end-to-end-encrypted _and_ is sent from the given sender _and_ your answer will be end-to-end-encrypted as well.
 
-- If there is **no padlock**, the message is usually transport encrypted eg. because you or the sender have disabled end-to-end-encryption or the sender uses an app that does not support end-to-end-encryption.
```
r10s, line 128: even if there is no padlock, the message is usually transport encrypted; the original sentence is correct therefore.

r10s (changes requested):

en/help.md
adbenitez: but there are something wrong here, "eg. because you or the sender have disabled end-to-end-encryption or the sender uses an app that does not support end-to-end-encryption" that is not an explanation of why it is transported encrypted, sounds like the original intention was to say that it is unencrypted because "eg. because you or the sender have disabled end-to-end-encryption or the sender uses an app that does not support end-to-end-encryption"

adbenitez: anyway if you say is ok, you are free to edit it :)

adbenitez: or I can do it...

adbenitez: anyway I think that if the idea is to say that it is encrypted it should say something like "eg. because your server uses SSL to encrypt the connection"

adbenitez: but in the end, that is the standard, and this is not actually the mean of "secure" people expect, if the server uses SSL that doesn't mean the users at risk, etc. are save, I don't think it is fine to transmit a feeling of false security :-/

adbenitez: in this context we are talking about e2e encryption and if there are no padlock  the messages is **not** e2e encrypted

adbenitez:
first time I do this, I hope clicking the "resolve conversation" button means I accept the change requested...

adbenitez:
anyway I still think there is something wrong with this statement... :P

PR #135, opened 2019-02-13
add link to gplay beta to download page

### LINKED ISSUE #134: link directly to the Google Play page
@adbenitez pointed that out via irc.
@r10s it's still beta isn't it? should we wait with adding the link then?

r10s:
closes #134

### DIFF
--- a/en/download.md
+++ b/en/download.md
@@ -5,12 +5,14 @@ layout: default-en
 
 # Delta/Android Downloads (Beta)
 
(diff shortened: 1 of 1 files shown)


hpk42:
could there be a beta on the "get it on google play" link?

r10s:
> could there be a beta on the "get it on google play" link?

yip, i will change this.

PR #140, opened 2019-02-20
add blog entry wrt gplay and oauth2

r10s:


r10s:
@hpk42 updated the image and added info that yandex is on the way

PR #141, opened 2019-02-21
init ukrainian webpage

r10s:
translation is in progress, however, i think this can be merged if it is technically okay, also the other languages reach 100% only seldomly :)

preview: https://deploy-preview-141--deltachat.netlify.com/uk/help

the header/footer cannot be modified in transifex, add a commit to _layouts/default-uk.html or just drop a note here :)


hpk42:
isn't this translation using the old page?  It's not a translation of the current page, is it?

r10s:
>isn't this translation using the old page? It's not a translation of the current page, is it?

it's the translation of the current english page (state yesterday). 

things that are not yet translated are just english, however. eg. the homepage is not yet translated, but "contribute", "help" and others are.

PR #142, opened 2019-02-22
improve 'own server' wording

r10s:
the "has no own server" wording seems to be german, cmp https://news.ycombinator.com/item?id=19216827
(fun-fact is that the webserver says there is no server, however, i think the intention is clear to most readers) 

i do not have an idea about this. maybe someone with good english knowledge or a native speaker can check?

cc @Valodim


hpk42:
it's probably an improvement ;) if there are other ideas, new PRs are cheap ;)

PR #143, opened 2019-02-23
Point to the flatpak for linux desktop installs

flub:
This should work better for unsuspecting linux users


ralphtheninja (approved):

adbenitez:
this is really useful since the appimage don't work, people need yo install libssl manually, to a normal user is eyes 'the app just dont work' I actually recommended not to provide the appimage, or warn the user about the external dependency

r10s:
merge it for now, i think this is already a huge improvement and warnings or so can to to another pr.

adbenitez:
yup

PR #153, opened 2019-03-28
straighten main menu

r10s:
- move menu-entry that leaves the page to the most right for a better user-experience when browsing the menu from left to right
- equalize order in different locales


r10s:
merging this, we'll see how this works out :)

PR #154, opened 2019-03-28
index.md: hint and mention email account login

testbird:
Background:
"First contact" potential new users are often surprised and not prepared to provide their email account credentials for the initial setup. Simply explicitly mentioning their email account login in any information that is presented at first (and in the beginning) should make them aware.

Changes:
Improved context hints and and explicit mention of the required email account login (earlier): In the title and already as second bullet point, to make new user contacts already more aware of this in order to be prepared for the installation (account setup).

I think the header change may be discussed separately, the changes in the paragraphs are worth it on its own.

A specific use-case example is be the invitation of old messenger-contact (budies) that need to get a brief picture about what is involved by using deltachat, and visit the homepage. 
(Discussion at: https://support.delta.chat/t/lets-switch-to-a-new-messenger/334)

### DIFF
--- a/en/index.md
+++ b/en/index.md
@@ -3,7 +3,7 @@ title: The messenger
 layout: default-en
 ---
 
-# Head back to the future with Email-Chat!
+# Back to the future with Email-Chat!
 
 <img src="../assets/logos/delta-chat.svg" width="160" style="float: left; margin: 20px;" />
 
--- a/en/index.md
+++ b/en/index.md
@@ -3,7 +3,7 @@ title: The messenger
 layout: default-en
 ---
 
-# Back to the future with Email-Chat!
+# Back to the future, with Email-Chat!
 
 <img src="../assets/logos/delta-chat.svg" width="160" style="float: left; margin: 20px;" />
 
--- a/en/index.md
+++ b/en/index.md
@@ -8,10 +8,10 @@ layout: default-en
 <img src="../assets/logos/delta-chat.svg" width="160" style="float: left; margin: 20px;" />
 
 **Delta Chat is as easy-to-use as Telegram or Whatsapp, but without the tracking or central control.**
-Check out our [GDPR compliancy statement](gdpr).
+Check out our [GDPR data protection compliancy statement](gdpr).
 
 **Chat with anyone if you know their e-mail address, even if they are not (yet) using Delta Chat!** 
-All you need is a standard e-mail account login.
+All you need is your own standard e-mail account login.
 
 **Delta Chat doesn't make you use their own servers**, it uses the most massive and diverse, federated open messaging 
 system ever: the existing e-mail server network.
--- a/en/index.md
+++ b/en/index.md
@@ -3,18 +3,18 @@ title: The messenger
 layout: default-en
 ---
 
-# Chat over e-mail and head back to the future with us!
+# Head back to the future with Email-Chat!
 
 <img src="../assets/logos/delta-chat.svg" width="160" style="float: left; margin: 20px;" />
 
 **Delta Chat is like Telegram or Whatsapp but without the tracking or central control.**
 Check out our [GDPR compliancy statement](gdpr).
 
+**Chat with anyone if you know their e-mail address, no need for them to install DeltaChat!** 
+All you need is a standard e-mail account login.
+
 **Delta Chat doesn't have their own servers** but uses the most massive and diverse open messaging 
 system ever: the existing e-mail server network.
-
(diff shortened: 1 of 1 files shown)


r10s:
not sure which variant is better, my english is not good enough to decide this.

however, some notes:

- shifting the wording always has an impact for all translators which have to retranslatate/check their translations. this is okay and the normal process, however, a detail to keep in mind

- i personally would target more low hanging fruits as the help resp. faq where changes have a much larger impact

testbird:
Ok, here are some explanations about the edits to consider: 

Concerning the top header change from,

**Chat over e-mail and head back to the future with us!**
to
**(Head) Back to the future with Email-Chat!**

the shortening and  rearranging allowed to
* remove the slower to read/grasp "... over ... and ... to ... with ... whom?"  construction
* Make the known "back to the future" appear earlier
* mentioning the reminder (Email) together with the subject of the page (chat) on a single final emphasis
* let the final emphasis be a preparer/curiosity creator word for the following facts and details...

... which go on to place in a Delta Chat, Delta Chat, ... touting rhythm:
  * DC is an app
  * easy-to-use
  * no tracking
  * chat with any email account in the world
  * think of your own email account
  * comes without its own servers
  * massive etc. network

So which I want, this to having become a decently nice short presentation so much. ;-)

What about even further removing the first word of the title?

testbird:
Main point: The first read (landing on the webpage) now leaves behind the notion that the Delta Chat app is an awesome way to use an email account.

testbird:
I think the header change may be discussed separately, the changes in the paragraphs are worth it on its own.

en/index.md
```diff
@@ -3,18 +3,18 @@ title: The messenger
 layout: default-en
 ---
 
-# Chat over e-mail and head back to the future with us!
+# Back to the future, with Email-Chat!
```
hpk42, line 6: "chat over e-mail" is a term we often use, relates also to "chat over imap" in some ways.

en/index.md
```diff
@@ -3,18 +3,18 @@ title: The messenger
...
 
 <img src="../assets/logos/delta-chat.svg" width="160" style="float: left; margin: 20px;" />
 
-**Delta Chat is like Telegram or Whatsapp but without the tracking or central control.**
-Check out our [GDPR compliancy statement](gdpr).
+**The Delta Chat app is as easy-to-use as Telegram or Whatsapp, but without any tracking or central control.**
+Check out our [GDPR data protection compliancy statement](gdpr).
```
hpk42, line 11: "easy to use" is in the eye of the beholder ... "like" is a more objective description.

en/index.md
```diff
@@ -3,18 +3,18 @@ title: The messenger
...
 
-**Delta Chat doesn't have their own servers** but uses the most massive and diverse open messaging 
-system ever: the existing e-mail server network.
+**Chat with anyone if you know their e-mail address, even if they are not (yet) using Delta Chat!**
```
hpk42, line 13: we try to be very clear that you can use delta chat with folks who don't use it ... so not sure about the "(yet)"

en/index.md
```diff
@@ -3,18 +3,18 @@ title: The messenger
...
 
-**Delta Chat is like Telegram or Whatsapp but without the tracking or central control.**
-Check out our [GDPR compliancy statement](gdpr).
+**The Delta Chat app is as easy-to-use as Telegram or Whatsapp, but without any tracking or central control.**
+Check out our [GDPR data protection compliancy statement](gdpr).
 
-**Delta Chat doesn't have their own servers** but uses the most massive and diverse open messaging 
-system ever: the existing e-mail server network.
+**Chat with anyone if you know their e-mail address, even if they are not (yet) using Delta Chat!** 
+All you need is your own standard e-mail account login.
```
hpk42, line 14: "standard e-mail account login" is a strange series of terms -- not sure that it helps.
It's true that you can use your existing e-mail account, but you can also create a new one. 
the important point is that DC just needs a standard e-mail account.  The tricky UX challenge is how to -- in app -- make it clear that DC is, technically, an e-mail client.  Problem being that "e-mail client" is not a well-understood term in non-tech circles.

hpk42:
made some comments, but will also close this PR.  I'd prefer to discuss further in offline settings, there are some things that could be improved i think.  but overall there are a lot more interesting issues ... eg sorting out releases, downloads, and making some better in-website linking, reworking the FAQ -- compared to this, spending much time on discussing slight wording changes on the home page is less useful.

PR #157, opened 2019-03-31
Mini HowTo

testbird:
Provide some guidance by listing what's involved, incl. explicitly mentioning the required configuration of an email account.

### DIFF
--- a/en/download.md
+++ b/en/download.md
@@ -4,7 +4,7 @@ layout: default-en
 ---
 
 # Mini HowTo
-1) Select your desired version for your platform below.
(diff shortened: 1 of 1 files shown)


hpk42:
just discussing with @jikstra about revising the download page a bit.  We need better links especially for desktop releases.  I think that coming to the download page, everyone looks for the correct download link, then tries to install, then sees the first screen after install.  That's the canonical flow and i don't think there needs to be a mini-howto for that but maybe a visual re-arrangement.  Over the next weeks there are several release upcoming so the download page should be improved step by step with the blog posts announcing them.

PR #181, opened 2019-10-02
de-emphasize GDPR on home page and explain why we focus on GDPR

### LINKED ISSUE #96: GDPR does not make much sense outside Europe
The GDPR session of the site does not make sense if you live outside Europe and make the app feel something eurocentrist. Also, there are other laws out there that touches in the same object, which is personal data protection as, for instance, Brazilian Law of personal data protection, approved in the current year (2018).

hpk42:
fixes #96

### DIFF
--- a/en/gdpr.md
+++ b/en/gdpr.md
@@ -3,6 +3,12 @@ title: GDPR and Privacy
 layout: default-en
 ---
 
+# Note on privacy and GDPR 
+
+The General Data Protection Regulation (GDPR) is often regarded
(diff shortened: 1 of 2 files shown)


r10s:
nice!

ghost:
Hm - the heading hierarchy of the page is not optimized, if we want to comply to "best practices", we should only have one h1 tag at the top and order the rest of the page beneath that: http://www.webpagemistakes.ca/heading-structure-heading-tag/

But we can merge this for now, and I can try to optimize this later.

ghost (approved):

PR #182, opened 2019-10-03
link to providers page

trosel:


ghost:
Thanks for that change, good idea - but maybe we should wait until the page looks a bit better, and has less question marks?

trosel:
@lefherz I don't think the page will ever have less question marks. The existing forum link has even less data. At least this new one is honest about what it knows.

Issue #183, opened 2019-10-09
Translations on download Page are outdated


Simon-Laux:
https://delta.chat/en/download vs. https://delta.chat/de/download

The content is different

hpk42:
On Wed, Oct 09, 2019 at 13:58 -0700, Simon Laux wrote:
> https://delta.chat/en/download vs. https://delta.chat/de/download
> 
> The content is different

Something easy to fix?  don't htink translators are reading this issue tracker here ...
and the german translation has been done by various people over time. 
So simply raising the issue does not gurantee anything ;)

r10s:
i think the whole page is reworked meanwhile and this issue can be closed.

PR #185, opened 2019-10-13
Update German translation

lukkigi:
Should fix issue #183

### DIFF
--- a/de/download.po
+++ b/de/download.po
@@ -7,24 +7,24 @@
 # Moni P. <monitx@deltachat.de>, 2019
 # W R, 2019
 # Sven, 2019
-# 
+# Lukas, 2019
 #, fuzzy
 msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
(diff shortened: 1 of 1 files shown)


r10s:
thanks!

when we merge this pr, we should not forget to do a `tx push -t -l de` afterwards - otherwise the changes will be overwritten by the next pull from transifex.

in general, we should overthink this whole handling of the localized website, it seems to be too complicated ... but this is another issue ;)

ghost:
@r10s on which machine do we need to do that push? Locally, and commit this? or on the machine? now, after the migration...

r10s:
> Locally, and commit this?

yip, there is no special machine involved, the logic is mainly in ./tools/t-dance - we take the english .md file, convet it to a po file, push that to transifex

when things are translatated, we pull the translations as .po files, convert that to .md files, that is pushed then.

r10s:
if we want to keep the international pages, it would be cool if that can be automated, eg. every change on /en could be pushed as .po files to transifex

Simon-Laux:
I believe this is outdated now

PR #186, opened 2019-10-15
wrote a blogpost about the workshop at /ETC

ghost:
Hi, a small blogpost about the latest workshop :) I am still in the process of asking the other referent whether they want their name referenced. And of course that's only my perspective on the workshop and the event, so I'm open for changes. Anyway, feedback welcome :) 

It could have more links and pictures as well. @dmetzlar do you have some? We could maybe take the group photo... but let's keep in mind that this is one of the deliverables for otf, don't know whether that has any implications^^


_posts/2019-10-16-etc-workshop.md
```diff
@@ -0,0 +1,85 @@
...
+
+Mostly women and nonbinary people attended, and brought a variety of skills
+they could share between each other. I am used to the elitism of nerd culture -
+the contrast was wild. I never saw a learning athmosphere like that, where
+everyone could learn from each other, without being afraid to admit they don't
+know everything.
+
+Donna and ---- gave a workshop there about installing mail servers - a very
+interesting topic for me, I installed my first own mail server only weeks ago.
+The workshop was part of the effort to make burner E-Mail accounts a widespread
```
hpk42: was burner-accounts how it was motivated at the time?   In any case it's not a readily known concept so would be good to find a link for it if mentioned this prominently.

ghost: We don't really have a link like that, do we? Maybe we could link to the OTF proposal, but I don't know how much sense that makes.

Is there a blocker for openly talking about that feature?

hpk42: it's more about not confusing readers -- mentioning concepts as if a reader should know them alienates folks that don't know them.  Here is a link that at least has a paragraph about it in an earlier post https://delta.chat/en/2018-11-17-deltaxi#new-planned-features-for-at-risk-and-other-users

ghost:
Okay, I linked to that part, hope that suffices. I also added some images and added Jugas name.

ghost:
I got an approval from Donna.

veit:
There are a few typos:
* athmosphere → atmosphere
* beatiful → beautiful
* faraday → Faraday

PR #194, opened 2019-11-13
Improve pages

Simon-Laux:
and add the new download page


Simon-Laux:
for easier reviewing https://github.com/deltachat/deltachat-pages/pull/194/files/7c67ba7752a73708aab58367ef03f083ffd7d005

ca/download.md
```diff
@@ -9,61 +9,97 @@ layout: default-ca
...
+    </div>
+</div>
+<script src="../../assets/js/logic.js"></script>
+
+# QR Code for this site
+If want to share this site quickly with a friend, you can use this qr code:
+
+<img style="float: center" src="../assets/home/get.delta.chat.png" alt="Qr code for https://get.delta.chat" width="200" />
+
+Or just visit `https://get.delta.chat` on you're friends device.
```
Jikstra: typo: `your`

r10s (commented):
i think the general layout should be a bit more "elegant" and should fit better to the overall layout. 
- eg. i would color the buttons in the color of the header, or chose a color that fits to that.
- also maybe just use lines instead of boxes
- the font size should match the rest of the page (we can also enlarge the font in the header)
- maybe the buttons should be of the same sizes

assets/js/logic.js
```diff
@@ -0,0 +1,29 @@
+function main() {
```
ghost: You could give assets/js/logic.js a more meaningful name, e.g. recommend.js - but maybe not important, as it is neither obvious what our other js file is doing. :woman_shrugging:

ca/download.md
```diff
@@ -9,61 +9,97 @@ layout: default-ca
...
-
-* Other APK downloads:  
-  `{% include fingerprint-local %}`
-
-Si vols imprimir les petjades SHA256 del certificat de signatura APK  pots usar per ex.  
-`keytool -list -printcert -jarfile <APK-file>`
+## Badges
 
+[<img src="../assets/home/get-it-on-gplay-beta.png" alt="Beta testing on Google Play" width="200" />](https://play.google.com/store/apps/details?id=chat.delta)
+[<img src="../assets/home/get-it-on-fdroid.png" alt="Get it on F-Droid" width="200" />](https://f-droid.org/app/com.b44t.messenger)
```
ghost: Why not just add the badges to the Android download box?

ca/download.md
```diff
@@ -9,61 +9,97 @@ layout: default-ca
...
+    </div>
+</div>
+<script src="../assets/js/logic.js"></script>
+
+# QR Code for this site
+If want to share this site quickly with a friend, you can use this qr code:
+
+<img style="float: center" src="../assets/home/get.delta.chat.png" alt="Qr code for https://get.delta.chat" width="200" />
+
+Or just visit `https://get.delta.chat` on your friends device.
```
ghost, line 98: Do we really need a QR code on this page which links to it?

en/verify_downloads.md
```diff
@@ -0,0 +1,18 @@
...
+
+# Verify Downloads
+
+For Android, you can verify the signing certificate on the APK matches one of the following SHA256 fingerprints:  
+
+* For F-Droid downloads:  
+  `{% include fingerprint-f-droid %}`
+
+* Other APK downloads:  
+  `{% include fingerprint-local %}`
```
ghost, line 14: Where do we get those variables from, where do we need to change them when we release something?

ghost (changes requested):

ghost:
Thanks for your extensive work in refactoring this page! Overall great. So I mainly have some questions, and yeah, the Android buttons should be in the box I think.

en/verify_downloads.md
Simon-Laux: there were there before, I just moved that part to another file. They are under the `_includes` folder

ca/download.md
Simon-Laux: Add or replace the other buttons? I was worried that it looks not too good if we mix mash the design, but I also don't want to remove them because maybe we'll find a solution for including them in the future.
Generally as with verify downloads section and Changelog I want to preserve the information that isn't in the new page otherwise. for the first time it doesn't have to be perfect.

ca/download.md
Simon-Laux: we advertised the testflight qr so I thought it would be useful.

Simon-Laux:
the translation system enables translators to break the whole site:
```
Liquid Exception: Liquid syntax error (line 208): Tag '{% включая standards-url %}' was not properly terminated with regexp: /\%\}/ in ru/help.md
```
```
- Смотрите [Стандарты, используемые в Delta Chat]({% включая standards-url %}).
```
the include code was translated too, which makes no sense cause it breaks the site

it/download.po
```diff
@@ -86,7 +86,7 @@ msgid ""
 "        </div>\n"
 "    </div>\n"
 "</div>\n"
-"<script src=\"../../assets/js/logic.js\"></script>"
+"<script src=\"../assets/js/logic.js\"></script>"
```
ghost: missed one here, I think ;) download-page.js

nb/download.po
```diff
@@ -86,7 +86,7 @@ msgid ""
 "        </div>\n"
 "    </div>\n"
 "</div>\n"
-"<script src=\"../../assets/js/logic.js\"></script>"
+"<script src=\"../assets/js/logic.js\"></script>"
```
ghost: Here too :)

uk/download.po
```diff
@@ -86,7 +86,7 @@ msgid ""
 "        </div>\n"
 "    </div>\n"
 "</div>\n"
-"<script src=\"../../assets/js/logic.js\"></script>"
+"<script src=\"../assets/js/logic.js\"></script>"
```
ghost: another one :)

ca/download.md
ghost: I would replace them. We could make the APK button just a text link instead - it's meant for users who know what they do, after all.

But yeah, of course. Perfection is boring :P Preserving information is good :)

Simon-Laux:
> maybe the buttons should be of the same sizes

@r10s what do you mean by that? same width?

> we can also enlarge the font in the header

I'll try to do that, but we'll do another pr for more global site changes.

> also maybe just use lines instead of boxes

something like this?:

uk/download.po
Simon-Laux: not too important that are po files

Simon-Laux:
maybe we should use http://docs.translatehouse.org/projects/translate-toolkit/en/latest/commands/html2po.html for conversion instead of txt2po and use html. In future we would need our own converter that ignores html logic, jekyll liquid logic and markdown logic

Simon-Laux:
@lefherz now you have your badges inside
Though personally don't really like them because they have all sorts of trademark compliance rules attached.

ghost (approved):

ghost:
I like the look :) and I think it still looks good next to the other buttons.

Yeah, trademark rules suck. I think it's interesting how these buttons generate user trust; how many users would probably rather download from GPlay store than from our website. (Which is aligned with the Android settings, "untrusted sources" etc).

Not playing along would maybe be better than making us again dependent on stupid G**gle infrastructure. On the other hand it introduces a critical thinking-barrier for our users. Which might be a good thing. The more I think about it, the more undecided I am :D

PR #200, opened 2019-11-15
using rsync instead of scp to push previews

ghost:
in #194 some previews don't appear on the page, although scp doesn't throw any errors. Let's try whether it's better with rsync.


ghost:
Ah, the error is a totally different one. The data is moved into a newly created subfolder each time after the first push.

Simon-Laux:
where is the link in the checks?, looks like it went missing

ghost:
Yeah, that's another problem, this commit is completely useless anyway, as I have to solve a different problem than I thought

PR #203, opened 2019-11-15
maybe this way we can get the commit hash for the API request

ghost:
With the current solution, the preview link is only posted to the checks for the 2nd commit, not for the first. If this shows the link for the first commit, we are fine.


ghost:
hm, nope.

PR #204, opened 2019-11-15
maybe this way we can get the commit hash for the API request, 2nd try

ghost:
see #203


ghost:
doesn't work either; I'm out of ideas for now.

PR #205, opened 2019-11-15
remove QR code because it is not worth the space on the download page…

hpk42:
… just to link to get.delta.chat


r10s:
yip, also thought that. or at least make it smaller and put it to the very and. but i think on any random device it is probably easier to just type in `get.delta.chat`

r10s (approved):

PR #206, opened 2019-11-15
use consistent word-separators and extensions in url

r10s:
let's keep things consistent:
- on the rest of the page, we use the minus-character and not the underscore-character to separate words
- on the rest of the page, we do not use the .html extension


Simon-Laux (approved):
Looks good.
Do we need the t-dance to merge this?

r10s:
> Do we need the t-dance to merge this?

not for merging, this can be done afterwards as well. 
it's. just that /en/ and the other languages are not the same, but i do not thing this is an issue for minor tweaks.

ghost:
Can you resolve the conflict and merge it?

PR #212, opened 2019-11-21
connect yml files to transifex and other fixes.

Simon-Laux:
fix verify-downloads,
connect yml files to transifex,
t-dance now fails on error


r10s:
hard to review, 150 files :) i think we should break things into manually changed lines and things that are generated, so do the t-dance script in a separate commit.
however, as the number of generated files seems to increase, in general, i think we should get these generated files out of the repo.

hpk42:
welcome to project resurrection :)

missytake:
@Simon-Laux considering the many conflicts this branch has with master, I suggest to throw it away and redo it.

PR #214, opened 2019-11-24
Update APK link to permalink

ghost:


ghost:
I have to change https://download.delta.chat/android/ so it doesn't redirect anymore, but instead shows a file listing; the stable link doesn't work, because you can't install it easily on Android if the file name is not like the build produced it.

ghost:
It would also be cool to have the desktop preview links farther above.

ghost:
So the stable link *does* seem to work - it was just a problem with Firefox Klar.

PR #216, opened 2019-11-26
Translations via weblate

missytake:
Let's see what this workflow looks like:


ghost:
Nice ^-^ You can Git squash commit via the Weblate add-on :)

missytake:
Okay, those are .po files only. I would be cool to have them converted to .md directly.

Is there a way to convert md -> po in weblate only, and convert it back to md before committing? This way, we would not have the PRs and our repository cluttered with .po files, and could just review the changes in an easy format.

Right now we have the https://github.com/deltachat/deltachat-pages/blob/master/tools/t-dance.sh script to deal with transifex, which needs to be executed at some point in the transifex workflow (which I have never done).

ghost:
@missytake :) Check Gajim tool I've suggested them, based on F-Droid i18n tool :)
https://dev.gajim.org/gajim/website/tree/master/tools/translation

missytake:
Cool, our t-dance script pretty much seems to work with this, we just commented out the tx pull/push stuff. \o/

missytake:
okay, so this is how we could do it:

- self-hosting
- how to push the changes to weblate automatically
  - github action which creates .po files and pushes them to weblate
- how to open PRs with md automatically
  - weblate user pushes .po files to the weblate github branch
  - github action opens PR, converts .po to .md, removes .po files

@BoFFire can you maybe change the branch of the weblate project to `weblate` instead of `master` so we can play around with the actions better?

hpk42:
moving this to "project resurrection" as it's not moving currently, and certainly not before 990-series releases are properly out.

PR #220, opened 2019-12-11
answered some questions about encryption #190

### LINKED ISSUE #190: FAQ: Security of Delta Chat in comparison
quote of the requests from the user:

   - discuss what is less secure than in PGP and in WhatsApp                                       
     when keys are validated via QR setup message? [pfs, ...]                                      
   - discuss metadata, you write                                                                   
     "Delta Chat even protects your metadata (who writes to whom?)"                                
     How? This data is still available for the mail providers                                      
     (which spreads this data more than on a centralized solution),                                
     as far I can see it does not encrypt headers or remove referrers,                             
     etc.                                                                                          
   - Differences between validated groups of 2 members vs.                                         
     1-1 chat with mutual validation.                                                              
   - compare also to pep (pretty easy privacy) and show it is the                                  
     better solution. I only did not install and test this app                                     
     cause of its permissions required on Android.

missytake:
- [x] Perfect Forward Secrecy
- [x] Metadata protection
- [x] Differences between encryption of 1:1 chat and group of 2
- [x] Why not pEp

closes #190

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -23,7 +23,8 @@ recipient to install the same app as yours, as with other messengers.
 ### What are the advantages of Delta Chat compared to other messengers?
 
 - _Independent_ of any company or services. _You_ own your data.
-- Your data is not saved on a central server; this way, in contrast to most other messengers, Delta Chat even protects your metadata (who writes to whom?)
+- Your data is not saved on a central server unless all users are using
+  the same e-mail servers
 - You do not distribute your address book to anyone.
 - _Fast_ by the use of Push-IMAP.
 - _Largest userbase_ - recipients _not_ using Delta Chat can be reached as well.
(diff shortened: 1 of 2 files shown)


missytake:
I also sneaked in a change in the headlines - they now have a proper hierarchy. The h1 headline is generated by jekyll, so we only need to specify h2 headlines in the text.

missytake:
Okay, @hpk42 ready for review :)

en/help.md
```diff
@@ -187,6 +187,19 @@ The user's profile shows some additional info:
 
 - OpenPGP. Key transport is done via [Autocrypt](https://autocrypt.org).
 
+### What is the difference between verified 1:1 chats and verified groups?
+
+- Verified 1:1 chats and verified groups are not the same, even if there are
```
r10s: i think, there are no "Verified 1:1 chats" at all, so they cannot be compared to sth. else.

maybe better write "1:1 chats with a verified contact" here.

missytake: Thanks for the feedback :) Better?

hpk42:
On Thu, Dec 12, 2019 at 08:29 -0800, missytake wrote:

> missytake commented on this pull request.
> 
> > @@ -187,6 +187,19 @@ The user's profile shows some additional info:
>  
>  - OpenPGP. Key transport is done via [Autocrypt](https://autocrypt.org).
>  
> +### What is the difference between verified 1:1 chats and verified groups?
> +
> +- Verified 1:1 chats and verified groups are not the same, even if there are
> 
> Better?

No, "Verified 1:1 chats" do not exists at all. They are not a thing. 
I think i posted something yesterday trying to explain this. 
Maybe you can re-check and per-use some fragments from what i wrote?

missytake:
Yes, bjoern already pointed that out, I pushed a correction.

en/help.md
```diff
@@ -182,19 +183,72 @@ The user's profile shows some additional info:
...
+- As Delta Chat is a decentralized messenger, the metadata of Delta Chat users
+  are not stored on a central server. However, they are stored on the mail
+  servers of the sender and the recipient of a message.
+- To additionally protect end-to-end-encrypted messages, most of the [headers
+  are end-to-end-encrypted](https://datatracker.ietf.org/doc/draft-autocrypt-lamps-protected-headers/)
+  as well. Only the headers which are necessary for transmitting the E-Mail from
+  sender to receiver remain unencrypted.
+- This mostly comes into effect in group chats; the mail server of the receiver
+  of an encrypted message does not know who else received the message; it only
+  knows the sender and the recipient.
```
hpk42: Each mail server currently know about who sent and who received a message by tracking the unencrypted To/Cc headers and thus determine who (in terms of e-mail address) is part of the group. Delta Chat itself could avoid  unencrypted To/Cc headers quite easily and always put them only into the encrypted section.  The main concern here is how it affects other Autocrypt-capable mail apps.

adbenitez:
> The main concern here is how it affects other Autocrypt-capable mail apps. 


We could do it opportunistic, sounds really cool to do it at least if everybody is using a DC-compatible app.
This is something people always complain about email, if DC improves it a bit, it is one more reason to use DC ;)

hpk42:
we can start hiding to/cc in verified groups: 
https://github.com/deltachat/deltachat-core-rust/issues/1032

otherwise tracking who is a DC client needs to be looked at
but i guess this better happens after or during implementation of #1032:

https://github.com/deltachat/deltachat-core-rust/issues/1053

PR #224, opened 2019-12-16
FAQ on avatars

ghost:


en/help.md
```diff
@@ -56,6 +56,17 @@ Other messages do not appear automatically, and are to be found in the main menu
...
+### Who sees my profile picture?
+
+- You can add a profile picture in your settings. If you write to your
+  contacts, other Delta Chat users automatically see it as your profile
+  picture.
+
+- Contacts who don't use Delta Chat see it as an E-Mail attachment.
+
+- For privacy reasons, no one sees your profile picture until you write a
+  message to them.
```
hpk42, line 67: it's fine but we need to say here that it's only sent occassionally along with your messages, and not with every message.

hpk42 (approved):
lgtm apart from the comment, please add a sentence and merge then ;)

r10s (approved):
with the comments of @hpk42 targeted, this lgtm

en/help.md
r10s: avatars are also shared by a secure join. as users may not be aware that this actually sends messages out, maybe mention this explicitly or maybe say "until you directly contact them".

r10s:
still not sure of the "write a messages to them" includes secure-join from the view of the user. but i do no longer want to block merging :)

r10s (approved):

ghost:
"add them via QR code" does mean that, or not? I wanted to use that phrase because I'm not sure how often users see the term "secure-join"

r10s:
> "add them via QR code" does mean that, or not?

sure, for whatever reason, i've not seen that :)

PR #237, opened 2020-01-07
Publish 0.901.0

ghost:
It's not sure yet whether the Windows portable is stable enough; test before merging.


Jikstra:
For sure more stable than the 0.840.0.

PR #242, opened 2020-01-09
Fall back to English translation

### LINKED ISSUE #238: menus broken on some languages
eg. https://delta.chat/ca/ displays the footer as `| |` the translations for "Imprint" etc. are missing.

and for https://delta.chat/fr/ the download (!) link in the header is gone:

Screen Shot 2020-01-08 at 22 21 23

Simon-Laux:
this could be way nicer code if liquid supported functions

fixes #238

### DIFF
--- a/_includes/download-boxes.html
+++ b/_includes/download-boxes.html
@@ -1,81 +1,82 @@
+{% assign txEn = site.data.lang['en']['en'] %}
 {% assign tx = site.data.lang[page.lang][page.lang] %}
 
 <div class="download-content">
     <div id="recommendation-section" hidden>
-        <h1>{{tx.download.recommendation}}:</h1>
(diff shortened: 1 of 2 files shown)


r10s (approved):
if this works, we should merge it (ci is not yet ready). at some point, we can re-think the approach.

ghost (approved):
works: https://staging.delta.chat/242/ca/

Simon-Laux:
The right way would to make and use an jekyll plugin in ruby for this. https://jekyllrb.com/docs/plugins/
maybe @pabzm knows how it is done. I believe his ruby skills are way higher than mine (mine are nonexistent)

pabzm:
I could write a jekyll plugin, yes. But if it works currently, I'd rather spend my time with other things. If it becomes a problem again, please hit me up then.

PR #245, opened 2020-01-10
Simplify FAQ, add basic script to generate offline help

r10s:
this pr adds a script that generates the local-help from the generated html-files.

currently, only android-help is generated in english, but desktop and other languages can be added by just adding corresponding lines in the python-script (generate_help(<desktop-path>) resp. generate_lang(<lang>))

moreover, this pr simplifies the faq page as such and adds "go to top"-links.

- corresponding android pr: https://github.com/deltachat/deltachat-android/pull/1207
- corresponding desktop pr: https://github.com/deltachat/deltachat-desktop/pull/1322
- preview of the _online_-faq: https://staging.delta.chat/245/en/help


en/help.md
```diff
@@ -5,21 +5,20 @@ render_toc: true
 header: Frequently Asked Questions
 ---
 
-## General
 
-### What is Delta Chat? 
-
-<img style="float: right;" width="400px" src="../assets/home/delta-what.png" />
```
pabzm, line 12: Ohh, I like the image! Do you want to remove it for pragmatic reasons, to keep the current task simple? Or to permanently get rid of it?

*Edit*: In general I would rather add more icons or simple images, than removing the only one. Walls of text are hard to deal with, in my opinion.

hpk42 (approved):
lgtm so far

en/help.md
r10s: i agree, i just wanted to have things easy as a first attempt. but i think we can re-add it :)

PR #253, opened 2020-01-15
adapt "sed" to work on mac

r10s:
freebsd seems to be different from unix here.
i did not figure out how to adapt the statement
that adds the "GENERATED" hint to the content.
however, comments in markdown are not really standard anyway,
so i decided to make the hint in the README clearer.

in general, i think the t-dance script
should be replaced by a python script as some point -
but, for now, it works again ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,10 +49,9 @@ Directory structure
 
 - `/en` - Contains the source language files in Markdown. Can be edited directly.
 
-- `/de`, `/es`, `fr` … - Contain the translated files; 
-  translation is done through .po files that are translated using Transifex, see above. 
-  You must not edit translated files directly here, 
-  but you can add additional files that are unique to the language.
+- `/de`, `/es`, `fr` … - Contain the translated files, use Transifex to edit them;
+  you **MUST NOT** edit any file in these directories,
+  they will be overwritten with the next pull from Transifex.
 
 - `/_layouts` - This directory contains a default layout template 
   for each language (the layout is referenced in Markdown using _layout: name_)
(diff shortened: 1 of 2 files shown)


hpk42 (approved):

ghost:
As long as the reviewers of PRs have read the README, it should be fine :) Thanks!

ghost:
Sorry, I should have tested this before merging - the script breaks on my local machine now. I reverted the merge, and open a new PR for further testing.

r10s:
@lefherz it is probably the `-I` option for `sed` - this is the only change. maybe we should not try to figure that out but take this as an opportunity to rewrite the script in python.

r10s:
nb: the script always fails with some weird errors when no action command is given. to try out the sed-thingie, call it as `./tools/t-dance.sh create-md`

r10s:
@lefherz or, did you tried the `-E` option, https://unix.stackexchange.com/questions/101059/sed-behaves-different-on-freebsd-and-on-linux

ghost:
Yeah, already saw those errors, we should just print `usage` in that case...

I'm replacing the sed command with one python command atm. Can you try out whether `tools/t-dance.sh pull` works on Mac? If it works both on Mac and Linux, that would be a start. It's on the `fix-sed-for-mac` branch...

The git history is not very pretty, but I tried my best^^

ghost:
trying sed -iE as well right now...

ghost:
sed -iE still fails for me with the same error: (just realized I didn't post it)

```
[...]
tx INFO: Done.
Creating markdown files from the translated po-files ...
processing ca/blog.po
sed: can't read s/lang: [a-z][a-z]/lang: ca/: No such file or directory
```

ghost:
What is the `""` argument in the sed command for?

r10s:
the `-I` tells to do things "in place" and the next parameter _might_ be the name/extension of backup files to create. this varies between implementations and this makes things unportable.

ghost:
ah. hrm. let's see if #257 is portable? or use gsed?

r10s:
let me check #257 first, maybe everything is fine then :)

PR #257, opened 2020-01-16
Fix sed for mac

ghost:
@r10s can you try this on mac?


ghost:
It adds massedit.py as a dependency though, which can be installed with `pip3 install massedit`.

r10s (approved):
yeah, that works! \o/ had to install via `pip3 install massedit`, though. mess everywhere.

PR #264, opened 2020-01-19
re-add faq-image

r10s:
targets comment of @pabzm at https://github.com/deltachat/deltachat-pages/pull/245#pullrequestreview-341869090 , however,  the image is optimized (24k instead of 164k, resolution is unchanged)   and shown a bit smaller. also i avoided all floating tricks, as this would make things more tricky on small screens.

preview: https://staging.delta.chat/264/en/help#what-is-delta-chat


pabzm:
Thanks!

My intention was to give the eyes something different to look at than the wall of text, though.
I feel that without floating the image rather just breaks that wall in two parts now.

I currently don't have a productive idea how to improve that, though.

If anyone else is in favour of this PR, please don't feel blocked. But if my previous comment is the main reason to do it, I'd rather postpone and revisit it later, with a clearer concept in my head.

r10s:
i tried floating-images, however, once the screen gets narrower this tends to look ugly, eg. as left of the image there is only a column of one word below each other (or even worse, one character over the other or so, seen on android).  
tweaking the css is doable, but needs some testing, esp. on android devices then.

but maybe, as a compromise, we can solve the problem by putting the image after once main chapter? so it does not break reading too much? in this case eg. after the chapter "what is delta?" and "groups?"

r10s:
i think i've found a solution with the last commit that works at least for web and android.



we could also use different classes or so, however, i think the problem in browsers with small screen is the same, only a bit more hidden.

ghost:
Looks great!

PR #268, opened 2020-01-21
New FAQs

### LINKED ISSUE #250: FAQ entry on differences to protonmail/tutanota/spike/criptext
all of the four offer e2e encryption (spike maybe not even that) only for their respective users. All of them don't offer IMAP/SMTP and thus are technically no e-mail providers, just allow to receive and send messages to e-mail addresses (unencrypted).

### LINKED ISSUE #267: New FAQ entries
We need some new FAQ entries:

- [x] About receiving messages in the background: https://twitter.com/delta_chat/status/1216457864870596609
- [x] About Delta Chat Web: https://twitter.com/delta_chat/status/1215640268633989121

ghost:
closes #250 

closes #267

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -347,6 +347,13 @@ capable) devices, you have to sync their encryption capabilities:
 - You are now synchronized, and can use both devices for
   sending and receiving E2E-encrypted messages with your communication partners. 
 
+### Are there any plans for introducing a Delta Chat Web Client?
+
+- There are no immediate plans but some preliminary thoughts.
+- There are two avenues for introducing web delta chat, but both are
+  significant work. For now, we focus on getting stable releases into all
+  appstores (Google Play/iOS/Windows/macOS/Linux repositories) as native apps.
+
 ### Why can I choose not to watch the Inbox?
 
 This is an experimental setting for some people who are experimenting with
@@ -402,6 +409,22 @@ Otherwise, deleting messages or multi-device setups might not work properly.
   see [Provider Overview](https://providers.delta.chat)
 
 
+### If Delta Chat uses E-Mail, is it really an _Instant_ Messenger?
+
+- Generally, E-Mail packets travel as fast as anything else on the Internet. So
+  the transmission itself is as instant as it can be - a few seconds at most.
+- There is another issue that can delay message deliverly though: receiving
+  messages in the background. For privacy reasons, Delta Chat doesn't use
+  Google Cloud Messaging (GCM) or the Apple Push Notification Service (APNS).
+- On Android, this is usually not a problem, if you allow Delta Chat to run in
+  the background. On iOS that's unfortunately less reliable. On Desktop, it's
+  no problem at all.
+- That Android and iOS kill apps running in the background is a general problem
+  of many apps. For more information, see
+  [dontkillmyapp.com](https://dontkillmyapp.com/).
+
+
 ### I'm interested in the technical details. Can you tell me more?
 
 - See [Standards used in Delta Chat]({% include standards-url %}).
+
--- a/en/help.md
+++ b/en/help.md
@@ -405,3 +405,13 @@ Otherwise, deleting messages or multi-device setups might not work properly.
 ### I'm interested in the technical details. Can you tell me more?
 
 - See [Standards used in Delta Chat]({% include standards-url %}).
+
+### Is there a "Don't Disturb Me" mode, where I don't get new messages?
+
+If you don't want to be reachable for a while, you can disable "Watch Inbox
+folder" in the settings. This way, Delta Chat doesn't receive new messages
+until you re-enable the option again.
+
+If you want Delta Chat to consume no bandwidth at all, you need to disable all
+"watch folder"-settings.
+
--- a/en/help.md
+++ b/en/help.md
@@ -353,6 +353,11 @@ capable) devices, you have to sync their encryption capabilities:
 - There are two avenues for introducing web delta chat, but both are
   significant work. For now, we focus on getting stable releases into all
   appstores (Google Play/iOS/Windows/macOS/Linux repositories) as native apps.
+- If you need a Web Client, because you are not allowed to install software on
+  the computer you work with, you can use the portable Windows Desktop Client,
+  or the AppImage for Linux. You can find them on
+  [get.delta.chat](https://get.delta.chat).
+
 
 ### Why can I choose not to watch the Inbox?
 
@@ -412,7 +417,7 @@ Otherwise, deleting messages or multi-device setups might not work properly.
 ### If Delta Chat uses E-Mail, is it really an _Instant_ Messenger?
 
 - Generally, E-Mail packets travel as fast as anything else on the Internet. So
-  the transmission itself is as instant as it can be - a few seconds at most.
+  chatting itself is as instant as it can be - a few seconds at most.
 - There is another issue that can delay message deliverly though: receiving
   messages in the background. For privacy reasons, Delta Chat doesn't use
   Google Cloud Messaging (GCM) or the Apple Push Notification Service (APNS).
--- a/en/help.md
+++ b/en/help.md
@@ -440,9 +440,10 @@ Otherwise, deleting messages or multi-device setups might not work properly.
   Protonmail, Tutanota, or Criptext. Those messages will not be End-to-End
   encrypted, though. The End-to-End encryption those providers offer is only
   working inside their platforms, and not compatible with anyone outside.
-- There are two exceptions: Tutanota has Autocrypt support on their roadmap,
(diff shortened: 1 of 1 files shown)


en/help.md
```diff
@@ -405,3 +405,13 @@ Otherwise, deleting messages or multi-device setups might not work properly.
...
 - See [Standards used in Delta Chat]({% include standards-url %}).
+
+### Is there a "Don't Disturb Me" mode, where I don't get new messages?
+
+If you don't want to be reachable for a while, you can disable "Watch Inbox
+folder" in the settings. This way, Delta Chat doesn't receive new messages
+until you re-enable the option again.
+
+If you want Delta Chat to consume no bandwidth at all, you need to disable all
+"watch folder"-settings.
```
Simon-Laux: I don't agree with making this hack official as it could introduce many problems

r10s: i would also not recommend these settings for being switched on and off officially. they are also too hidden and more advanced settings in a section that is not for everyday-usage.

you can disable notifications if you do not want to being disturbed (however, most ppl will just silence their phone)

and if you want to save traffic, you can "kill" the app. but this is also nothing for an "official" help.

r10s (changes requested):
so, i would not bring up this question in the official faq at all.  i think it is okay to not answer _every_ asked question (eaq :) - but we can add this to the forum.

comradekingu:
The official FAQ was going to be removed entirely the other day (if so good riddance). It is a grim document scaring away potential users. The idea of a FAQ symposium is as counterproductive as a FAQ is to getting any sort of overview.

Who says "Hey, could you give me the lowdown on Delta Chat, and could you give it to me in the form of randomly assorted questions, possibly throwing in the odd bonus hack?"

en/help.md
comradekingu: On the assumption the user set up another folder to watch, avoiding a specific one, or fewer than all isn't going to help(?)

ghost:
I added answers to two more questions which came up on Twitter. They come up quite often. In the future, we can just link to them in response.

en/help.md
```diff
@@ -402,6 +409,22 @@ Otherwise, deleting messages or multi-device setups might not work properly.
   see [Provider Overview](https://providers.delta.chat)
 
 
+### If Delta Chat uses E-Mail, is it really an _Instant_ Messenger?
+
+- Generally, E-Mail packets travel as fast as anything else on the Internet. So
+  the transmission itself is as instant as it can be - a few seconds at most.
```
r10s: maybe add the word "chatting" here - sth, as "So chatting itself is as instant as it can be  - a few seconds at most". this would be a bit less abstract.

en/help.md
```diff
@@ -347,6 +347,13 @@ capable) devices, you have to sync their encryption capabilities:
 - You are now synchronized, and can use both devices for
   sending and receiving E2E-encrypted messages with your communication partners. 
 
+### Are there any plans for introducing a Delta Chat Web Client?
+
+- There are no immediate plans but some preliminary thoughts.
+- There are two avenues for introducing web delta chat, but both are
+  significant work. For now, we focus on getting stable releases into all
+  appstores (Google Play/iOS/Windows/macOS/Linux repositories) as native apps.
+
```
r10s, line 361: if this is really an faq, i would also point to the independent desktop client as a possible solution here. this is sth. delta can do - what eg. whatsapp cannot.

ghost:
I'll quickly write something more addressing #250 - I will ask again for feedback afterwards.

en/help.md
```diff
@@ -424,6 +424,17 @@ Otherwise, deleting messages or multi-device setups might not work properly.
...
 
 
+### Is Delta Chat compatible with Protonmail / Tutanota / Criptext?
+
+- Yes and No. 
+- Yes, you can use Delta Chat to send E-Mails to people who use Protonmail,
+  Tutanota, or Criptext.
+- No, you can not use your Tutanota or Criptext account with Delta Chat; they
+  do not offer receiving mails via IMAP.
+- You can use Protonmail with Delta Chat if you use their [IMAP
```
r10s: not sure, however, i think even the bridge has issues (cc @pabzm ) - so maybe it is better to say just "no".
we can add a "howto" to the forum, but i would not add links to shitty stuff in the official faq.

ghost:
Ah, I saw your IMAP Bridge comment a bit late. I removed that part.

ghost:
Oh, note that this PR talks about the Windows portable client, which is not yet released, and not available on get.delta.chat. Maybe we should only merge this after the next desktop release, which is pretty near afaik.

en/help.md
```diff
@@ -402,6 +414,36 @@ Otherwise, deleting messages or multi-device setups might not work properly.
...
+### Is Delta Chat compatible with Protonmail / Tutanota / Criptext?
+
+- Yes and No.
+- No, you can not use your Protonmail, Tutanota, or Criptext account with Delta
+  Chat; they do not offer receiving mails via IMAP.
+- In any case you can use Delta Chat to send Messages to people who use
+  Protonmail, Tutanota, or Criptext. Those messages will not be End-to-End
+  encrypted, though. The End-to-End encryption those providers offer is only
+  working inside their platforms, and not compatible with anyone outside.
+- Tutanota has Autocrypt support [on their
```
r10s: i would not add deep links to some hard to understand discussions to the official-faq. remember: the official-faq is for the non-nerd-end-user (parents etc :) and i do not think this link is helpful for them.

maybe we can mention the Tutanota-plans - but i've not even found sth. about this on _their_ official pages, so i am not sure how useful this is.

so maybe just strike the last point.

en/help.md
```diff
@@ -402,6 +414,36 @@ Otherwise, deleting messages or multi-device setups might not work properly.
   see [Provider Overview](https://providers.delta.chat)
 
 
+### If Delta Chat uses E-Mail, is it really an _Instant_ Messenger?
+
+- Generally, E-Mail packets travel as fast as anything else on the Internet. So
+  chatting itself is as instant as it can be - a few seconds at most.
```
hpk42: That's maybe a bit bold.  Rather: 

Sending and receiving messages takes a few seconds, typically.  Sometimes there are cases where it takes longer but that is arguably true as well for any other messenger.

en/help.md
```diff
@@ -402,6 +414,36 @@ Otherwise, deleting messages or multi-device setups might not work properly.
...
+- There is another issue that can delay message deliverly though: receiving
```
hpk42:
```suggestion
- There is an issue that can delay message delivery, though: receiving
```

en/help.md
```diff
@@ -402,6 +414,36 @@ Otherwise, deleting messages or multi-device setups might not work properly.
...
+  messages in the background. For privacy reasons, Delta Chat doesn't use
+  Google Cloud Messaging (GCM) or the Apple Push Notification Service (APNS).
+- On Android, this is usually not a problem, if you allow Delta Chat to run in
+  the background. On iOS that's unfortunately less reliable. On Desktop, it's
+  no problem at all.
+- That Android and iOS kill apps running in the background is a general problem
+  of many apps. For more information, see
+  [dontkillmyapp.com](https://dontkillmyapp.com/).
```
hpk42, line 432: suggestion: 

Instant chatting works fast if both parties are actively using the app. It's sometimes slower if the app is running in background. Receiving messages then can take minutes because both Android and iOS often stop Delta Chat from running in the background, and only wake it up occassionally.  This artifical delay is usually worse on iOS than on Android. Note that Delta Chat doesn't use Google Cloud Messaging (GCM) or the Apple Push Notification Service (APNS), because this leads to user tracking and central control which Delta Chat aims to avoid as much as feasible.  

However, that Android and iOS kill apps running in the background is a problem for many legitimate apps. For more information, see [dontkillmyapp.com](https://dontkillmyapp.com/).

en/help.md
hpk42: additionally it's not clear how much of a commitment this issue is.  I am also for striking the last sentence with the link and say instead: 

Delta Chat can e2e-encrypt through any e-mail provider with any [https://autocrypt.org/dev-status.html](Autocrypt-enabled e-mail app).

en/help.md
```diff
@@ -347,6 +347,18 @@ capable) devices, you have to sync their encryption capabilities:
 - You are now synchronized, and can use both devices for
   sending and receiving E2E-encrypted messages with your communication partners. 
 
+### Are there any plans for introducing a Delta Chat Web Client?
+
+- There are no immediate plans but some preliminary thoughts.
+- There are two avenues for introducing a Delta Chat Web Client, but both are
```
hpk42: 2-3 different avenues 

(since today) :)

hpk42 (approved):

PR #280, opened 2020-02-02
Update the version of desktop and put repetitive version string into …

Jikstra:
…a variable


ghost:
The download links throw 404 errors. While the folder indeed has the `v` in the beginning, the filenames of the downloads don't use the `v` in the version number.

Jikstra:
I changed the upload script running on the bbmac to drop the v and ruploaded the 0.999.1 release, so it should now work.

Jikstra:
I would like to have it consistent, so trying to fix it on the electron builder side.

ghost:
The links all work: https://staging.delta.chat/280/en/download

And portable is visible now as well.

PR #281, opened 2020-02-03
Fix the link to Autocrypt-enabled e-mail apps

miikka:
The Markdown link syntax was the wrong way around.


r10s:
thanks for taking care and for fixing 👍

Issue #139, opened 2019-02-19
Translations not updated frequently


adbenitez:
I had made a mistake with a link, I had fixed it as soon as I noticed it, but that was some days ago and the translations seems not to be updated yet, see the broken GDPR link: https://delta.chat/es/

r10s:
i've just updated the translations.

currently, it's a manual process, basically `t-dance pull` has to run locally from the `tools` directory and there are only few persons that have set up the tools needed on their machines.

so, if someone is up to improve this process, this would be very welcome :)

ghost:
Does anything speak against a github action which executes something like `t-dance.sh pull && t-dance.sh push` and commits it to master? maybe after each pull request?

I'm currently doing that manually from time to time, and take a look at the diff before I push it to GitHub. But if there is never any problem, we could just automate this of course.

r10s:
i'd say a `t-dance.sh push` should always work after commits/merges to master - unless we break things badly. but then it is our fault :)  

not sure about a `t-dance.sh pull` - doing that too fast or too often can be annoying for the translators as "their" language page is full of english texts then - so, maybe better do that manually, once or twice a week or so. this also gives us the option do do a quick `diff` and see if there are rough errors.

we can see how this works, maybe with the automated `t-dance.sh push`, translations will be faster so that we can rethink doing also the pull automatically.

Simon-Laux:
we could either only push reviewed language strings from transifex or the action could just see if jekyll build still works afterwards.

ghost:
To keep in mind: when someone does a `tools/t-dance.sh pull` locally, and pushes this to master, the github action from #285 gets executed again, and pushes them to transifex.

PR #283, opened 2020-02-05
Fixed yaml translations

### LINKED ISSUE #210: readd YAML files to t-dance
with the help of [yaml2po](http://docs.translatehouse.org/projects/translate-toolkit/en/latest/api/convert.html#module-translate.convert.yaml2po) and po2yaml.

ghost:
closes #210 

Seems the last steps are really not hard. Testing right now whether this does the job already.


ghost:
This can unfortunately not be tested with our staging environment, as the yaml files are generated by `tools/t-dance.sh pull`.

To test this:
* check out this branch
* execute `tools/t-dance.sh pull` (you might have to setup transifex locally for this)
* run `jekyll serve` to build it locally.

Following things should be the case:

* on http://127.0.0.1:4000/fr/, you should see the top menu translated.
* on http://127.0.0.1:4000/fr/download/, you should see "Recommended for you" translated to french.
* on http://127.0.0.1:4000/nb/, you should see the top menu in english, because the strings are missing.

Am I forgetting something or can we merge this?

Simon-Laux:
Do you plan to do the t-dance in a separate pr?

ghost:
You mean #139? yes.

Simon-Laux:
please rebase this, also did you get it running on your local machine?

tools/t-dance.sh
```diff
@@ -28,7 +28,7 @@ pull_po_translations_from_tx() {
 	#find ../_data/lang/ -type f -not -name 'en.*' -delete
 	rm -r translations || true
 	mkdir translations
- 	#ln -s -T ../../_data/lang translations/delta-chat-pages.yml
+ 	ln -s -T ../../_data/lang translations/delta-chat-pages.yml
```
r10s: is this `../../` really correct? we're in the tools directory when this script is running, `../../` would go outside the repo imu, not sure if this is desired. also, it is not clear to me what file should be linked and why. maybe add an comment on that.

tools/t-dance.sh
```diff
@@ -40,7 +40,7 @@ pull_po_translations_from_tx() {
 
 
 push_po_sources_to_tx() {
- 	#ln -s -T ../../_data/lang translations/delta-chat-pages.yml || true
+ 	ln -s -T ../../_data/lang translations/delta-chat-pages.yml || true
```
r10s: same here.

tools/t-dance.sh
Simon-Laux: yes I believe its correct, we could maybe do a `cd translations` before and a `cd ..` afterwards?

r10s: so it is not correct :)
however, why not do a `ln -s -T ../_data/lang translations/delta-chat-pages.yml`?
i have disabled the lines some time ago because it stops the whole script from working - and i do not really understand the gist up to now. `../_data/lang` is a directory - and this should be linked to the file `translations/delta-chat-pages.yml`? and why not copy them as the other files some lines below?

@lefherz do you have an idea?

ghost: `../../` is correct, because this relative path is not interpreted relatively from the view of `PWD`, but relatively from the link, which is in `/tools/translations/`.

The advantage of the link is that tx push & pull always talk to the same file, and can apply diff logic. With the .po files, this is done in the background throught txt2po etc., but in the yml cases it might lead to conflicts. Hard to say, because I can't estimate all possible outputs of tx pull & txt2po, but as it works with `ln` for now, I would keep it that way.

ghost:
I got it running on my local machine, quite fine actually.

ghost:
I'd like to merge #285 before this PR, can someone take a look at it?

tools/t-dance.sh
r10s: k, so with the added `cd translations` it makes more sense.

r10s (changes requested):
getting `ln: illegal option -- T` on macos, but maybe also on other unix. in my experience, `ln` always causes troubles, so i would really try to get rid of this. 

tx pull/push is not called directly, so, maybe a cp will do the job as well?

r10s:
i think there is still another `ln` - does the `cp` in general also work on your machine?

ghost:
Okay, I replaced the `ln` with `cp`, but only in pull. That works, I tested it.

In push it doesn't seem to work, if I replace it with `cp`, `t-dance.sh push` throws the following error:

```
tx INFO: Pushing source file (translations/delta-chat-pages.indexpo/en.po)
tx INFO: Pushing resource delta-chat-pages.referencespo
tx INFO: Pushing source file (translations/delta-chat-pages.referencespo/en.po)
tx INFO: Pushing resource delta-chat-pages.verifiy-downloadspo
tx INFO: Pushing source file (translations/delta-chat-pages.verify-downloadspo/en.po)
tx INFO: Pushing resource delta-chat-pages.yml
tx ERROR: You don't seem to have a proper source file mapping for resource delta-chat-pages.yml. Try without the --source option or set a source file first and then try again.
tx INFO: Done.
```

But that doesn't matter much, if push doesn't work on macOS. Since #285, push is regularly executed by a GitHub action, we don't really need to do that manually anymore. As the GitHub actions don't have a problem with `ln`, I would leave it as it is.

r10s:
i still don't get it.

- `tx push` expects the source files to be in the `tools/translations` directory.
- for the markdown stuff, we convert+copy all files `create_po_sources()`
- why cannot we just copy the yml file before calling `tx push`? it is only a single file, so doing a `cp "../_data/lang/en.yaml" "translations/delta-chat-pages.yml/en.yml"` should do the job

i did a pr on top of this one with the suggested chages, https://github.com/deltachat/deltachat-pages/pull/298

r10s (approved):

PR #285, opened 2020-02-06
Push strings to transifex on push to master

### LINKED ISSUE #139: Translations not updated frequently

ghost:
closes #139

### DIFF
--- a/.github/workflows/tx-push.yml
+++ b/.github/workflows/tx-push.yml
@@ -21,6 +21,7 @@ jobs:
         pip install massedit
     - name: tx init
       run: |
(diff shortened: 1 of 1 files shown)


ghost:
This is not really testable, as there are no branches on transifex. On the other hand, it doesn't hurt much.

Anyway, we should check the latest run of this action again after merging #283 to master; it makes changes to the t-dance script, and even though it should just work, let's take a closer look.

PR #292, opened 2020-02-13
double ticks in groups #290

### LINKED ISSUE #290: FAQ: do the double ticks in group messages mean that everyone has read my message?
"Can I ask you a question? In groups the double ticks mean that everyone has read the message? 🤔"

Not sure which is the current behavior:

- [ ] double ticks appear if one person has read the message
- [x] double ticks appear if >= 50% of the recipients have read the message
- [ ] double ticks appear if all recipients have read the message

I'm also not sure what happens if some group members have disabled read receipts.

ghost:
closes #290

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -174,14 +174,13 @@ about feedback which makes the app more secure for all of our users.
 
 ### What do double ticks mean in group messages? Has everyone read my message now?
 
-- A message shows double ticks, if at least 50% of the recipients have sent you
-  a "read receipt" in the background.
-- This behavior protects people who disabled read receipts in their privacy
-  settings. If you want to understand the background of this design decision,
-  [this blogpost](https://delta.chat/en/2017-07-06-read-receipts-and-social-pressure)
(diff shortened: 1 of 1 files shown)


en/help.md
```diff
@@ -172,6 +172,18 @@ about feedback which makes the app more secure for all of our users.
   can still write, but are no longer notified of any new messages.
 
 
+### What do double ticks mean in group messages? Has everyone read my message now?
+
+- A message shows double ticks, if at least 50% of the recipients have sent you
+  a "read receipt" in the background.
```
hpk42: "in the background" sounds a little weird ... and it's better to avoid key points here but just have one paragraph IMO. 
here is a refined overall suggestion:  "A message shows double ticks after more than half the the recipients have seen the message on their display. Note that some recipients might have disabled read-receipts for privacy-reasons. If more than 50% in a group disabled read receipts, the double ticks will never show up. For more detailed information, see [this blogpost]."

PR #297, opened 2020-02-20
update translations

r10s:
just a `./tools/t-dance pull`, last one was about 3w ago

### DIFF
--- a/it/gdpr-website.md
+++ b/it/gdpr-website.md
@@ -1,64 +1,62 @@
----
-title: Privacy policy website
+titolo: Privacy del sito web
 lang: it
----
 
 
-# Privacy Policy
+#  Politica Privacy 
 
-**Status: January 21, 2020**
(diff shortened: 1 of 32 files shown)


it/gdpr-website.md
```diff
@@ -1,64 +1,62 @@
----
-title: Privacy policy website
+titolo: Privacy del sito web
```
ghost: Can you do a pull again? I changed that in transifex.

ghost (changes requested):
Apart from that it looks good :) it doesn't yet pull yaml though, so I'll do this again after merging #283 .

it/gdpr-website.md
r10s: thanks, good catch!
maybe we should add a simple check to the t-dance script.

ghost (approved):
Thanks!

PR #299, opened 2020-02-21
create directory explicitly

r10s:
should fix https://github.com/deltachat/deltachat-pages/runs/460438194?check_suite_focus=true


ghost:
I already pushed a fix directly to master. Do you think we need the -p flag? the translations directory is already created a few lines above.

r10s:
nah, i think it's fine.

PR #303, opened 2020-02-21
call 'tx push' only if we're on master

### LINKED ISSUE #302: Do a check before tx push - are we on the master branch?
Calling `tools/t-dance.sh push` is quite dangerous and can destroy the work of our translators, if it's done from an outdated branch.

We should check whether it's executed on the master branch. It would be also good to check whether it's up to date with origin/master.

r10s:
avoids some accidents waiting to happen :)

closes #302


ghost:
Looks good, thanks!

PR #304, opened 2020-02-22
generate ios offline help

r10s:
- add ios target directory

- as our ios-help-viewer currently lacks a back-button (and this also seems to be superfluous currently), add some back-to-top links as a simple solution

targets https://github.com/deltachat/deltachat-ios/issues/520 and https://github.com/deltachat/deltachat-ios/issues/539

### DIFF
--- a/tools/create-local-help.py
+++ b/tools/create-local-help.py
@@ -30,7 +30,7 @@ def write_file(filename, content):
     f.close()
 
 
-def generate_file(destdir, lang, file):
+def generate_file(destdir, lang, file, add_top_links):
     print("generate local help in " + destdir + "/" + lang + "/" + file)
 
     content = read_file("../_site/" + lang + "/" + file)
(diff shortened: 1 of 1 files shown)


ghost (approved):
Looks good, but I don't have an iOS device to further test how it looks like.

Is it intended that 

```
content = "<h3>headline</h3><p>lorem ipsum.</p>"
top_link = "<p class=\"back\"><a href=\"#top\">^</a></p>"
```

return

```
'<p class=\\"back\\"><a href=\\"#top\\">^</a></p><h\\1>headline</h3><p>lorem ipsum.</p><p class="back"><a href="#top">^</a></p>'
```

, so that you have to back links pointing to the top, at the above and below every entry? Wouldn't one suffice?

r10s:
> so that you have to back links pointing to the top, at the above and below every entry?

they are not.

the top-links are only above every headline and at the very end.

PR #306, opened 2020-02-24
fix automated tx push (master branch only)

### LINKED ISSUE #302: Do a check before tx push - are we on the master branch?

ghost:
closes #302

### DIFF
--- a/.github/workflows/tx-push.yml
+++ b/.github/workflows/tx-push.yml
@@ -25,5 +25,8 @@ jobs:
         #tx set --auto-remote https://www.transifex.com/projects/p/delta-chat-pages/
     - name: tx push
       run: |
+        git checkout master
+        git pull origin master
+        git status
         tools/t-dance.sh push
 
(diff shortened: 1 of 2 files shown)


.github/workflows/tx-push.yml
```diff
@@ -25,5 +25,8 @@ jobs:
         #tx set --auto-remote https://www.transifex.com/projects/p/delta-chat-pages/
     - name: tx push
       run: |
+        git checkout master
+        git pull origin master
+        git status
         tools/t-dance.sh push
```
r10s: this needs to be  `push--do-this-only-from-master` as well then, or?

ghost: You are entirely correct. :see_no_evil:

r10s (approved):

PR #309, opened 2020-02-27
revert "desktop -> 1.0.0" because of encryption bug

r10s:
this reverts commit 1a98333941cc03ef0ff385259ec1b933e8c33393.

desktop 1.0.0 uses core25 that has an [encryption bug](https://github.com/deltachat/deltachat-core-rust/issues/1313) related to the new ecc keys. i think it is better to no offer this version for download currently (android+ios version using core25 are also not really released yet)


ghost:
The links work well on staging, so I'll merge it.

PR #316, opened 2020-03-05
one comment for a preview link

### LINKED ISSUE #315: If preview link isn't pushed to the checks at the bottom, comment
## Problem

For the first commit built by the [preview build action](https://github.com/deltachat/deltachat-pages/blob/master/.github/workflows/jekyll-preview.yml#L44), posting the link to the details always fails with the following API response:

```
{
  "message": "Not Found",
  "documentation_url": "https://developer.github.com/v3"
}
```

This is because we can't get the actual commit hash of that first commit; no idea why, it works for all subsequent commits.

## Solution

We used to post that link to the comments anyway. It was pretty annoying, as it was posting that with *every* commit. But if we only let the bot comment if the API doesn't allow posting it to the details, so only for the first commit, we only have one bot comment. It would even make it easier for first-time contributors, who maybe don't know that the preview link is hidden in the details.

ghost:
closes #315

### DIFF
--- a/.github/workflows/jekyll-preview.yml
+++ b/.github/workflows/jekyll-preview.yml
@@ -42,6 +42,7 @@ jobs:
         local: "_site/*"
         remote: "/var/www/html/staging/${{ steps.prepare.outputs.prid }}/"
     - name: "Post links to details"
+      id: details
       if: steps.prepare.outputs.uploadtoserver
       run: |
         # URLs for API connection and uploads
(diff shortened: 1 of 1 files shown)


ghost:
Starting with the second commit it's not commented anymore, but instead posted to the checks. Interestingly, even when you force-push a changed first commit. So I think this proves that it works.

Simon-Laux:
nice, having this fallback.

Simon-Laux (approved):

PR #322, opened 2020-03-12
tweak blog overview

### LINKED ISSUE #321: blog overview is pretty large
the [blog overview](https://delta.chat/en/blog) is meanwhile pretty large, currently about **7mb** - and growing.

the large size was introduced by showing the images in the blogposts at https://github.com/deltachat/deltachat-pages/pull/282 - before that, the overview was only about **30kb**.

we can handle this issue by 

1. adding a paging thingie, jekyll seems to support things there, i never dived into that

2. showing only the first 10-or-so messages and load the next 10-or-so or so messages when scrolling down (as in the forum)

3. load images only if they come to view

4. show images only for the first 10-or-so messages - this would also mitigate the problem that several images are a bit weird.

further alternatives would be to remove the images or to ignore the issue. or to use more optimized images - but after all experience, this won't work in practice, creating a blog post is already hard enough :) maybe the overview can create thumbnails automatically, but this is some additional effort.

1.+2. would sacrifice easy on-site-search and 2.+3. would require js afaik, so a not-so-good-solution

i think 4. should be pretty easy to do, just had a look, checking the index in the look at https://github.com/deltachat/deltachat-pages/blob/master/_includes/blog.html#L2 should do the job. probably a one-line-change.

r10s:
- do not show wide images with weird excerpts, eg. the upper 10% or an image, esp. on smaller screens
- hand-optimized the preview-images for the most recent 10 posts
- do not show images for older posts, optimizing all seems too time-consuming and also this results in several mb for the blob-overview, last 7mb, this closes #321


_includes/blog.html
```diff
@@ -1,5 +1,7 @@
+{% assign index = 0 %}
 {% for post in site.posts %}
-{% if post.image %}
+{% assign index = index | plus:1 %}
```
r10s, line 3: what a weird syntax ...

Simon-Laux (approved):

PR #323, opened 2020-03-15
tweak download page

r10s:
this pr cleanups the download page, and adds some missing badges.

before/after:

Screen Shot 2020-03-16 at 10 29 52 Screen Shot 2020-03-16 at 10 30 01


Simon-Laux:
why did you switch the flathub svg for an png?

_sass/_downloadpage.scss
```diff
@@ -1,82 +1,54 @@
...
-}
 
 
-.download-content {
-    display: block;
-    padding: 0 4pt;
+div.download-content {
 
-    // #recommend>.box {
+    /* mobile first */
```
Simon-Laux: for that it looks worse than before IMHO
Screen Shot 2020-03-16 at 13 20 18
The platform right is not intuitive as we read LTR and not RTL

_sass/_downloadpage.scss
```diff
@@ -1,82 +1,54 @@
...
         flex-direction: row;
         flex-wrap: wrap;
         align-items: center;
+        margin: 1em 0;
     }
 
-    .box>.buttons>a:not(.img-badge) {
-        background: var(--dl-button-bg);
+    div.box div.buttons a:not(.img-badge) {
+        background: #111;
```
Simon-Laux, line 34: should be 
```suggestion
        background: #000;
```
if you wanna copy the look of the image badges.

_sass/_downloadpage.scss
```diff
@@ -1,82 +1,54 @@
...
         color: white;
         font-weight: bold;
         padding: 5px;
         margin: 3px;
         text-decoration: none;
-        font-size: 18px;
+        font-size: 16px;
         height: max-content;
-        border-radius: var(--dl-button-roundness);
+        border: 2px solid #444;
```
Simon-Laux, line 43:
```suggestion
        border: 1.1px solid #888;
```

_sass/_downloadpage.scss
```diff
@@ -1,82 +1,54 @@
...
+        border-radius: 5px;
         line-height: 1.23;
+        min-width: 144px;
     }
 
-    .box>.buttons>a.img-badge {
+    div.box div.buttons a.img-badge {
```
Simon-Laux, line 49: why did you remove the `>` signs? do you want to allow for more flexibility?
in the case you also want to go for readability, you could also write it like:
```scss
div.box {
 div.buttons {
  a.img-badge {
  }
 }
}
```

_sass/_downloadpage.scss
r10s: i think it is okay, the platform is already implied very much by the icons

r10s: wrt image: the reason for the change was the wrong aspect ratio of the flathub image.

it is a png, but it is not worse than before, https://staging.delta.chat/323/assets/badges/flathub.png

the issue is more that we leave it up to the browser to do the scaling, might be that svg and png are handled differently.  maybe we should use prescaled images, but this can be done another time in another pr.

_sass/_downloadpage.scss
r10s: yip, for readability - and also as the constraint "direct child" is not needed. nesting css is only possible in sass and i do not want to depend on that.

_sass/_downloadpage.scss
r10s: plain black is too much here, i tried that

_sass/_downloadpage.scss
```diff
@@ -1,82 +1,54 @@
...
+    div.box div.buttons a:not(.img-badge) {
+        background: #111;
         color: white;
         font-weight: bold;
         padding: 5px;
         margin: 3px;
         text-decoration: none;
-        font-size: 18px;
+        font-size: 16px;
         height: max-content;
```
Simon-Laux:
```suggestion
        height: 48px;
        box-sizing: border-box;
```

_sass/_downloadpage.scss
r10s: let's do not discuss things pixel-wise :) esp. as the plan is to use "real images".

Simon-Laux:
- This "get it on" implies some kind of appstore in my eyes. (windows, mac os).
- I don't agree replacing `GNU+Linux` with just `Linux`. 
- I don't see an advantage in removing the flathub command, as it can save time for experienced users. also the moving of the arch command there doesn't look good. Then you could just use an badge image for arch, too: 'get it on AUR'.

_sass/_downloadpage.scss
Simon-Laux: I thought you want to replicate the other badges :thinking: ?

Simon-Laux:
This modern windows logo is not colored, its either light blue or white. When doing such things we should keep the logo guidelines in mind.
Also now that that I think about it we might need a notice that the logos are belonging to the specified companies or sth. like that?

Simon-Laux:
I like having the source code an testing version bellow the buttons.

_sass/_downloadpage.scss
Simon-Laux: how about moving the plaform names on the left site rotated by -90 degrees?:
76742330-3bc5b700-6771-11ea-9686-7f70ad9d4cfa

r10s:
> This modern windows logo is not colored, its either light blue or white

this "get it on" is very widely used and is offered from microsoft this way at https://developer.microsoft.com/en-us/store/badges/ - colored.
for shops, microsoft uses the term "get it from microsoft"

> I don't agree replacing GNU+Linux with just Linux.

is it about the dispute between members of the free software community and open-source software community? I do not think many users will care about that. but i can change it back if you do :)

> I don't see an advantage in removing the flathub command, as it can save time for experienced users. also the moving of the arch command there doesn't look good. Then you could just use an badge image for arch, too: 'get it on AUR'.

i was a bit unsure about that, esp. i did not like the weird mix of command, brackets and links.
so i will add a AUR download button and re-add and streamline the commands as well.

Simon-Laux:
> this "get it on" is very widely used and is offered from microsoft this way at https://developer.microsoft.com/en-us/store/badges/ - colored.

that's a different logo.

> I do not think many users will care about that

some nerds will care and it gnu linux is a different thing than android linux for example.

> i did not like the weird mix of command, brackets and links.

I know it is a challenge to make it look good. Commands are a huge time-saver and those who can use them appreciate them.

r10s:
> that's a different logo.

but it is colored :) but anyway, i think it is an overall improvement, ppl will recognise the logo. which logo would you use?

ghost:
the one from https://developer.microsoft.com/en-us/store/badges/ ? ^^

ghost:
I really like the tweaks btw. I'm still looking at the exact changes, but I don't think I have changes to request.

r10s:
> the one from https://developer.microsoft.com/en-us/store/badges/ ? 

they say "download from microsoft", we can use them if we're finally in the store.

currently, the "get it on windows" is more correct, it is used that way on may sites, it uses the same colored windows logo.

Simon-Laux:
> it is used that way on may sites,

example please.

~~I would use a white windows logo instead~~ edit: better don't use the logo at all as it might be violating their trade mark thingie and we don't really need it, appimage and debian buttons look just fine. Also having a distinction what comes from our site and what is an external link is good IMO.
~~Also please change the text, as "get it on windows" implies an external site or appstore and not a simple exe download IMO.~~
"Get for Windows", "Download for Windows" are better IMHO.
And for mac please mention that it is a dmg file or also say that you get it 'for' not 'on'

_includes/download-boxes.html
```diff
@@ -81,6 +65,23 @@ <h1 id="noscript-headline">{%if tx.menu.download != ""%}{{ tx.menu.download }}{%
                 <a href="https://download.delta.chat/desktop/{{VERSION_DESKTOP}}/DeltaChat%20{{VERSION_DESKTOP}}.exe">Get portable version (experimental)</a>
             </div>
         </div>
+        <div class="box" id="linux">
+            <div class="title">GNU/Linux</div>
+            <div class="buttons">
+                <a href="https://aur.archlinux.org/packages/deltachat-desktop-git/"><small>Get AUR for</small> Arch</a>
```
Simon-Laux:
```suggestion
                <a href="https://aur.archlinux.org/packages/deltachat-desktop-git/"><small>Build from</small>AUR</a>
```

r10s:
> example please.

https://joplinapp.org, https://ready2order.com/en/downloads/windows/ - but there are many more. i am sure you can find them yourself. there are also many white-window-buttons with "get it on", however, this is nitpicking, also i think the colored windows flag is still more known.

also, this problem will be gone when we're in the microsoft store, i think the exe-download will get smaller as for the android-apk.

> better don't use the logo at all as it might be violating their trade mark thingie 

i do not think, that using the windows logo for a download link causes problems - do you have any example where this was the case in the past?

> and we don't really need it

i disagree. the logos and buttons are a very easy to recognize element for the users.

btw. i would also like to have buttons for arch, debian/ubuntu and appimage -  but i did not find them at the first glace - we can optimize that later.

_includes/download-boxes.html
r10s: k, if this is what AUR is doing :)

Simon-Laux:
https://play.google.com/intl/en_us/badges/
they have strict guidelines, not only google.
https://www.microsoft.com/en-us/legal/intellectualproperty/trademarks/usage/general.aspx
https://query.prod.cms.rt.microsoft.com/cms/api/am/binary/RE1voQq <- guidelines for windows logos.

> the logos and buttons are a very easy to recognize element for the users.

then having iPhone & iPad written next to iOS is useful too... enough nitpicking...
We could integrate a symbol into the platform name like https://fontawesome.com/icons/windows?style=brands
https://fontawesome.com/icons/linux?style=brands
https://fontawesome.com/icons/apple?style=brands

it is done that way at https://keepassxc.org/download/

_sass/_downloadpage.scss
r10s: hm, i do not like that, also this rotated text is still causing problems here and there and is not widely supported, see eg. https://caniuse.com/#search=rotate at least not in this pr.

i want to have a look that is not too different from rest of the website and i want to have pretty simple markup. maybe we want to sth. as for http://getfirefox.com, so showing the remommendation much more outstanding. but this is also out of scope.

Simon-Laux: can you try moving the name to the left then? that should look better or back at the top of the boxes.
edit: moving the text to the left that the buttons are shown on the right at the end of the line requires some more css magic than just setting the float to left. So I vote for moving the title back to the top.

r10s:
i think this fontawesome cost money, requires registration or so, i did not dived into that, but it seems to be more complicated as a png image - and it does not solve the trademark question.

Simon-Laux:
> i think this fontawesome cost money

wrong, it only costs money if you want all their icons. and a registration is also not needed at all.
Anyways I don't say we need to use font-awesome, I just refer to the possibility of using icons.
We can outsource some of the responsibility to the icon maker, that was my thought here.

r10s:
might be that there are free plans, but going to https://fontawesome.com/start, the first they ask me is an email address. for just a png, this seems to be out of scope. so, this is out of scope of this pr, also as it does not help on trademark and logo guidlines.

Simon-Laux:
scroll down a tiny bit and there is an download link.... anyways that sub discussion doesn't bring us forward, please read my edit on my previous comment.

r10s:
> please read my edit on my previous comment.

what was the edit? can you please rephrase? is it about icons? i know, this can be handy, but for this pr, this is out of scope. also, as said, this won't help for the remaining question about trademark and logo guidelines.

Simon-Laux:
Let's discuss this in a short av @r10s, this comment storm is inefficient.

ghost:
Sounds good, I would join in as well :)

r10s:
sure - but what is left to discuss?

Simon-Laux (approved):

PR #326, opened 2020-03-21
update translations

r10s:


ghost:
I think those syntax errors are none^^ https://www.iwillteachyoualanguage.com/learn/french/french-tips/french-punctuation

ghost:
french is weird

r10s:
it is a "not the same space that we use in English; it is, “une espace insécable,” or a non-breaking space, but that’s beside the point" - ~~so using no space seems to be better than using a standard-space here.~~ 

~~but, however, all i did is to make it consistent. i fixed only some very few places, the very most places were not using a space before the question mark.~~

EDIT: thinking it over, it is probably better to add the space where it is missing in the help faq, just browsed through some french pages, they also do it that way ...

ghost:
Cool :) I take it you made the changes on transifex, not only on GitHub? As we don't want them to be overwritten by the next tx pull?

r10s:
yip, i did the changes on tranxifex.

@lefherz parles tu français? would be great to have the help translated to get it in-app :)

ghost:
No, unfortunately not - I just heard of this rule once. :7

PR #327, opened 2020-03-23
update desktop version to 1.1.1

Simon-Laux:


r10s (approved):

_includes/download-boxes.html
```diff
@@ -74,8 +74,8 @@ <h1 id="noscript-headline">{%if tx.menu.download != ""%}{{ tx.menu.download }}{%
                 <a href="https://flathub.org/apps/details/chat.delta.desktop" class="img-badge">
                     <img src="../assets/badges/flathub.png" alt="{%if tx.download.flathub != ""%}{{ tx.download.flathub }}{%else%}{{ txEn.download.flathub }}{%endif%}" />
                 </a>
-                <a href="https://download.delta.chat/desktop/{{VERSION_DESKTOP}}/deltachat-desktop_{{VERSION_DESKTOP}}_amd64.deb"><small>Get .deb for</small> Debian/Ubuntu</a>
-                <a href="https://download.delta.chat/desktop/{{VERSION_DESKTOP}}/DeltaChat-{{VERSION_DESKTOP}}.AppImage"><small>Get Universal</small> AppImage</a>
+                <a href="https://download.delta.chat/desktop/v{{VERSION_DESKTOP}}/deltachat-desktop_{{VERSION_DESKTOP}}_amd64.deb"><small>Get .deb for</small> Debian/Ubuntu</a>
+                <a href="https://download.delta.chat/desktop/v{{VERSION_DESKTOP}}/DeltaChat-{{VERSION_DESKTOP}}.AppImage"><small>Get Universal</small> AppImage</a>
```
ghost, line 78: Is this the current upload location of the release script? It would be nice if we wouldn't have to change these lines every release, and if it would be consistent.

r10s: imu, the folder name was prefixed by a `v` - for whatever reason. i do think, that these lines need adaption on every release, but @Simon-Laux probably knows better.

Simon-Laux: no they don't need adaption every time.
Anyway don't worry about it as the fact how often we need to change it is not important at all. just point to the right folder and done.

r10s: i wanted to write "i do not think, that these lines need adaption on every release" :) but id did not :)

PR #333, opened 2020-03-30
update desktop version to 1.2.0

Jikstra:
Didn't test the releases yet as i'm currently only on the laptop. @Simon-Laux can you maybe test the .dmg via vnc?


r10s:
i updated my mac to 1.2.0, works as a first glance.

in general, however, i would recommend to post updates to some testing channels first - eg. the "Testing Delta Chat" group and the forum - but also irc.

PR #334, opened 2020-03-31
draft new multi-tool needfinding report

hpk42:


r10s:
nice :) some technical suggestions:

- i would add the drawing also to he page. we have done this everywhere in the past, and this is also about recognizong that you're on the correct post (kind of "enlarge"). also we currently do not show the preview images for older posts and the post may be linked from somewhere else (twitter) - it would be a pity to not get the drawing in these cases. right-align before the first paragraph should do the job.

- if would add a paragraph somewhere so that the preview text is a bit shorter, maybe before "Supporting organizational use"

- also the black border looks a bit to a funeral card to me, might be a german thing, but anyway :)

EDIT: i already did the changes, shall i commit them directly to this pr?

r10s:
i made https://github.com/deltachat/deltachat-pages/pull/335 that can be merged into this pr.

PR #335, opened 2020-03-31
tweak 2020 needfinding

r10s:
some tweaks as suggested in the original pr #334, should me merged into #334


ghost:
Great :) personally I would make the link to the report appear above the image of its frontpage, for the people who don't scroll past the image.

r10s:
well, ppl who do no scroll have a problem with our page anyway :)

PR #337, opened 2020-04-12
add information about minimal requirements

r10s:
this pr adds a very short, condensed line about the minimal requirements across all versions.

i think the "Changelog" section is the closest match and not that wrong,
adding it directly below the boxes looks bad
and i do not want to add an extra headline,
there is no need to make this too much outstanding, ppl will find it.
for the same reason, i do not want to add it directly to the boxes.

wrt desktop: i copied the information from: https://www.electronjs.org/docs/tutorial/support

### DIFF
--- a/en/download.md
+++ b/en/download.md
@@ -11,6 +11,14 @@ downloads: true
 * [iOS](https://github.com/deltachat/deltachat-ios/blob/master/CHANGELOG.md)
 * [Core](https://github.com/deltachat/deltachat-core-rust/blob/master/CHANGELOG.md)
 
+The desktop versions do not require Delta Chat to be installed on a phone.
+
+Minimal requirements:
+Android 4.1 Jelly Bean
+or iOS 11, iPhone 5s or iPad 5/Air/Mini
+or Windows 7, macOS 10.10 Yosemite, Ubuntu 12.04, Fedora 21 or Debian 8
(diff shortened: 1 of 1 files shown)


en/download.md
```diff
@@ -11,6 +11,14 @@ downloads: true
 * [iOS](https://github.com/deltachat/deltachat-ios/blob/master/CHANGELOG.md)
 * [Core](https://github.com/deltachat/deltachat-core-rust/blob/master/CHANGELOG.md)
 
+The desktop versions do not require Delta Chat to be installed on a phone.
+
+Minimal requirements:
+Android 4.1 Jelly Bean
+or iOS 11, iPhone 5s or iPad 5/Air/Mini
+or Windows 7, macOS 10.10 Yosemite, Ubuntu 12.04, Fedora 21 or Debian 8
+or compatible systems.
```
Simon-Laux, line 20: I would add line breaks here otherwise this is collapsed into a single paragraph.

Simon-Laux (approved):
otherwise LGTM

en/download.md
r10s: i know, but this is intended. i do not want to make this take too much space on the page, there are other important information. i think, if ppl want to know some details, they will find their system in these lines.

PR #342, opened 2020-05-02
replace Friendica with Mastodon, add social links to footer

ghost:
The Friendica account doesn't seem to work at all anymore :/ But a good opportunity to verify our Mastodon account, and make the social links a bit more visible on the website. Nobody scrolls down to the bottom of the blogposts anyway^^


hpk42:
thanks for caring for mastodon btw! :)

r10s:
yeah, thanks a lot, now i also have an entry point ;)

PR #346, opened 2020-05-10
Update deltachat-desktop to v1.3.4

Jikstra:


hpk42:
now that android/core needed to be updated (there was a problem with mails that do not have a proper message-id format) it probably makes sense to wait and do a 1.3.2. @r10s is packaging up an android 1.6.0 with the fixed core.

Jikstra:
@hpk42 done, shall we put it into the testing channels?

r10s (approved):

PR #347, opened 2020-05-13
merge gdpr-website into gdpr page

hpk42:
google complained that our home page uses gdpr-website but the android app references "gdpr".

This PR merges the two pages into one document to avoid the confusion but only with the english version.  I think this has to go through a transifex-round and i cannot change the DE/ES/ etc. pages directly, right?


ghost:
Not the best solution, using an old container for this, but for now it should suffice... I will open an issue to fix this otherwise in the long term.

For now, the preview is here: https://staging.delta.chat/347/en/gdpr

en/gdpr-website.md
```diff
@@ -3,62 +3,5 @@ title: Privacy policy website
...
-cancellation and, if necessary, restriction of the processing. You can revoke
-the declaration of consent of the mailing list at any time. You have the right
-to receive the data in a structured common and machine-readable format and the
-right to complain to a supervisory authority.
-
-## Amendment of the privacy policy
-
-We reserve the right to update the present data protection regulations in view
-of the constantly changing legal and technical situation.
+This page has been moved to the "privacy statement on apps and websites" (gdpr).
```
r10s, line 6: i would add a clickable link

en/gdpr.md
```diff
@@ -1,17 +1,17 @@
 ---
-title: GDPR and Privacy
+title: Privacy for Delta Chat App and Websites
 lang: en
 ---
 
-# Note on privacy and GDPR 
+# Privacy Policy Delta Chat Apps
```
r10s, line 6: h1 and h2 have the same layout it seems, but here are to many top-level-headlines on the page.

i would suggest to use h2 (##) here and for "Privacy Policy for Website visits "  - for all other headlines, use h3 (###)

en/gdpr.md
```diff
@@ -51,3 +49,62 @@ with GDPR requirements.
 
 This statement reflects the General Data Protection Regulation (GDPR) as of 16.05.2018.
 
+# Privacy Policy for Website visits
```
r10s, line 52: also here h2 (##) - all other headlines h3 (###) - this is also what we are doing on the help pages.

en/gdpr-website.md
r10s: or a 301 redirect or so

PR #352, opened 2020-06-14
add notice for windows that you need the visual c++ redist

Simon-Laux:
We should have done this earlier, anyways this is just a band-aid until we figure out how to bundle/redistribute those dlls.


hpk42 (approved):

r10s:
@Simon-Laux good to have this hint, as long as the issue is not fixes. thanks for adding.

PR #355, opened 2020-07-03
add links to the stores and upgrade the download version to 1.4.x

Simon-Laux:
should we only merge this after we have the updated versions in the stores?
Update: I'll upload the store versions this evening, and it takes days before its in the store anyway so I'll merge this now.

### DIFF
--- a/_includes/download-boxes.html
+++ b/_includes/download-boxes.html
@@ -1,6 +1,6 @@
 {% assign txEn = site.data.lang['en']['en'] %}
 {% assign tx = site.data.lang[page.lang][page.lang] %}
-{% assign VERSION_DESKTOP = "1.3.4" %}
+{% assign VERSION_DESKTOP = "1.4.3" %}
 
 <div class="download-content">
     <div id="recommendation-section" hidden>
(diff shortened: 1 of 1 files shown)


r10s:
great! /o/  \o\  /o/ ... \o/

Simon-Laux:
though now there is the question how we can make the two options more distinct as its not apparent that it is a file download on the non-appstore button.

r10s:
maybe just use the textual buttons as for linux, having one apple and ios logo should be recognizale enough.

text could be "Get Setup.exe for Windows" and "Get .DMG for macOS"  or so.

btw, is the "you also need to install Microsoft's Visual C++ Redistributable" hint still required?

Simon-Laux:
> maybe just use the textual buttons

that's also what I thought of.

> text could be "Get Setup.exe for Windows" and "Get .DMG for macOS" or so.

we could probably even leave the platform/OS out of there, because its in a box labeled like the platform and next to the button with the logo?

> btw, is the "you also need to install Microsoft's Visual C++ Redistributable" hint still required?

no, but it still can't hurt... but we should weaken it instead, like if it doesn't work try installing vcredist. or sth along those lines?

PR #362, opened 2020-08-02
add some users voices

r10s:


Simon-Laux:
desktop isn't very accessible for blind people so I'm not sure whether we should include that quote right now.
https://github.com/deltachat/deltachat-desktop/issues/1791

Simon-Laux (approved):
LGTM

ghost (approved):

r10s:
> desktop isn't very accessible for blind people so I'm not sure whether we should include that quote right now.
> [deltachat/deltachat-desktop#1791](https://github.com/deltachat/deltachat-desktop/issues/1791)

the source-line says "about the Android version", so i'd say this is fine (iOS version is also accessible afaik)

hpk42:
good start!

however, the user stories/voices link appears a bit unmotivated at the end of the front page. 

also two quick "user voice" tweet links FYI: 

https://twitter.com/gozala/status/1281346020664729600

https://twitter.com/skadge/status/1276515066393878529

r10s:
> however, the user stories/voices link appears a bit unmotivated at the end of the front page.

thats true. i first added the voices directly on the homepage, however that appears not so nice, too much text on the start page. but we can probably make the link a bit nicer and/or frame it by a small headline or so. i'll add a commit, i will also incorporate the two links you provided - thanks!

PR #367, opened 2020-08-05
alternative blog preview

r10s:
due to a lack of a paging functionality
(which i personally do not miss ;)
and to keep the overall page size small,
we did not add an image for the 10th oldest blob post.

this results in some "text desert"
that is not really nice and makes the page
a bit unclear when scolling down.
too many information :)

this commit changes that by just adding the clickable title
for older posts.
moreover, the "new posts" are increased to 12 (so, roughly a year ;)

by the decreased amount of text,
the page is now smaller, both, in size and visually.
also, this hides the fact that old blog posts are not always so nice
(eg. they do not have a preview image, so paging would not be nice)

i find the blog much nicer this way,
see preview.

### DIFF
--- a/_includes/blog.html
+++ b/_includes/blog.html
@@ -1,17 +1,13 @@
 {% assign index = 0 %}
 {% for post in site.posts %}
 {% assign index = index | plus:1 %}
-{% if post.image and index <= 10 %}
+{% if post.image and index <= 12 %}
 <div class="post-box">
   <a href="{{ site.baseurl }}{{ post.url }}">
     <div class="featured-image">
       <img src="{{ post.image }}" />
     </div>
-{% else %}
-<div>
-  <a href="{{ site.baseurl }}{{ post.url }}">
-{% endif %}
-  <h2>{{ post.title }}</h2>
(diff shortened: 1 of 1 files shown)


Simon-Laux (approved):

ghost:
Hm, I don't really see how the new solution is less of a text desert :7 I think I'd rather add images for old posts as well, I think I even did once, maybe that branch is still somewhere. It was rejected, because the images were not very good, I think.

r10s:
> I don't really see how the new solution is less of a text desert

it is just far less text.

> I think I'd rather add images for old posts as well

that quite some work, i do not think this is worth the effort.

and when we did it the"the simple way" by just showing the the blog overview was several 10mb large - several posts have quite large images where others do not have images at all.

the approach suggested in this pr is to show the most recent posts "full" - and things "after the fold", that are typically on page 2 or so, by just their headlines.

however, i have no strong opinion on that, this pr just came into my mind, i would also be happy with closing it unmerged :)

ghost:
Ah, you're right, I completely forgot about the loading time dilemma. In that case, let's just merge it.

PR #373, opened 2020-09-13
update translations, run bundle update

r10s:
see commit messages for details.


Gemfile.lock
```diff
@@ -4,32 +4,32 @@ GEM
     addressable (2.7.0)
       public_suffix (>= 2.0.2, < 5.0)
     colorator (1.1.0)
-    concurrent-ruby (1.1.5)
+    concurrent-ruby (1.1.7)
```
Simon-Laux, line 7: you don't mention the update of gem dependencies, was this intentional?

Simon-Laux (approved):

Gemfile.lock
r10s: good catch - it sneaked in more or less accidentally, my setup was broken and i did various python/ruby/node/whatever updates 🧐

i just had a closer look:

with the older jekyll version, recent ruby complains with tons of warnings as 

```
./deltachat-pages/vendor/bundle/ruby/2.7.0/gems/jekyll-4.0.0/lib/jekyll/convertible.rb:41: warning: Using the last argument as keyword parameters is deprecated
```

and i had to update the ruby version for $reasons on my system - however, as it seem to work on delta.chat (preview is fine) as well and we have to upgrade at some point anyway, it is probably fine to merge that in now.

if we encounter problems, we can revert.

however, i will split the language commit from the jekyll-update commit.

PR #375, opened 2020-09-20
tweak help

### LINKED ISSUE #254: remove android-specific stuff from help.md
help.md should be generic and should not refer to how things can be done on a specific os.

in general, this is true, however, the few "click-paths" should be removed or adapted.

r10s:
- use wording so that that fits on all os targeted (we are using the same help on all os as well as on the webpage)
- fix some meanwhile invalid things

closes #254

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -46,7 +46,7 @@ Delta Chat automatically shows:
 - Messages from contacts **contacted by you**
 - **Replies** to messages sent by you
 
-Other messages do not appear automatically, and are to be found in the **Contact requests**. If desired a chat can be started from there.
+Other messages do not appear automatically, and are to be found in **Contact requests**. If desired a chat can be started from there.
 
 
 ### What about spam?
--- a/en/help.md
+++ b/en/help.md
@@ -244,7 +244,7 @@ If you are not near the chat partner, you can check the status manually in the "
 
 ### How can I check the encryption of messages?
 
-- If a little **padlock** is shown beside a message, this denotes the message is end-to-end-encrypted and is sent from the given sender.
+- A little **padlock** shown beside a message denotes whether the message is end-to-end-encrypted from from the given sender.
 
 - If there is **no padlock**, the message is usually transported unencrypted e.g. because you or the sender have turned off end-to-end-encryption, or the sender uses an app without support for end-to-end-encryption.
 
--- a/en/help.md
+++ b/en/help.md
@@ -100,7 +100,7 @@ about feedback which makes the app more secure for all of our users.
 
 ### Which permissions does Delta Chat need?
 
-Depending on the used operating system,
+Depending on the operating system in use,
 you may be asked to grant permissions to the app.
 This is what Delta Chat does with these permissions:
 
--- a/en/help.md
+++ b/en/help.md
@@ -316,7 +316,7 @@ If you are not near the chat partner, you can check the status manually in the "
 
 - Yes. The best way is to send an Autocrypt Setup Message from the other e-mail client. Look for something like **Start Autocrypt Setup Transfer** in the settings of the other client and follow the instructions shown there.
 
-- Alternatively, you can import the key manually at "Settings" or "Advanced settings" and then "Import secret keys". Caution: Make sure the key is not protected by a password, or remove the password beforehand.
+- Alternatively, you can import the key manually in "Settings" or "Advanced settings" and then "Import secret keys". Caution: Make sure the key is not protected by a password, or remove the password beforehand.
 
 If you don't have a key or don't even know you would need one - don't worry: Delta Chat generates one as needed, you don't have to hit a button for it. 
 
--- a/en/help.md
+++ b/en/help.md
@@ -210,7 +210,7 @@ This is what Delta Chat does with these permissions:
   messages are not encrypted until an Autocrypt-compliant app is available again. 
 
 - If you want to rather avoid end-to-end-encrypted e-mails by default, 
-  use the corresponding Autocrypt setting at "Settings" or "Advanced settings".
+  use the corresponding Autocrypt setting in "Settings" or "Advanced settings".
 
 
 ### If end-to-end-encryption is not available, is the connection not encrypted at all?
--- a/en/help.md
+++ b/en/help.md
@@ -46,7 +46,7 @@ Delta Chat automatically shows:
 - Messages from contacts **contacted by you**
 - **Replies** to messages sent by you
 
-Other messages do not appear automatically, and are to be found in the main menu at **Contact requests**. If desired a chat can be started from there.
+Other messages do not appear automatically, and are to be found in the **Contact requests**. If desired a chat can be started from there.
 
 
 ### What about spam?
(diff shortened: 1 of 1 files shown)


en/help.md
```diff
@@ -127,8 +131,8 @@ about feedback which makes the app more secure for all of our users.
 
 ### Creation of a group
 
-- Select **Add group** from the "Sandwich menu" in the upper right corner of the chat overview.
-- On the following screen, select the **group members** and hit the check mark in the upper right corner. After that, you can define a **group name**.
+- Select **New chat** and then **New group** from the main menu or hit the "+" button on Android.
```
Simon-Laux: maybe we could add an inline image/icon of the main menu symbol.

r10s: yeah, maybe, but i would not do that as part of this pr. alone this line is already a great improvement and it might be that it is totally fine this way.

also, such things make maintainability a lot harder - and we already needed 8 months to fix a simple line ;)

r10s: for desktop, we could also say  "sandwich menu" instead of "main menu" maybe this is better, although this is not 100% correct.

r10s: there is also a unicode character: ⋮

r10s: maybe "Select **New chat** and then **New group** from the menu in the upper right corner or hit the corresponding button on Android/iOS"

Simon-Laux: I have not much opinion on this, maybe just add the unicode caracter and call it a day, but I think ios user interface is different isn't it?

r10s: yip, ios is different. but "upper right corner" will do the job there. the icon we are using is used in tons of other apps. i do not expect much problems there. i also think "upper right corner" is sufficient for desktop. let's try that out, we can change it later anyway :)

en/help.md
```diff
@@ -46,7 +46,7 @@ Delta Chat automatically shows:
 - Messages from contacts **contacted by you**
 - **Replies** to messages sent by you
 
-Other messages do not appear automatically, and are to be found in the main menu at **Contact requests**. If desired a chat can be started from there.
+Other messages do not appear automatically, and are to be found in the **Contact requests**. If desired a chat can be started from there.
```
comradekingu:
```suggestion
Other messages do not appear automatically, and are to be found in **Contact requests**. If desired a chat can be started from there.
```
(?)

en/help.md
```diff
@@ -98,7 +98,11 @@ if you want to verify that your credentials are handled securely. We are happy
 about feedback which makes the app more secure for all of our users.
 
 
-### Which Android permissions does Delta Chat need?
+### Which permissions does Delta Chat need?
+
+Depending on the used operating system,
```
comradekingu:
```suggestion
Depending on the operating system in use,
```

en/help.md
```diff
@@ -206,7 +210,7 @@ about feedback which makes the app more secure for all of our users.
   messages are not encrypted until an Autocrypt-compliant app is available again. 
 
 - If you want to rather avoid end-to-end-encrypted e-mails by default, 
-  use the corresponding Autocrypt setting in "Settings / Advanced settings".
+  use the corresponding Autocrypt setting at "Settings" or "Advanced settings".
```
comradekingu: why "at"?

en/help.md
```diff
@@ -220,14 +224,17 @@ about feedback which makes the app more secure for all of our users.
 
 ### How can I verify cryptographic status with a sender? 
 
-The user's profile shows some additional info:
+If you are physically nearby the chat partner:
```
comradekingu:
```suggestion
If you are within immediate distance of the chat partner:
```

en/help.md
```diff
@@ -220,14 +224,17 @@ about feedback which makes the app more secure for all of our users.
...
 
-- Tap "QR Invite code" on Android and then use the "Scan QR code"
-  button on another device to scan this code. If both devices are online,
+- Select **QR Invite code** on one device and then **Scan QR code**
+  on the other one and scan the code. If both devices are online,
   they will introduce a chat channel with each-other (if it doesn't exist already) 
   and the encryption keys will also be verified.  Both will see a
   "sender verified" system message in their 1:1 chat. 
 
+If you are not nearby the chat partner, you can check the status manually in the "Encryption" dialog
```
comradekingu:
```suggestion
If you are not near the chat partner, you can check the status manually in the "Encryption" dialog
```

en/help.md
```diff
@@ -237,7 +244,7 @@ The user's profile shows some additional info:
 
 ### How can I check the encryption of messages?
 
-- If a little **padlock** is shown beside a message, this denotes the message is end-to-end-encrypted _and_ is sent from the given sender _and_ that your answer will be end-to-end-encrypted as well.
+- If a little **padlock** is shown beside a message, this denotes the message is end-to-end-encrypted and is sent from the given sender.
```
comradekingu:
```suggestion
- A little **padlock** shown beside a message denotes whether the message is end-to-end-encrypted from from the given sender.
```

en/help.md
```diff
@@ -309,7 +316,7 @@ The user's profile shows some additional info:
 
 - Yes. The best way is to send an Autocrypt Setup Message from the other e-mail client. Look for something like **Start Autocrypt Setup Transfer** in the settings of the other client and follow the instructions shown there.
 
-- Alternatively, you can import the key manually in "Advanced settings / Manage private keys". Caution: Make sure the key is not protected by a password, or remove the password beforehand.
+- Alternatively, you can import the key manually at "Settings" or "Advanced settings" and then "Import secret keys". Caution: Make sure the key is not protected by a password, or remove the password beforehand.
```
comradekingu: why "at"? "from" is good(?)

en/help.md
```diff
@@ -350,7 +357,7 @@ For other programs, you can find a solution online.
 If you want to use the **same account** on different devices, you should export
 a backup from the old device, and import it into the new device:
 
-- On the old device, choose "Settings / Chats and media / Backup". Enter your
+- On the old device, go to "Settings" or "Settings / Chats and media" and then to "Backup". Enter your
   screen unlock PIN, pattern, or password. Then you can click on "Start
   Backup". This saves the backup file to your device. Now you have to transfer
   it to the other device somehow.
```
comradekingu, line 363:
```suggestion
  it to the other device in a secure manner.
```

en/help.md
r10s: yeah, that's better, thanks.

en/help.md
r10s: maybe right, but out of scope of this pr.

en/help.md
r10s: i thought that is more correctly, but no idea why :) however, as it was "in" before, i will just leave it this way.

```suggestion
- Alternatively, you can import the key manually in "Settings" or "Advanced settings" and then "Import secret keys". Caution: Make sure the key is not protected by a password, or remove the password beforehand.
```

en/help.md
r10s:
```suggestion
  use the corresponding Autocrypt setting in "Settings" or "Advanced settings".
```

en/help.md
comradekingu: I don't think the document isn't trying to give readers clues on how to do things right…

r10s:
@comradekingu thanks a lot for review! very helpful!

PR #382, opened 2020-10-24
update russian translation

r10s:

PR #383, opened 2020-10-24
add russian to offline help

r10s:
successor of #382 

this will make the russian faq/help, that is now also 100% translated for, available offline.


r10s:
merging as needed for ios/android releases.

PR #386, opened 2020-10-30
Update some English grammar in help.md

ninelima:


r10s (approved):
i think, this is an improvement, thanks a lot.

we hesitated merging this,  as updating the english source results in taking care of the translations - in this case, a re-translation seems not to be needed, i will try to take care of that now :)

PR #391, opened 2020-11-20
Swipe to reply blogpost fixed up

comradekingu:


r10s (approved):
i've overseen a bit, sorry for that. i am no english native - therefore, i think, the changes are better than the original text i did at that time :) at least, they sound better to me :)

as most changes affect grammar only, i think it is reasonable to merge that.

PR #402, opened 2021-01-22
Translate to Portuguese

rosano:


r10s:
@rosano thanks a lot for the translations!

as the file edited in this pr will be overwritten by the next translation-update from Transifex,
i've copied the strings manually to Transifex.

To see the result directly, i've also started translation-update from Transifex, things are in pr https://github.com/deltachat/deltachat-pages/pull/403 now and will be online soon.

if possible, for the future, it would be great if you can do the translations using the Transifex service, see https://www.transifex.com/delta-chat/public/ - you can log in with your github  account there :)

r10s:
it's online now, https://delta.chat/pt/, thanks again!

PR #409, opened 2021-02-05
split first paragraph of ios-blogpost

r10s:
the first paragraph is shown in the blog overview
and is a bit long there, also the "Note ..."
sentence does not make much sense in the overview.

with this change, the "Read on..." link will appear one sentence before,
the remaining first paragraph has a reasonable length,
similar to all the other posts then.


r10s:
hm, does not really make a big difference, but make the understanding of the second paragraph harder. closing this without merging

PR #413, opened 2021-02-07
fix empty-string-comparison and avoid empty <h1> tags

r10s:
came over that via https://github.com/Shopify/liquid/issues/223#issuecomment-20881188 - checking for an empty string seems to be a hard thing in jekyll :) not sure, if the `!= blank`thing has ever worked (though also recommended in the link above)


hpk42 (approved):
lgtm

PR #419, opened 2021-02-23
added video calls to FAQ

ghost:
Not sure if we should already put this in the FAQ - it is still an experimental feature, after all.

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -439,7 +439,7 @@ Otherwise, deleting messages or multi-device setups might not work properly.
 To enable video chats, go to the advanced settings -
 under "Experimental Features" you can enter a "Video Chat Instance". You can
 use different video chat platforms for this, e.g. Jitsi. To choose the main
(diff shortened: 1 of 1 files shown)


Simon-Laux:
what about instruction on the other platforms?

ghost:
I included Android - is the process the same on iOS?

r10s:
> Not sure if we should already put this in the FAQ - it is still an experimental feature, after all.

i'd say no.

maybe better put that to the howto-section in the forum for now?

the thing is still experimental, i assume, we would get more issues reported if we push ppl in the faq to it, eg. why "cannot i start calls in groups". i think, user expectations are far lower when it is flagged as "experimental" and user are not pushed to it.

there are some plans to streamline things after the next release (email compat), so maybe better wait until then.

also, ios does not have the feature yet, and if we add it there, it will probably go to the "paper clip" and can be used as an attachment (maybe that would be a nicer approach for android as well, it is in the main menu only because it was the simplest way to do iirc)

r10s:
k, closing this for now.

PR #424, opened 2021-03-13
preview links are in https://download.delta.chat/desktop/

r10s:


Simon-Laux:
I would rather call it "all desktop versions" instead of preview: there is an directory called preview at that location which contains highly unstable builds from pull request versions, we already had someone trying them out thinking they were real future versions.

generally if we want to have this site user-facing we might need to change structure/naming.

Jikstra:
> I would rather call it "all desktop versions" instead of preview: there is an directory called preview at that location which contains highly unstable builds from pull request versions, we already had someone trying them out thinking they were real future versions.
> 
> generally if we want to have this site user-facing we might need to change structure/naming.

I think that previews for our pr builds is just highly inaccurate. CI needs to get fixed, not the webpage.

PR #425, opened 2021-03-13
Update desktop to 1.15.3

Jikstra:


Simon-Laux:
let's rather do a 1.15.3 with the darktheme qr code scan problem fixed. as darktheme is now reliably detected and selected from the OS settings.

Simon-Laux (approved):

PR #434, opened 2021-04-16
add fr to local help generation

Simon-Laux:
and tell jekyll to ignore the optionional `venv` folder


r10s (approved):
nice! the [webpage is already updated](https://delta.chat/fr/help), and yes, it is great to the french translation info the offline help as well.

thanks a lot to the french translators 🙏

PR #435, opened 2021-04-28
update translations

r10s:


r10s:
just scanned over the german changes - great!

PR #442, opened 2021-05-07
fix post reference URL

adbenitez:
https://delta.chat/en/user-voices#company-notification-system-migrated has wrong link to forum post


Simon-Laux (approved):

r10s:
nb: similar to app-strings, changes [languages unequal to `en` will be overwritten on the next translation pull](https://github.com/deltachat/deltachat-pages/blob/master/README.md#directory-structure); these changes have to go to transifex to stay persistent.

i will fix that on transifex.

PR #447, opened 2021-05-18
help: not watching inbox skips mails by non-DC-users

### LINKED ISSUE #440: Watching the Inbox makes sense to turn off, if you have...
Hi,

IMHO the documentation about "Why can I choose not to watch the Inbox?" is missing something. I'm using a Mailcow mail server, that has a sieve filter to move all DeltaChat mails to the DeltaChat folder (see: https://mailcow.email/2021/01/14/back-to-the-future-with-delta-chat/ and https://github.com/mailcow/mailcow-dockerized/commit/3dece1a05c97c1b989b432bd0609710fef91e95c)

So I've turned off to watch the inbox. But this is only fine, as long as the chat partners are using the DeltaChat client, too. If an answer comes from a usual mail client, it of course does not have the Chat-Version header and therefore stays in the inbox. It never reaches DeltaChat then.

Therefore the "Watching the Inbox makes sense to turn off, if you have..." list should also have an entry:

- if you don't want to chat with non DeltaChat users

ghost:
closes #440

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -386,10 +386,11 @@ server-side rules. Not all providers support this, but with some you can move
 all mails with a "Chat-Version" header to the DeltaChat folder. Normally, this
 would be done by the Delta Chat app.
 
-Watching the Inbox makes sense to turn off, if you have both:
+Watching the Inbox makes sense to turn off, if:
 
-- enabled a server-side rule to move all messages with Chat-Version header to the DeltaChat folder, and
-- have set the "Show classic emails" setting to "no, chats only".
+- you don't want to communicate with people who don't use Delta Chat,
+- you have enabled a server-side rule to move all messages with Chat-Version header to the DeltaChat folder, and
+- you have set the "Show classic emails" setting to "no, chats only".
 
 In this case, Delta Chat doesn't need to watch the Inbox.
 


r10s (commented):
imu, by recent changes, the assumption 

_Inbox-watch-off == i do not want to communicate with non-deltas_

has become wrong as all folders are scanned from time to time anyway; so the "Show classic emails" switch is the only setting that prevents these mails from being shown; cc @Hocuri - can you confirm this?

that in mind, not sure if an rewording is needed at all.

Hocuri:
> so the "Show classic emails" switch is the only setting that prevents these mails from being shown

Yes, I can confirm. All folders that are not watched will be scanned from time to time instead. And, in fact, the long-term goal is to remove the `Watch * folder` setting and "smartly" figure out which folders to watch.

Wrt #440:

`"have set the "Show classic emails" setting to "no, chats only".` also implies `you don't want to chat with non DeltaChat users`. I mean, that's what "no, chats only" does, it hides non-DeltaChat messages. The wording is a little confusing I think, maybe one day we can rename "no, chats only" to "no, only DeltaChat messages" or similar.

Conclusion: I'd also say, close this PR and #440.

PR #449, opened 2021-05-21
update email-compat blog to new avaibility

r10s:


missytake (approved):

Simon-Laux:
I would rather merge this when desktop is in testing, which it isn't yet currently

PR #464, opened 2021-07-15
remove outdated push statement

### LINKED ISSUE #462: faq: wording of background notification service usage outdated
we technically use APN for the heartbeat-wakeup

r10s:
closes #462

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -444,9 +444,6 @@ Otherwise, deleting messages or multi-device setups might not work properly.
 - Receiving messages then can take minutes because both Android and iOS often
   stop Delta Chat from running in the background, and only wake it up
   occasionally. This artificial delay is usually worse on iOS than on Android.
-- Note that Delta Chat doesn't use Google Cloud Messaging (GCM) or the Apple
-  Push Notification Service (APNS), because this leads to user tracking and
-  central control which Delta Chat aims to avoid as much as feasible.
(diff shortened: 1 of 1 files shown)


missytake:
The "However" in the next sentence is still weird...

r10s:
> The "However" in the next sentence is still weird...

any suggestions?

i still find it somehow fitting, at least not more weird as before :)

en/help.md
```diff
@@ -444,9 +444,6 @@ Otherwise, deleting messages or multi-device setups might not work properly.
 - Receiving messages then can take minutes because both Android and iOS often
   stop Delta Chat from running in the background, and only wake it up
   occasionally. This artificial delay is usually worse on iOS than on Android.
-- Note that Delta Chat doesn't use Google Cloud Messaging (GCM) or the Apple
-  Push Notification Service (APNS), because this leads to user tracking and
-  central control which Delta Chat aims to avoid as much as feasible.
 - However, Android and iOS kill apps running in the background is a
```
missytake, line 447:
```suggestion
- That Android and iOS kill apps running in the background is a
```

missytake: something like this?

missytake (approved):
But all in all, I think we can also leave it like this, I don't have a reaaally good suggestion.

r10s:
yip, let's leave it as is for now. i think, at some point, the faq needs a larger update anyway.

Issue #466, opened 2021-07-30
Protonmail Bridge for SMTP/IMAP


fossdd:
I found that Protonmail offers a tool called [Protonmail Bridge](https://protonmail.com/bridge/) which should bridge the Protonmail API.

I don't know if that works, but may you can add it to `help.md`

r10s:
yip, this is known - and the information is already shown to the user on login and at https://providers.delta.chat/protonmail .

adding detailed information about special providers to the general faq seems to be too much.

fossdd:
Ah okay, agree. It's just that it looks more like it completly doesn't work with Protonmail.

r10s:
we can maybe tune down the statement in the help, sth, as "... they do not offer receiving mails via unrestricted IMAP." however, in a way it is correct already now, protonmail-server as such does not offer IMAP, this needs the bridge. also, it does still not work for android and ios, it only works for paying desktop users that do not want to go for multi-device. that is pretty broken :)

fossdd:
Yes. That would be nice

r10s:
i think this is sufficiently handled

PR #468, opened 2021-08-22
protonmail is indeed compatible with OpenPGP, just not with Autocrypt.

missytake:
We received some feedback that the FAQ is slightly incorrect - this should cover it :)

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -456,9 +456,9 @@ Otherwise, deleting messages or multi-device setups might not work properly.
   Chat; they do not offer receiving mails via IMAP.
 - In any case you can use Delta Chat to send Messages to people who use
   Protonmail, Tutanota, or Criptext. Those messages will not be End-to-End
-  encrypted, though. The End-to-End encryption those providers offer is only
-  working inside their platforms, and not compatible with Autocrypt, the
-  standard Delta Chat uses.
+  encrypted, though. The End-to-End encryption those providers offer is not 
+  compatible with [Autocrypt](https://autocrypt.org/), the standard Delta Chat
+  uses.
 - Delta Chat can e2e-encrypt through any e-mail provider with any
   [Autocrypt-enabled e-mail app](https://autocrypt.org/dev-status.html).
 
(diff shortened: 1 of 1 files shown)


missytake:
It might make sense to directly include #466 as well in this PR... maybe tomorrow^^

missytake:
btw, see https://staging.delta.chat/468/en/help#is-delta-chat-compatible-with-protonmail--tutanota--criptext for preview.

en/help.md
```diff
@@ -456,10 +456,15 @@ Otherwise, deleting messages or multi-device setups might not work properly.
...
 - In any case you can use Delta Chat to send Messages to people who use
   Protonmail, Tutanota, or Criptext. Those messages will not be End-to-End
-  encrypted, though. The End-to-End encryption those providers offer is only
-  working inside their platforms, and not compatible with anyone outside.
+  encrypted, though. The End-to-End encryption those providers offer is not 
+  compatible with [Autocrypt](https://autocrypt.org/), the standard Delta Chat
+  uses.
 - Delta Chat can e2e-encrypt through any e-mail provider with any
   [Autocrypt-enabled e-mail app](https://autocrypt.org/dev-status.html).
+- Users who pay for Protonmail Desktop might be able to use the IMAP Bridge for
```
r10s: i think, this is much too much text in an _f_aq (that also goes to offline help) for an provider that is actually flagged as being broken.

r10s: maybe say:
- No, you cannot simply log in to Delta Chat using your Protonmail, Tutanota or Criptext account; they do not support open standards for that.
- Yes, you can use Delta Chat to send Messages to people who use these providers. Again, due to a lack of supported standards, these mails are not end-to-end encrypted.

r10s:
just came over https://github.com/ProtonMail/proton-bridge/issues/220 - so, i think, treating protonmail as broken is even more reasonable.

missytake:
(thanks for pushing this^^)

r10s (approved):
still not super-happy with the framing, but it is more correct now, so let's move forward :)

PR #475, opened 2021-09-17
float image right

missytake:


r10s (commented):
on mobile devices, the link is broken into two parts now, this is a bit weird visually, however, i can also imagine other problems when parts of the same link are floating around somewhere else.

Screen Shot 2021-09-18 at 11 36 02


of course, this is fixable, however, even if technically correct i am not sure if i would like it more and if it is worth the effort. the image is the central thing in the post, not just a decorative element (as we're partly using images/screenshots in other posts)

so, i would just leave things as they are :) (or i missed the gist of the pr :)

missytake:
Yeah, it's non-optimal... the problem is, with the markdown layout we can only float the image to the right by integrating it into a paragraph. Within the status quo, download link is the caption of the image; this doesn't work when it's part of the paragraph (I couldn't think of a way to do this properly). So this PR would move the link downwards, making it stand alone.

I slightly prefer the float right optics, bjoern the status quo... @hpk42?

Simon-Laux:
maybe link can also be integrated into the floating part, but current version conveys some kind of logic: the paragraph explains it then there is the document, would be different with an unrelated image, but this time the image represents an the document/object that was talked about.
So I like the current version better too.

PR #472, opened 2021-09-11
tweak 1.22 blog post to desktop realities

r10s:
that should be merged when 1.22 desktop is actually out.

PR #478, opened 2021-09-22
bump desktop version to 1.22.1

Simon-Laux:
dmg works on Mac, maybe we should self check the other release binaries too before merging this.


r10s (approved):

r10s:
should probably merged together with https://github.com/deltachat/deltachat-pages/pull/472

PR #491, opened 2021-11-01
adapt help to reality

r10s:
adapt or remove statements that are no longer true that way.

this pr is not about refactoring/reworking help. it is only about fixing most obvious things that are just no longer true since the last updates.

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -49,13 +49,6 @@ Delta Chat automatically shows:
 Other messages do not appear automatically, and are to be found in **Contact requests**. If desired a chat can be started from there.
 
 
-### What about spam?
-
-- Messages in common spam folders are ignored and contained addresses are not considered as known contacts.
-- As messages from unknown contacts do not pop up automatically, normally, there is **no spam**.
-- However, if needed, you can **block** any contact.
-
-
 ### Does Delta Chat support images, videos and other attachments?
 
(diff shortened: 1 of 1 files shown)


en/help.md
```diff
@@ -49,13 +49,6 @@ Delta Chat automatically shows:
 Other messages do not appear automatically, and are to be found in **Contact requests**. If desired a chat can be started from there.
 
 
-### What about spam?
-
-- Messages in common spam folders are ignored and contained addresses are not considered as known contacts.
```
r10s, line 54: messages from spam folders are not ignored;  they are read and replies are picked up. however, this is nothing for an FAQ.

en/help.md
```diff
@@ -49,13 +49,6 @@ Delta Chat automatically shows:
...
-- As messages from unknown contacts do not pop up automatically, normally, there is **no spam**.
```
r10s, line 55: messages from unknown contacts may pop up as "Contact requests". this is already pointed out before.

en/help.md
```diff
@@ -49,13 +49,6 @@ Delta Chat automatically shows:
...
-- However, if needed, you can **block** any contact.
```
r10s, line 56: leaving this as the only point in "what about spam" seems unneeded. also i do not think that this is sth. for an FAQ, we do not discuss all options here, and "block" is meanwhile pretty much in the face.

adbenitez (approved):

PR #493, opened 2021-11-10
fix weird breaking in mobile GDPR-blog-layout

r10s:
on small and mobile screens, the image teared the text into two parts (instead of the text flowing around the image)

before/after:

Screen Shot 2021-11-10 at 20 46 14 Screen Shot 2021-11-10 at 20 46 08

i also made the image slightly larger, so that there is an "enhance" effect when coming from the list.

the text is not modified.


r10s:
just merged that in to fix the live-issue, we can do a subsequent pr if there are comments or so.

PR #495, opened 2021-11-11
fixed links and linked legal references to GDPR articles

Yettte:

PR #497, opened 2021-11-12
privacy policy: we don't use Android Firebase

missytake:
No idea how this slipped through the review.


hpk42 (approved):
it's not so bad and let's revive it when/if we go for android push notifications -- some folks have been asking for it and there are some phone models that do not work well regarding background receiving.

missytake:
Let's merge this after #495 to avoid confusion.

missytake:
rebased, looks fine, I'll just merge it.

PR #499, opened 2021-11-17
add a direct android download link (as on desktop)

r10s:
add a button for the apk as for all other platform links.
previously, the link pointed to https://download.delta.chat/android/
where the user has to figure out the correct file themself
(out of some dozend files :)

little drawback is that `VERSION_ANDROID` needs to be updated on releases, however compared to all the other things to do, this is probably fine (and also allows to change the download link back on issues :)

### DIFF
--- a/_includes/download-boxes.html
+++ b/_includes/download-boxes.html
@@ -1,6 +1,7 @@
 {% assign txEn = site.data.lang['en']['en'] %}
 {% assign tx = site.data.lang[page.lang][page.lang] %}
 {% assign VERSION_DESKTOP = "1.22.2" %}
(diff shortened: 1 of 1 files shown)


missytake (approved):
Wow I love this. It even reduces lines of code... genius.

Do we need to do anything like taking strings out of transifex or so? Does this happen automagically?

r10s:
> Do we need to do anything like taking strings out of transifex or so? Does this happen automagically?

tbh, i do not really know for .yml :) for the .md files, removed strings are removed from transifex on the next push, however.

but we can have a look at that at some subsequent pr, maybe there is even more in the .yml that is no longer needed ...

PR #502, opened 2021-12-17
clarify initial setup for translation

k2s:
There is typo on Transifex in word `verifiy`, locally it is `verify`, after `tx set` command config is updated and `t-dance.sh` script doesn't work corectly. So I added instruction to use sed command to fix it. Best option would be to fix the type on Transifex.

Also I added instruction not to reinit the .tx folder and to run the setup commands in tools folder instead in root of the repo.


r10s (approved):
thanks a lot for taking care!

PR #512, opened 2022-01-11
incorporated feedback from group and forum qr code howto

Yettte:


r10s (approved):
review is mainly about replaced asset.

i also did a quick side-by-side comparison of old and new pdf, however, not in detail - i assume, this is was already discussed in background channels :)

hpk42:
Screenshot from 2022-01-12 12-27-39
looks good apart from this minor detail in the screenshot where the sentence starts somewhere inside the graphic and it's "how" instead of "now".  This is not a show stopper so approving anyway (you can merge yourself now i think but could also fix this little issue and merge then, your choice).

hpk42 (approved):

Yettte:
I corrected the sentence, thanks for the feedback.

Yettte:
it's fixed, thanks for the feedback!

PR #521, opened 2022-02-17
mailbox.org doesn't need preparation, actually

missytake:
We had a small mistake in our provider analysis - I updated the pdf to correct the value.

I already updated https://github.com/deltachat/eppdperf/blob/main/results/e-mail_provider_comparison.ods, where the more detailed view is available.


Simon-Laux (approved):

Simon-Laux (changes requested):
the preview picture needs to be updated

r10s (approved):

PR #524, opened 2022-03-30
fix bitcoin address

missytake:
@hpk42 please double-check - which one is the correct address? :see_no_evil:


hpk42 (approved):

hpk42:
thanks for noticing!

PR #531, opened 2022-05-06
get.delta.chat: bump android version to 1.30.1

r10s:
upload as https://download.delta.chat/android/deltachat-gplay-release-1.30.1.apk is already done.

maybe wait with merging until it is also approved at least in gplay ...


r10s:
> maybe wait with merging until it is also approved at least in gplay ...

it is approved, so we can move forward and merge this pr :)

missytake (approved):

PR #534, opened 2022-05-12
Add download button

rosano:


Simon-Laux (approved):

missytake:
Hm, can we somehow center this button, and maybe move it above the screenshots?

Screenshot_2022-05-24_13-09-04

Screenshot_2022-05-25_13-44-26

(preview: https://staging.delta.chat/537/en/)

I don't know what you discussed about this though, so if there is a design idea behind it, I don't want to stand in the way.

r10s:
> Hm, can we somehow center this button, and maybe move it above the screenshots?

i think, it is perfectly fine where it is now, nothing on the page is centered, so a centered button would look a bit mispositioned. esp. if the screenshots wrap on smaller screens.

i suggest to merge that in as is therefore, but you already said, that your point is no veto or so :) 

nb: thanks for the screenshot, i have not "seen" the change before - meanwhile @rosano can also do feature branches so that we get a nicer preview.

r10s (approved):

PR #535, opened 2022-05-20
run `bundle update` to fix error

r10s:
this fixes "Could not find public_suffix-4.0.6 in any of the sources"
error with ruby 3.1

updating some gems, however, seem to be useful anyway.


missytake:
I'm having an entirely different error (when running `jekyll serve` with ruby 2.7): `/usr/lib/ruby/2.7.0/rubygems/dependency.rb:311:in `to_specs': Could not find 'addressable' (~> 2.4) among 70 total gem(s) (Gem::MissingSpecError)`

It doesn't disappear with this branch unfortunately. not a blocker, just means I can't review it right now - but I didn't look into it for very long, I can try again in the next days. If anyone else can review it meanwhile, that would help, too.

(Isn't it anyway a nobrainer to merge this?)

r10s:
> Isn't it anyway a nobrainer to merge this?

yeah, it should be :) i created the pr mainly for documentation :)

merging this therefore directly as your issue seems to be unrelated (too old ruby?) and the server manages to build the page as well.

PR #536, opened 2022-05-24
fix link on Spanish and Turkish homepage

dotlambda:
reported by @M4ss1ck


missytake (approved):
Thanks @M4ss1ck for finding this!

missytake:
wait, this transifex stuff is confusing sometimes. Maybe we need to fix this in transifex, not on our site

missytake:
Yep, I tried it out; we can merge this, but it will be overwritten by the next translations pull.

I fixed it in spanish on transifex, but I can't edit the turkish string because of transifex-internal permissions. @r10s what's the normal move for this?

missytake:
Okay, I fixed it on transifex - now it will not be overwritten by a tx pull.

PR #540, opened 2022-05-30
update desktop version and add other downloads

Simon-Laux:
desktop version to 1.30.0
and add rpm and tar.gz


r10s (approved):
great, so many additional options!

layout-wise also still fine :)

PR #550, opened 2022-07-10
AEAP blog post

Hocuri:


_posts/2022-07-10-aeap.md
```diff
@@ -0,0 +1,19 @@
+---
+title: Changing your email address made easier 
+author: hocuri
+---
+
+In Delta Chat 1.32, we added experimental support for changing your email address and telling your contacts about it, which we call AEAP (Automatic Email Address Porting). This means that you can now change your own email address in the settings (under "Password and account" - TODO, on Android we may want to make the button stand out more since I myself sometimes have issues seeing it; still then, it's unfortunately rather hard to describe where it is), and things will just work! Kind of, this is still a very new feature so there may be bugs, and also, your contacts will only know that you changed your address once once they got a message from you. Delta Chat doesn't do any magic here, it just makes changing your address a bit smoother.
```
Hocuri: TODO: _This means that you can now change your own email address in the settings (under "Password and account" - TODO, on Android we may want to make the button stand out more since I myself sometimes have issues seeing it; still then, it's unfortunately rather hard to describe where it is),_

How do we describe where people should go? Just _under/at "Password and account"_? Or _"Profile settings -> Password and account"_? The problem is that, under Android, "Password and Account" is in the profile settings, which have the avatar and displayname shown, but there is no string that we could tell the user to tap on.

Hocuri:
Hm, the footnote at the end looks a lot like it's part of the main text, and seems like we need an image for the post

_posts/2022-07-10-aeap.md
r10s: i added a commit, fixing some thing. it says now "Settings / Your Profile / Password and Account" - we are using this term also elsewhere, i think, this is okay.

r10s:
@Hocuri i added the screenshots and did some formatting, see my commit message for details.

i am done with the pr so far, feel free to overwrite my changes as you like ;)

missytake:
Maybe we should add something about the trust model? That this works *because* the private key stays the same? And that it also works if you lost access to your old account, but still have the private key?

...ah right, don't tell users about keys :upside_down_face:

Hocuri:
> ...ah right, don't tell users about keys

"and your encryption key stays the same and is not altered"

I just noticed that we already do, but I think it's fine at this point, we don't require the user to know what a PGP key is, or asymmetric encryption.

Advanced stuff like the trust model could go into the footnote, though I don't have the time to write this myself right now (need to fix all the other AEAP issues).

r10s:
> Hm, the footnote at the end looks a lot like it's part of the main text

i formatted the footnotes to be a bit nicer :)

_posts/2022-09-14-aeap.md
```diff
@@ -0,0 +1,43 @@
...
+
+The Delta Chat 1.32 releases introduced support for new providers (most notably 163.com), a bigger internal "move to Tokio" infrastructure change resulting in better speed, and the first release of AEAP mechanisms. AEAP enables for e-mail what number porting does for mobile phone communications, i.e. to more easily change between communication providers.
+
+TLDR; How to migrate to a new provider with Delta Chat:
+
+- Probe and change your address at "Password and Account" seetings screen in Delta Chat
+- Make your old email provider forward all emails to your new email address
+- Continue chatting as usual :)
+
+By design, no user confirmation dialogues aka "is this ok?" / "are you sure?" dialogues are presented because user studies show that most people just click "ok" or "cancel" as needed to move forward and rarely understand the implications. If you later discover you want to stick around with your old e-mail address just move back to it :)
```
r10s: well, there is at least _some_ confirmation dialog, it is even shown for android in the blog post just below the statement that no confirmation dialog is shown. ios/desktop have a similar dialog.

(cmp preview at https://staging.delta.chat/550/en/2022-09-14-aeap)

maybe strike the paragraph or point out the fact it is reversible differently.

_posts/2022-09-14-aeap.md
```diff
@@ -0,0 +1,43 @@
+---
+title: Introducing Automatic E-mail Address Porting (AEAP)
+author: hocuri, holga
+image: ../assets/blog/2022-07-aeap-0.png
+---
+
+The Delta Chat 1.32 releases introduced support for new providers (most notably 163.com), a bigger internal "move to Tokio" infrastructure change resulting in better speed, and the first release of AEAP mechanisms. AEAP enables for e-mail what number porting does for mobile phone communications, i.e. to more easily change between communication providers.
+
+TLDR; How to migrate to a new provider with Delta Chat:
```
r10s:
```suggestion
TL;DR - How to migrate to a new provider with Delta Chat:
```

r10s (approved):
in general, this lgtm.

once content is settled, e-mail vs. email typo should be synced and maybe some layout nitpicks pop up. but no blockers :)

_posts/2022-09-14-aeap.md
Hocuri: I think that he means that no confirmation dialog is shown on the receiving side, but yes I also stumbled over that part.

r10s: ahhhh :) that makes sense yes. but then the "if you later" speaks again about sending side.

maybe split this paragraph into two to clarify:

```
By design, to the receiver, no user confirmation dialogues
aka "is this ok?" / "are you sure?" dialogues are presented
because user studies show that most people just click "ok" or "cancel" 
as needed to move forward and rarely understand the implications. 

[images]

If you later discover you want to stick around with your old e-mail address
just move back to it :)
```

hpk42:
i stroke the offending para, it's not really needed.

r10s (approved):

PR #560, opened 2022-09-15
Add PGP key requirements to the FAQ page

kotochor:
It took me a few hours to figure out why I can't import my existing RSA4096 key on both my devices. The app only sad that it didn't find any private key in the folder. If the key check won't be implemented, then at least the FAQ page should have an explanation of the bug.

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -340,18 +340,22 @@ passphrase encryption and the password and try the import again.  If you want
 to keep your passphrase you'll have to create an e-mail alias for use
 with Delta Chat such that Delta Chat's key is tied to this e-mail alias.
 
-Delta Chat supports common OpenPGP private key formats, however, it
-is unlikely that private keys from all sources will be fully supported. This
-is not the main goal of Delta Chat. In fact, the majority of new users 
-will not have any key prior to using Delta Chat.
-We do, however, try to support private keys from as many sources as possible. 
-
 Removing the password from the private key will depend on the
 software you use to manage your PGP keys. With Enigmail, you can set your
 password to an empty value in the Key Management window. With GnuPG you can set
 it [via the command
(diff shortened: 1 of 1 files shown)


r10s:
thanks for figuring that out!

that is maybe just a bug that needs to be fixed in core, there is also an open issue at https://github.com/deltachat/deltachat-desktop/issues/2030 (and an older one closed at https://github.com/deltachat/deltachat-desktop/issues/1226 ) 

we can still add a hint to the faq, however, as that will be temporary and to not add additional burden to the translators, we should keep the diff small (so do not move paragraphs around).

@missytake can you confirm that 4096 bit rsa secret keys are not working?

missytake:
Sorry that it took so long to check up on this.

I just successfully imported an RSA4096 key in Delta Chat 1.32.1 on xubuntu 20.04. It is annoying that you can only choose a folder and not a file; but well, after I tried it with an otherwise empty folder it worked.

Maybe it works for you now, as well?

I suggest to close this PR, because it doesn't reflect the current state of the implementation.

kotochor:
I have the same version of Delta Chat 1.32.1 installed from Flatpak on Pop!_OS 22.04 LTS.
I've checked right now. It still doesn't work, comparing to ed25519 key.
Got this error message:
`IMEX failed to complete: No private keys found in "<...>".`

Thank you for checking. It's going to be my own mysterious error

gerryfrancis:
@uladz-p FYI, I had the same issue like you with a RSA 4096-bit key. I have just retested with the latest versions, and I have been able to import the key successfully. Maybe you want to give the latest versions of Delta Chat another try. :) The only condition is that the key must not contain a passphrase, otherwise an import will fail.

kotochor:
@gerryfrancis you are right. I have successfully imported RSA 4096-bit private key on desktop and mobile. No error message.
The same installation, but updated to the latest version

Issue #542, opened 2022-06-03
delta.chat/en/ has a big download button and suggests get.delta.chat which are 404 errors


Mikaela:
- Operating System (Linux/Mac/Windows/iOS/Android): Fedora 36 and iOS 15.5
- Delta Chat Version: N/A
- Expected behavior: Pressing the download links on https://delta.chat/en gives me the download options sending me to https://delta.chat/en/download (like the Download link on top of the page)
- Actual behavior: I am given a 404 error.
- Steps to reproduce the problem: Navigate to https://delta.chat/en on either Firefox 102.0b3 or Safari on iOS 15.5, scroll to bottom of page and press download or get.delta.chat
- Screenshots:

* ![kuva](https://user-images.githubusercontent.com/831184/171931200-daa70569-9abe-459d-bb6b-df11a7a77df3.png)
   * The problematic links
 * ![Supposed download page](https://user-images.githubusercontent.com/831184/171931263-c3d0a451-4c1a-4ea9-b787-efd8a93ecf8f.png)
   * The link target. Additionally clicking the link results to #541.

- Logs: N/A or I don't have access to the server to see them.

Mikaela:
I am unable to reproduce these two issues using a clean Firefox profile through `firefox --ProfileManager`, but I have no idea how to clean up my iOS to not result to this issue. I also doubt being the only one encountering these.

r10s:
weird. i cannot reproduce that either.

cc @missytake - maybe you have an idea? maybe some strange redirect conditions? maybe related to the weird double-adding of `/en/` on 404's? (eg. if you try to open <https://delta.chat/en/does-not-exist>

missytake:
Hm. @Mikaela which languages are configured in your browsers?

Mikaela:
In Firefox apparently only Finnish which I think is the default upon downloading from https://releases.mozilla.org/pub/firefox/releases/102.0b3/linux-x86_64/fi/

In iOS I assume it uses the system languages which are in this order: Finnish, English (Ireland), Esperanto, Swedish, Espaňol and Russian.

Mikaela:
Yesterday I noticed this was reported in the forum already with more details

* https://support.delta.chat/t/website-not-working-with-certain-accept-language-header/2043?u=aminda

r10s:
i think, this was only a side-effect of fixed https://github.com/deltachat/deltachat-pages/issues/618 - if not, please open a fresh issue with detailed information

PR #561, opened 2022-09-25
add toc and static anchor for donate to contribute page

Simon-Laux:
Table of Contents is that people can directly see a link to the donate section.
Anchor is that we can link to the donate section from the apps: https://delta.chat/contribute#donate-anchor (notice the missing language in path, this will be replaced by automatically the user's language and the static (not translated anchor) ensures they jump to the right place on all languages).

we need to regenerate/upload the thing to/from transifex after merging. ping @r10s 

this is a first step, I think we need a dedicated site for money donation sometime anyways.
Also the text could be updated, liberapay is only recurring payments for example (that info is missing).

### DIFF
--- a/en/contribute.md
+++ b/en/contribute.md
@@ -2,9 +2,10 @@
 title: Contribute
 lang: en
 render_toc: true
+header: Contribute or Donate
 ---
 
-# Channels and repositories
+## Channels and repositories
 
 - [Fediverse](https://chaos.social/web/@delta) and
   [Twitter](https://twitter.com/delta_chat) for announcements.
@@ -21,7 +22,7 @@ render_toc: true
 - [Delta Chat repositories](https://github.com/deltachat/) where you can 
   find the code for DeltaChat apps and this web site.
 
-# Translations and Bug Reports 
+## Translations and Bug Reports 
 
 - [Translations on Transifex](https://www.transifex.com/delta-chat/public/)
 - [Delta Chat Android Issues](https://github.com/deltachat/deltachat-android/issues)
@@ -37,7 +38,7 @@ Please don't hesitate to reach out (delta at merlinux eu) if you are interested 
 
 <a id="donate-anchor"></a><!-- don't translate this html anchor!-->
 
-# Donate Money 
+## Donate Money 
 
 If you monetarily support DeltaChat you help it to stay and grow as an independent project that works and is oriented towards its users: 
 
--- a/en/contribute.md
+++ b/en/contribute.md
@@ -1,6 +1,7 @@
 ---
 title: Contribute
 lang: en
+render_toc: true
 ---
 
 # Channels and repositories
@@ -33,6 +34,8 @@ C, Rust, Java, Swift, Javascript or Python on Android, iOS, Windows, Linux or Ma
 We typically offer 20 hours per week contracts or employments (if based in germany). 
 Please don't hesitate to reach out (delta at merlinux eu) if you are interested to help our efforts!
 
+
+<a id="donate-anchor"></a><!-- don't translate this html anchor!-->
 
 # Donate Money 
 


r10s:
> this is a first step, I think we need a dedicated site for money donation sometime anyways.

then better use sth. as https://delta.chat/donate and [add a redirect](https://github.com/deltachat/sysadmin/blob/master/delta.chat/delta.chat)  - having a very specific link in the apps will make it hard to change the page.

en/contribute.md
```diff
@@ -34,7 +36,9 @@ We typically offer 20 hours per week contracts or employments (if based in germa
 Please don't hesitate to reach out (delta at merlinux eu) if you are interested to help our efforts!
 
 
-# Donate Money 
+<a id="donate-anchor"></a><!-- don't translate this html anchor!-->
```
obsoleszenz, line 39: Let's just name it donate? The url looks nicer and we are not doing any custom css anyways that would select on ids

missytake:
I added the redirect :) it is executed *before* the language redirects, so it should work. We can link to https://delta.chat/donate.

(OTOH, unfortunately there is still #542, so maaaybe we shouldn't rely on the language redirects to much at the moment.)

missytake:
I'm not super happy with the look of the TOC right now. It would be nice if it would stand out more as a separate thing... like if there was a small box around it. But we can increment on this later.

missytake:
> we need to regenerate/upload the thing to/from transifex after merging. ping @r10s

The transifex push happens [automatically](https://github.com/deltachat/deltachat-pages/blob/master/.github/workflows/tx-push.yml), but as soon as it is translated, we need to pull in the translations with everything else of course.

en/contribute.md
```diff
@@ -1,9 +1,11 @@
 ---
 title: Contribute
 lang: en
+render_toc: true
+header: Contribute or Donate
```
missytake, line 5: Not sure about changing this, too. :shrug:

en/contribute.md
missytake, line 39:
```suggestion
<a id="donate"></a><!-- don't translate this html anchor!-->
```

yes, it's an anchor, do we need to have it in the anchor? Would look slightly prettier this way, but not important (it's not very user-facing anyway.

Is `<!-- don't translate this html anchor!-->` visible for the translators? Or do they only see `donate` this way? Either way, this way we might spot it when we pull the translations from transifex, and fix it before merging them in.

missytake (approved):
All in all I'm not super sure if this PR is an improvement; the TOC doesn't look very good as it is right now, and the anchor link is not thaaat useful as on desktop the view can't scroll all the way down to the donate headline anyway (the page is too short). This could be improved by adding an image to the bottom; e.g. a group picture of delta chat contributors, or a "shut up and take my money" meme (but I'm not sure about that either xD)

So I'm just not very sure in either direction, and approving that mainly because it won't break anything, and I don't want to stand in the way of progress :D

en/contribute.md
Simon-Laux: as long as its not translated and it works (delta.chat/donate is updated too) I'm fine with it.

PS: we can't use the autogenerated anchors, because they are translated.

Simon-Laux:
so we merge this now?

r10s:
i just rechecked; i agree with @missytake that the toc does not look good, the page is small anyway and it is quite verbose now, quite some repetition.

i get the idea of the pr, easier forwarding to donate, however, the page is more complex and verbose now.

not sure if this pr is an improvement overall. maybe better have a separate donate page and link? or reorder the page? important things atop? solution may also be to do nothing for now, i always found the site quite easy to navigate, ppl scroll fast through pages anyway, cc @rosano 

also, maybe wait with that until we relayout the page? having a toc used for one small page does not make things easier wrt design.

if the anchor is needed, maybe split this off to another pr, that can be merged  quickly (but i would also not block merging this pr, if that is considered an overall improvement by others :)

rosano:
I would move the donate stuff to a seperate page at `/donate` as the anchor linking can be unpredictable depending on how it's accessed. It should be as easy as possible for people to give money and putting it at the bottom of a contribute page might be hard to find. Linking from the contribute page though is definitely a good idea.

The table of contents is visually weak because it's styled exactly the same as the other lists, but also not so necessary as the headings themselves are quickly scrollable. If the goal is to make it easier to donate, I would put a donate 'button/link' (with similar style to the 'Download' button on the homepage) at the top, before the channels section. Could even be paired with some kind of heading, but just a link would be a simple start. It needs some visual differentiation order to stand out.

So I'm not sure if it would be better to change this PR or start a new one without the TOC. If it's helpful I can offer to make a new one and do the work of moving the donate content to a seperate page + also putting a button on this page.

Simon-Laux:
+1 for making a new pr @rosano

missytake:
stale, closed for now.

PR #564, opened 2022-10-10
Updating the FAQ

missytake:
closes #510
closes #565 

- [x] is this still up-to-date? https://delta.chat/en/help#which-messages-do-appear-in-delta-chat
- [x] the security review mentioned in https://delta.chat/en/help#what-is-a-verified-group-why-is-it-experimental has happened
- [x] It would be wise to link to https://support.delta.chat/t/import-backup-to-ios/1628 in https://delta.chat/en/help#can-i-use-delta-chat-on-multiple-devices-at-the-same-time
- [x] how private are private apps? (webxdc trust model)
- [x] where can I get private apps?
- [x] how can I create my own private apps?
- [x] what is the "recently seen" dot and what doe it mean? (eg. it does no tracking, in contrast to whatsapp etc.)
- [x] what happens if I turn on auto-delete messages from server?
- [x] what happens if I turn on auto-delete messages from client?
- [x] "experimental" section
  - [x] Can I share my location with my chat partners?
  - [x] Does Delta Chat support Audio/Video Calls?
  - [x] How does the experimental database encryption work? 
  - [x] how can I change my account? (AEAP)
  - [x]  What are broadcast lists and how do they work?

PR #567, opened 2022-10-29
update translations

r10s:


missytake (approved):
I checked the links, looks fine.

The anchor äinks (#encryption) are brekaing - is this intended? It *llooks* like nothing would fail.

r10s:
> The anchor äinks (#encryption) are brekaing - is this intended? It llooks like nothing would fail.

the anchor was removed at https://github.com/deltachat/deltachat-pages/pull/564/files#diff-d4c3e5414f550d84ffce94a04386b7a390beaad75539782e303085bcd914d9b3L201 - and therefore also in all translations.

not sure, if it is used somewhere, however, if there is some dangling link somewhere, one would still land atop of the page, so maybe good enough.

missytake:
On the staging page it works, I think `_includes/toc.html` creates the anchor links, not markdown. So it doesn't affect anything. It's just a bit weird that the toc.html code is already there since 2019 and those anchor links only get removed now. But whatever.

Anything in the way of merging?

r10s:
> Anything in the way of merging?

no :)

PR #576, opened 2022-11-30
fixing link

### LINKED ISSUE #575: FAQ (in German): Broken link, pull from Transifex needed
Another pull is needed from Transifex, because the link to "Private Apps" in section "Was bedeutet der grüne Punkt?" (in German) was missing a single dash, therefore the link does not work currently. Please apply, thanks! :)

missytake:
closes #575, thanks @gerryfrancis for the notice


missytake:
works now.

PR #579, opened 2022-12-08
migrate to new tx tool

### LINKED ISSUE #578: migrate to new `tx` tool
the currently use `tx` tool written in Python is sunset end Nov22 [^1].

there is a new tool under the same name, written in Go now.

installation and migration is not a big deal, https://developers.transifex.com/docs/cli i did this locally recently for [android](https://github.com/deltachat/deltachat-android/pull/2434) and [ios](https://github.com/deltachat/deltachat-ios/pull/1764).

i will prepare a pr for deltachat-pages as well, however, this also requires to use the new tool in ci, not sure how this part works out, cc @missytake 

[^1]: https://github.com/transifex/transifex-client

r10s:
- [x] adapt code to new `tx` tool, see https://github.com/deltachat/deltachat-pages/issues/578
- [x] ~~adapt ci to use the new `tx` tool~~ - EDIT: seems not to be needed, ci building pages does not interact with transifex

closes https://github.com/deltachat/deltachat-pages/issues/578

### DIFF
--- a/.github/workflows/tx-push.yml
+++ b/.github/workflows/tx-push.yml
@@ -3,7 +3,9 @@ name: Push new strings to Transifex
 on: 
   push:
     branches:
-      - master
+      # if we made sure, the new tx-tool works and is used by CI, we can switch back to branch "master"
+      # (is there a less hacky way to disable a workflow?)
+      - master-disabled
 
 jobs:
   push-translations:
--- a/tools/.tx/config
+++ b/tools/.tx/config
@@ -1,58 +1,69 @@
 [main]
 host = https://www.transifex.com
 
-[delta-chat-pages.blogpo]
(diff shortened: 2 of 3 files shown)


r10s:
hm, i was expecting some ci error, but maybe the old tool (as used on ci) can deal with the new section names of the new tool used locally (this is the only change).

i will push a change to `en/index.md` to check if the page is acutually okay.  EDIT: that worked out, things work as expected, unless i've overseen somethings :) EDIT: reason is that the CI building the pages does not interact with transifex at all. maybe we should disable pushing to transifex from CI for now, so that this does not mess up things (and it is questionable anyway)

r10s:
i re-run `tx migrate`, is suggest to merge this in so that things do not  get stale again.

PR #581, opened 2022-12-11
update translations

r10s:


missytake:
A shame, it seems this PR rolls back a lot of already existing translations back to english...

missytake:
Did you already use the new tool from https://github.com/deltachat/deltachat-pages/pull/579 for this PR?

r10s:
> A shame, it seems this PR rolls back a lot of already existing translations back to english...

probably because of minor changes in the english sources.

> Did you already use the new tool from https://github.com/deltachat/deltachat-pages/pull/579 for this PR?

no, however, this is mostly independent. i'd merge this pr before #579, however, just in case #579 makes problems :)

r10s:
(not sure why this pr was a draft, it was never meant to be a draft :)

rosano:
I think the rollbacks were triggered by this change which changed some whitespace https://github.com/deltachat/deltachat-pages/commit/4195601679a15ffd12f62a645282d703eecef4bb

rosano:
I was able to find one string where I can use the old translation even though the source changed (the option is under the `Suggestions` tab) https://www.transifex.com/delta-chat/delta-chat-pages/translate/#ca/contributepo/445295773

I'm not sure whether doing this 34 times per string because of whitespace changes will be sustainable. Wondering if there's a known way to warn transifex in advance and match them up correctly.

r10s:
> I think the rollbacks were triggered by this change which changed some whitespace https://github.com/deltachat/deltachat-pages/commit/4195601679a15ffd12f62a645282d703eecef4bb

good catch! i hotfixed the accidental whitespace changes on master https://github.com/deltachat/deltachat-pages/commit/b8b555bab2307b7f08361a2a725c5680b9fb7892 - we'll see if transifex is smart enough to find that 100% matching translation automatically ...

not sure, how we can issue a warning or fix that smartly for the future. in general, i'd say editors should leave lineends alone, also because of the [special meaning of "double space" at the end of a line](https://daringfireball.net/projects/markdown/syntax#p).

r10s:
> we'll see if transifex is smart enough to find that 100% matching translation automatically ...

k, that worked out, existing translations saved ✅ 

i rebased, and the locale pages now also have the homepage updated wrt https://github.com/deltachat/deltachat-pages/pull/586

let's merge that in before things get stale again :)

PR #582, opened 2022-12-12
documented docker build command

missytake:
The normal `bundle install` and `jekyll serve` stuff fails often for many people, because who bothers to keep their ruby stuff up-to-date? This is not optimal, but might be helpful to some.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,12 @@ Testing changes
   and execute `jekyll serve --destination <path-to-build-dir>` 
   in the source directory.
 
+- If you get Ruby dependency problems, you can also build the page in a docker
+  container:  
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -39,6 +39,12 @@ Testing changes
   and execute `jekyll serve --destination <path-to-build-dir>` 
   in the source directory.
 
+- If you get Ruby dependency problems, you can also build the page in a docker
+  container:  
+  ```sudo docker run -v $PWD:/srv/jekyll -v $PWD/_site:/srv/jekyll/_site jekyll/builder:3 /bin/bash -c "chmod 777 /srv/jekyll && jekyll build --future"```  
+  But the links will not work, so you need to open the pages one after the other.
+  You can find them in the `/site` subdirectory.
```
hpk42, line 46:
```suggestion
  You can find them in the `_site` subdirectory.
```

hpk42 (commented):
it worked although i failed to run it in a loop. still worth to merge already i think.

README.md
missytake: oops, overlooked the suggestion. I pushed it directly to master.

PR #583, opened 2022-12-12
document how encryption can be guaranteed

### LINKED ISSUE #566: FAQ: How to use delta in a way that protects metadata and never sends anything unencrypted ?
This came up [on Mastodon](https://hostux.social/@LibreFaso/109228069261751395), and is a common need from users who want maximum security because they don't know their threat model exactly.

My answer on Mastodon:

"The best way to do that is creating a verified group and turning on self-destructing messages :) verified groups are always encrypted, self-destructing messages delete the messages on the server, which at least deletes the metadata in the aftermath. If you need the messages on your device, but not on the server, you can also agree in the group to turn on "delete messages from server automatically"."

Now this should become an FAQ item.

missytake:
closes #566

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -307,9 +307,11 @@ If you are not near the chat partner, you can check the status manually in the "
 - If there is **no padlock**, the message is usually transported unencrypted e.g. because you or the sender have turned off end-to-end-encryption, or the sender uses an app without support for end-to-end-encryption.
 
 
-### How can I ensure every message is encrypted, and metadata deleted as quickly as possible?
+### How can I ensure message encryption and deletion?
 
-The best way to do that is creating a verified group and turning on
+The best way to ensure every message is encrypted,
+and metadata deleted as quickly as possible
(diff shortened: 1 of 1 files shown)


en/help.md
```diff
@@ -306,6 +306,28 @@ If you are not near the chat partner, you can check the status manually in the "
 
 - If there is **no padlock**, the message is usually transported unencrypted e.g. because you or the sender have turned off end-to-end-encryption, or the sender uses an app without support for end-to-end-encryption.
 
+
+### How can I ensure every message is encrypted, and metadata deleted as quickly as possible?
+
+The best way to do that is creating a verified group and turning on
```
hpk42: don't start a first para sentence with a "that" reference.  Maybe the header better reads "Message encryption and deletion"  and then you can re-use the current header in the first sentence or so?

hpk42 (approved):
apart from the comment i think this is ready to merge.

PR #577, opened 2022-12-05
add a community standards doc, based on the autocrypt code of conduct 

hpk42:
This is a ported version from https://github.com/autocrypt/autocrypt/blob/master/CODE_OF_CONDUCT.rst

Missing:
- [x] e-mail addresses for nami and xenia as "last resort" contact persons.  
- [x] make it part of the "to be translated" docs

PR #584, opened 2022-12-14
Move donate to separate page

rosano:
Continuing @Simon-Laux's effort from #561, this moves the donate section to a separate page and adds a button from the contribute page. I will probably need to move some things over from transifex and maybe @r10s can help me understand how it works?


r10s:
> maybe @r10s can help me understand how it works?

there is not much to do, the english page is already there at https://staging.delta.chat/584/en/donate (in the previous or  https://github.com/deltachat/deltachat-pages/pull/577#issuecomment-1345620927 the error was that the frontmatter was missing and the page was just not generated at all, not even in english.

so, for the english page, there is nothing to do.

for the translations, once this pr is merged, the strings need to be pushed to transifex via `./tools/t-dance push` - currently, this is done automatically on each commit to master, however, this is a bit questionable, but still (wording is often only settled a few days after pushing to master, there are typos etc. - and that add unnecessary load to translators).

finally pulling the translations is done manually with `./tools/t-dance pull` - manually, to get aware of broken links and things like that.

assets/css/styles.scss
```diff
@@ -2,12 +2,14 @@
...
-.download-button {
-	border: 1px solid black;
+.cta-button {
+	border: 1.5px solid black;
 	border-radius: 5px;
 	padding: 10px;
 	color: black;
 	font-weight: bold;
+	display: inline-block;
+	margin: 10px 0;
```
r10s, line 10: in several side-discussions i got the hint that the button style is not really nice and does not fit nicely to the rest of the page. i share this critique.

i we now use the button style at more places, we should think over adapting it.

i would suggest to have a button with a background color matching the menu background and a white text. the size is probably fine.

en/contribute.md
```diff
@@ -3,6 +3,8 @@ title: Contribute
 lang: en
 ---
 
+[Donate Money](donate){: .cta-button}
```
r10s: this might be a bit too much in the face at that position, i would move that to the end of "channels and repositories" - or maybe right-floating to that paragraph if there is space.

en/donate.md
```diff
@@ -0,0 +1,12 @@
+---
+title: Donate Money
+lang: en
+---
+
```
r10s, line 5:
```suggestion

# Donate Money
```

we have a headline on virtually all pages, to match the layout, we should add one here as well.

en/index.md
```diff
@@ -26,7 +26,7 @@ style="float: left; margin: 10px;display: block;box-shadow: 5px 5px 2px #777;" a
 
 <img src="../assets/blog/screenshots/2020-01-09-delta-chat-iOS-weekend-group-chat.png" width="110" style="margin: 10px" alt="A screenshot of Delta Chat on IOS" /> 
 
-<a class="download-button" href="https://get.delta.chat">Download</a>
+[Download](https://get.delta.chat){: .cta-button}
```
r10s, line 26: nice - i was not aware markdown/jekyll has this shortcut :)

r10s (commented):
great you push that forward!

Simon-Laux:
maybe also add a link to the donate-money page: "Donate your Time/expertise instead"/"Other ways to contribute"

en/contribute.md
rosano: done

rosano:
@Simon-Laux done

assets/css/styles.scss
rosano: done

assets/css/styles.scss
```diff
@@ -2,11 +2,15 @@
...
 	border-radius: 5px;
 	padding: 10px;
-	color: black;
+	display: inline-block;
+	margin: 10px 0;
+
+	background: linear-gradient(120deg, #77888f, #364e59);
+	color: white;
 	font-weight: bold;
 }
```
r10s, line 15: i think, to have the button white , sth. as 
```
cta-button a {
    color: white;
}
```
is needed. currently, the anchor has the "baby blue" used elsewhere which does not look nice on the blueish background.

rosano: The class is the `a` element also, but actually I think there is a difference between how stuff renders for me vs on the site somehow, the current button is black text where as on the site it's blue. In any case, I added an `!important` which should hopefully render as intended:

Screen Shot 2022-12-15 at 18 00 13

r10s: yeah! that button looks good! 

but !important or not, safari shows it to me as:

Screenshot 2022-12-15 at 18 25 55

i think, the issue is we've defined `a:hover` explicitly - and browsers have different precedences if tagname and class definitions conflict. so we probably just need to say `a.cta-button, a.cta-button:hover` in the selector.

en/donate.md
```diff
@@ -0,0 +1,36 @@
...
+Help DeltaChat develop and grow as an independent project serving the people who use it:
+
+## Euros
+
+*Please always specify the account holder 'Björn Petersen'*
+
+__IBAN__: DE86100777770428658900
+
+__SWIFT__: BIC NORSDE51XXX
```
r10s:
```suggestion
__SWIFT/BIC__: NORSDE51XXX
```

i think, this is more correct, "SWIFT" and "BIC" are used mostly synonym in this context; at least the "SWIFT: BIC ..." seems wrong as it implies somehow that the SWIFT-code starts with BIC.

en/donate.md
```diff
@@ -0,0 +1,36 @@
+---
+title: Donate Money
+lang: en
+---
+
+# Donate Money
+
+Help DeltaChat develop and grow as an independent project serving the people who use it:
```
r10s:
```suggestion
Help Delta Chat develop and grow as an independent project serving the people who use it:
```

we are [not strict](https://github.com/deltachat/interface/issues/16) with the spelling of "Delta Chat", however, within a context it should be consistent (and on delta.chat we use "Delta Chat", modulo errors :)

assets/css/styles.scss
rosano: I think this might be a `:visited` issue, so I added a definition for that, can you see if that fixes it?

rosano:
@r10s any other suggestions or can we move forward on this?

assets/css/styles.scss
r10s: it is still not working for me on safari.

r10s:
> @r10s any other suggestions or can we move forward on this?

the css for the link is still not working for me on safari. that needs to be fixed.

maybe we should also surround the "Donate" button by a little bit of text or by a headline or so - or move it to the right. the current layout looks a but strange to me, the button does not really fit the section, imho. maybe okay and good enough :)

however, as this affects some very basic things, i'd like to hear some more voices, @Simon-Laux , @link2xt , @adbenitez  :)

Screenshot 2022-12-22 at 19 07 50

rosano:
> the css for the link is still not working for me on safari. that needs to be fixed.

Which version is this? Mine is an old 12.1.2 and the text is white. Could you also try inspecting the element with devtools and seeing which rule is triggering that blue text?

> maybe we should also surround the "Donate" button by a little bit of text or by a headline or so - or move it to the right. the current layout looks a but strange to me, the button does not really fit the section, imho. maybe okay and good enough :)

I was avoiding adding anything that needs new translations. Should we be concerned about that? Or is it fine for one line to be English at the moment? Could also look through all existing translations to find something that works as a header and hope it slots in without doing weird stuff in Transifex.

r10s:
> Which version is this? Mine is an old 12.1.2 and the text is white. Could you also try inspecting the element with devtools and seeing which rule is triggering that blue text?

mine is 16.1. i can have a look at that, i had sth like that 20y ago already - browser inconsistencies wrt link colors - seems not to be solved since then :)

rosano:
Is the deploy preview not updating since 15fbc47? https://staging.delta.chat/584/en/contribute is not actually showing the last change with `:visited` defined.

rosano:
> the css for the link is still not working for me on safari. that needs to be fixed.

Was this a local build or on staging? Seems like deployment was blocked for some reason (maybe merge conflict) and should reflect the current changes now.

it was staging, i will double check local.

link2xt:
Donate page looks too spacey for me, mostly because of a lot of headings. Can they be made smaller or turned into list items?

link2xt:
Maybe also add a "Donate" link to the footer, between "References | Source Code | Imprint | Privacy Policy"? Otherwise this page cannot be reached from the top-level navigation, it's always two clicks away, while other pages are directly reachable.

assets/css/styles.scss
r10s: you're right below, the issue was the failing ci, the preview is outdated. building it locally, things are okay.

r10s:
> Was this a local build or on staging? Seems like deployment was blocked for some reason (maybe merge conflict) and should reflect the current changes now.
> 
> it was staging, i will double check local.

i double checked local building - this is fine. so the issue indeed seems to be the non-updating ci preview :)

rosano:
> Donate page looks too spacey for me, mostly because of a lot of headings. Can they be made smaller or turned into list items?

I would deal with this in a future pull request. The original intention here was just to move this to a seperate page so it can be linked easily to avoid relying on hashes.

rosano:
> Maybe also add a "Donate" link to the footer, between "References | Source Code | Imprint | Privacy Policy"? Otherwise this page cannot be reached from the top-level navigation, it's always two clicks away, while other pages are directly reachable.

Done. Could also be in the top navigation but maybe let's try something more subtle for now. @r10s I'm not sure if I should touch the YAML—if not I can revert this.

en/donate.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Open Collective
+
+For one-time or recurring donations: [opencollective.com/delta-chat/donate](https://opencollective.com/delta-chat/donate)
+
+## Bitcoin
+
+Wallet [bc1qt3xhfvwmdqvxkk089tllvvtzqs8ts06u3u6qka](bitcoin:bc1qt3xhfvwmdqvxkk089tllvvtzqs8ts06u3u6qka)
+
+* * *
```
r10s: although `<hr>` is defined as a "thematically semantic break" in html5 - it is still a not-so-nice-horizontal line in practise. i would just remove this, also as we are not using that elsewhere and use for spacing instead.

rosano: Done.

rosano:
Anything in the way of merging this?

r10s:
> > Donate page looks too spacey for me, mostly because of a lot of headings. Can they be made smaller or turned into list items?
> 
> I would deal with this in a future pull request. The original intention here was just to move this to a seperate page so it can be linked easily to avoid relying on hashes.

but then - why not just leave the central paragraph and wording as is and iterate from there? this would also preserve most translations.

on current master, it is already less spacey and a list of items, i would also prefer this layout:

Screenshot 2022-12-26 at 13 20 24

rosano:
> this would also preserve most translations.

Right I forgot about that, reverted 👍🏽. What else?

r10s (approved):
lgtm now, i still think, the "donate" button is a bit unexpected at that position, but i also do not have a good idea about an alternatve. maybe it is good enough and we can iterate from here.

given the larger impact of a separate "donate" page and comparable larger changes, maybe give a little more time for other voices before merging.

link2xt (approved):

rosano:
@Simon-Laux This was coming from your PR so maybe you can share if you have anything to add?

hpk42 (approved):
lgtm

PR #586, opened 2022-12-14
Move logo to navigation bar

### LINKED ISSUE #460: Put the logo on the Download page
We got the feedback that there should be a logo on the download page: https://social.tchncs.de/@mray/106185401661663580

I agree with this feedback.

We could put a headline "Get Delta Chat for your device!" + Logo on the right to the top of the page.

rosano:
This makes it visible and consistent on all pages while freeing up space and design possibilities on the home page.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -25,6 +25,7 @@
 
 <div id="menu">
 <ul>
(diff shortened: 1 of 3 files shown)


r10s:
ftr: before/after:

Screenshot 2022-12-14 at 21 46 51 Screenshot 2022-12-14 at 21 47 01

r10s (approved):
i like it :) but apart from that, it makes sense and is pretty common to have the logo repeated at some static place.

PR #587, opened 2022-12-14
Link to cosmos

rosano:
This adds a small banner to the bottom of the page similar to https://cosmos.delta.chat so that readers can easily be aware of and discover various sites related to Delta Chat.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -19,6 +19,8 @@
 
 <link rel="stylesheet" href="../assets/css/styles.css" type="text/css" />
 <link rel="alternate" href="https://delta.chat/feed.xml" type="application/atom+xml" title="{%if tx.rss_news_title != ""%}{{ tx.rss_news_title }}{%else%}{{ txEn.rss_news_title }}{%endif%}" />
+<script defer src="https://cosmos.delta.chat/banner.js"></script>
+<link rel="stylesheet" href="https://cosmos.delta.chat/banner.css" type="text/css" />
 </head>
 
 <body id="top">


r10s (approved):
a super-global menu? i like it :) that sth. like this needed underlines a bit the decentralised and growing structure :)

for other reviewers: there are corresponding pr on most other sites, eg. https://github.com/webxdc/website/pull/29 - and you need javascript enabled when opening  https://staging.delta.chat/587/en/

as this is a larger, global change, maybe wait for at least one other review before merging that in.

adbenitez (approved):

r10s:
k, let's give it a try +1 - as this one is merged in, it makes sense to merge in the other open cosmos pr as well.

r10s:
i merged webxdc and bots - are there more open pr?

btw, maybe the forum should get the banner as well?

rosano:
> i merged webxdc and bots - are there more open pr?

just https://github.com/webxdc/.github/pull/1

> btw, maybe the forum should get the banner as well?

I wasn't sure how to do it + it might also become a distraction as there's so many interfaces in discourse + someone might be using the Discourse iOS app,  I imagine would need to be done more carefully, should I make an issue?

r10s:
> I wasn't sure how to do it + it might also become a distraction as there's so many interfaces in discourse + someone might be using the Discourse iOS app, I imagine would need to be done more carefully, should I make an issue?

one can argue that the forum is part of delta.chat as it is linked  in the main menu - so, maybe it is not needed to have a banner there, esp. as there are drawbacks. i'd be fine with leaving things as is for now and iterate over at some point :)

PR #588, opened 2022-12-15
Use more compact font

rosano:
This is aesthetic but also a bit more compact as the letters are less
wide.

### DIFF
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -169,7 +169,7 @@ body {
 #content ul li { list-style: disc outside; margin: 0.5em 0.2em; padding-left: 0.5em; }
 ol li { list-style: decimal; margin: 0.5em 0.2em; padding-left: 0.5em; }
 
-h1,h2 { font-weight: normal; font-size: 24px; letter-spacing: 1px; line-height: 1.4em; color: #000; }
+h1,h2 { font-weight: normal; font-size: 26px; letter-spacing: 1px; line-height: 1.4em; color: #000; }
 h3 { font-weight: bold; font-size: 16px; letter-spacing: 1px; color: #159957; }
 h4 { font-weight: bold; letter-spacing: 1px; color: #159957; }
 
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -155,10 +155,14 @@ p.blogdate {
 /*colors*/
 
 body {
-    font-family: Verdana, Arial, Helvetica, sans-serif;
+    font-family: 'Lucida Grande', Arial, Verdana, Helvetica, sans-serif;
     font-size: 15px;
     line-height: 1.7em;
     color: #000;
+}
+
+#menu, h1, h2, h3, h4, h5, h6 {
+    font-family: Arial;
 }
 
 #content ul, ol { margin-left: 1.5em; }
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -156,8 +156,8 @@ p.blogdate {
 
 body {
     font-family: Arial, Verdana, Helvetica, sans-serif;
-    font-size: 15px;
-    line-height: 1.7em;
+    font-size: 16px;
+    line-height: 1.6em;
     color: #000;
 }
 
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -162,7 +162,7 @@ body {
 }
 
 #menu, h1, h2, h3, h4, h5, h6 {
-    font-family: Tahoma;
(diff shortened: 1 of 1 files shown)


Simon-Laux:
no visible changes on firefox on linux

rosano:
Before:

Screen Shot 2022-12-15 at 19 15 11

After:
Screen Shot 2022-12-15 at 19 15 09

The first point about messaging anyone fits on one line.

rosano:
> no visible changes on firefox on linux

The fonts might not be available? But nothing was removed, only an extra option added if available.

r10s:
there is still some letter spacing i added that time for h1...h3 - is this still on purpose?

apart from that, indeed, "Lucida Grande" seems not to be considered a web-safe font (random link: https://www.cssfontstack.com) "Lucida Grande" is mostly mac - not even windows.

rosano:
> there is still some letter spacing i added that time for h1...h3 - is this still on purpose?

Wasn't aware of it but could be left in or taken out. I think it's too compressed without the spacing, but with Verdana it's too wide.

> "Lucida Grande" seems not to be considered a web-safe font (random link: https://www.cssfontstack.com) "Lucida Grande" is mostly mac - not even windows.

Updated to use Arial everywhere and Tahoma for headings.

missytake:
On my xubuntu with Firefox this change seems to make it worse right now.

Status quo:

Screenshot_2022-12-16_10-10-35

With this PR:

Screenshot_2022-12-16_10-10-42

**edit:** okay, not necessarily worse, aesthetics are subjective I guess - is this what Tahoma is supposed to look like?

rosano:
> is this what Tahoma is supposed to look like?

Weird because I only see `sans-serif` fonts in the whole project. Can you let me know what's the computed style/font for you? I can only imagine this happening because of browser defaults or custom user CSS. Here's what I see:

Screen Shot 2022-12-16 at 12 36 23

missytake:
I think it's because I don't have Tahoma and there is no fallback font configured?

Screenshot_2022-12-16_12-58-36

rosano:
> I think it's because I don't have Tahoma and there is no fallback font configured?

I thought it would use the body font… @missytake Would you mind editing either in the browser or in source to see if `font-family: Tahoma, inherit;` works? Just because the issue is on your machine. Worst case, we can use `font-family: Tahoma, sans-serif;`

missytake:
`font-family: Tahoma, inherit;` doesn't work, but `font-family: Tahoma, sans-serif;` works :)

rosano:
Ok hopefully that should be resolved for @missytake. Any other feedback?

missytake (approved):
works now :)

r10s:
the body text is harder to read with this pr.

this is just because the font is overall smaller now - even if the pt size is not changed. is this on purpose?

to get the effective same font-size, it needs to be increased from 15px to 16px and the line-height can be decreased from 1.7em to 1.6em - side by side we would have the same font-size but a bit more compact.

overall, i am +/-0 with this pr,  maybe it is good to have sth. fresh after these years.

rosano:
> needs to be increased from 15px to 16px and the line-height can be decreased from 1.7em to 1.6em - side by side we would have the same font-size but a bit more compact.

Done. I increased the heading size to give more contrast to the body text.

> overall, i am +/-0 with this pr, maybe it is good to have sth. fresh after these years.

I have (perhaps incorrectly) perceived various desires to improve the site design and I'm trying to do small incremental changes rather than big ones, so it might be a more noticeable improvement when they're combined. Perhaps I should also just propose non-incremental changes so that the difference is more obvious.

r10s:
> and I'm trying to do small incremental changes rather than big ones

this is very welcome!

and i think, after some forth and back, font is nice now. different, but nice :)

> so it might be a more noticeable improvement when they're combined. 

good point.

r10s (approved):

_sass/_base.scss
```diff
@@ -155,17 +155,21 @@ p.blogdate {
 /*colors*/
 
 body {
-    font-family: Verdana, Arial, Helvetica, sans-serif;
-    font-size: 15px;
-    line-height: 1.7em;
+    font-family: Arial, Verdana, Helvetica, sans-serif;
+    font-size: 16px;
+    line-height: 1.6em;
```
r10s:
```suggestion
    line-height: 1.5em;
```

did i say 1.6em? triying it over, i mean 1.5em :) with that, and the increased font-size, height and spacing are same as before (and aesthetic :), with, of course, the more compact font.

r10s (approved):

PR #589, opened 2022-12-15
Enlarge emojis on homepage

rosano:
If they are used as 'images' it could be more visible in larger size

### DIFF
--- a/assets/css/styles.scss
+++ b/assets/css/styles.scss
@@ -22,4 +22,5 @@
 .home-features + p, .home-features + p + p, , .home-features + p + p + p {
 	display: flex;
 	align-items: center;
+	justify-content: flex-start;
 }
--- a/assets/css/styles.scss
+++ b/assets/css/styles.scss
@@ -19,6 +19,9 @@
 	padding: 10px;
 }
 
+/* to avoid converting the markdown to HTML,
+sibling selectors are used here to select
+the first, second, and third p after .home-features */
 .home-features + p, .home-features + p + p, , .home-features + p + p + p {
 	display: flex;
 	align-items: center;
--- a/assets/css/styles.scss
+++ b/assets/css/styles.scss
@@ -16,31 +16,37 @@
 
 :root {
 	--emoji-size: 34px;
+	--emoji-spacing: 8px;
 }
 
-.home-emoji {
-	font-size: var(--emoji-size);
-	padding: 10px;
+.home-features + p::before {
+	content: "💬";
+	min-height: var(--emoji-size);
 }
 
 /* to avoid converting the markdown to HTML,
 sibling selectors are used here to select
 the first, second, and third p after .home-features */
-.home-features + p, .home-features + p + p, , .home-features + p + p + p {
-	display: block;
+.home-features + p + p::before {
+	content: "🥳";
 	min-height: var(--emoji-size);
-
-	span {
-		float: left;
-	}
 }
 
-@media only screen and (min-width: 640px) {
+.home-features + p + p + p::before {
+	content: "🔒";
+	min-height: var(--emoji-size);
+}
 
-  .home-features + p, .home-features + p + p, , .home-features + p + p + p {
-  	display: flex;
-  	align-items: center;
-  	justify-content: flex-start;
-  	white-space: nowrap;
-  }
+.home-features + p::before, .home-features + p + p::before, .home-features + p + p + p::before {
+	font-size: var(--emoji-size);
+	padding-right: var(--emoji-spacing);
+	position: relative;
+	bottom: -5px;
+}
+
+.home-features + p, .home-features + p + p, .home-features + p + p + p {
+	padding-left: 1.5em;
+	text-indent: -2.8em;
+	position: relative;
+	left: 1.5em;
 }
--- a/assets/css/styles.scss
+++ b/assets/css/styles.scss
@@ -19,24 +19,25 @@
 	--emoji-spacing: 8px;
 }
 
+/* first feature emoji */
 .home-features + p::before {
 	content: "💬";
 	min-height: var(--emoji-size);
 }
 
-/* to avoid converting the markdown to HTML,
-sibling selectors are used here to select
-the first, second, and third p after .home-features */
+/* second feature emoji */
 .home-features + p + p::before {
 	content: "🥳";
 	min-height: var(--emoji-size);
 }
 
+/* third feature emoji */
 .home-features + p + p + p::before {
 	content: "🔒";
 	min-height: var(--emoji-size);
 }
 
+/* all feature emojis */
 .home-features + p::before, .home-features + p + p::before, .home-features + p + p + p::before {
 	font-size: var(--emoji-size);
 	padding-right: var(--emoji-spacing);
@@ -44,6 +45,7 @@ the first, second, and third p after .home-features */
 	bottom: -5px;
 }
 
+/* all feature paragraphs */
 .home-features + p, .home-features + p + p, .home-features + p + p + p {
 	padding-left: 1.5em;
 	text-indent: -2.8em;
--- a/assets/css/styles.scss
+++ b/assets/css/styles.scss
@@ -14,8 +14,12 @@
 	content: unset;
 }
 
+:root {
+	--emoji-size: 40px;
+}
+
 .home-emoji {
-	font-size: 40px;
+	font-size: var(--emoji-size);
 	padding: 10px;
 }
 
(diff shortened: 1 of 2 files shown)


hpk42 (approved):
fine with me, maybe a bit larger font also?

assets/css/styles.scss
```diff
@@ -13,3 +13,13 @@
...
+
+.home-emoji {
+	font-size: 40px;
+	padding: 10px;
+}
+
+.home-features + p, .home-features + p + p, , .home-features + p + p + p {
+	display: flex;
+	align-items: center;
+}
```
r10s, line 54: there is sth wrong with that, also, this looks a bit too complicated to me, i have no idea how to maintain that without thinking much about it:

Screenshot 2022-12-15 at 14 06 59

if we do not find an easier solution, better leave the flow as is, on small screens, things will wrap as now, this is still okay.

assets/css/styles.scss
```diff
@@ -13,3 +13,13 @@
 #content h1:hover .anchor::before {
 	content: unset;
 }
+
+.home-emoji {
+	font-size: 40px;
```
r10s:
```suggestion
	font-size: 32px;
```

i'd go for slightly smaller emojis:
- this matches better the delta-chat-logo (so that emojis are not bigger than that).
- i'd go for the nicer 💬 instead of 🗨️  - and 💬 is larger on most platforms anyway.

en/index.md
```diff
@@ -5,12 +5,12 @@ lang: en
 
 # Delta Chat is a messaging app that works over e-mail 
 
-<br>
-🗨️ Message anyone with an e-mail address [even if they don't use Delta Chat](https://www.youtube-nocookie.com/embed/8LbrGXKZN70).
+<section class="home-features"></section>
+<span class="home-emoji">🗨️</span> Message anyone with an e-mail address&nbsp;[even if they don't use Delta Chat](https://www.youtube-nocookie.com/embed/8LbrGXKZN70).
```
r10s: i'd go for 💬 instead of the bit weird black bubble (it always confused me, but if it was smaller it was less visible :)

💬 also matches better in size with  🥳  and 🔒 - at least on the platforms i tried here.

r10s (changes requested):
yeah, much better in general. 

just some technical comments.

for the font-size @hpk42 suggested to enlarge: i would not change this, at least not in this pr.  let's see first if the change the font overall in https://github.com/deltachat/deltachat-pages/pull/588

assets/css/styles.scss
rosano: Should be more responsive now.

assets/css/styles.scss
rosano: Done

en/index.md
rosano: Done

r10s:
thinking it over, i am not sure if only larger emojis are worth the following disadvantages:
- all text needs to be retranslated
- the translations need to contain lots of special characters, after all experience, this will fail and we have to do a lot of refinishing - and/or the homepage will look broken in some locales (which is not immediately noticed by us)
- complicated css that needs to be understood and maybe maintained - or may become unneeded but forgotten completely with the next text update
- there are platforms where the emojis are not _that_ great, having them enlarged, well, maybe it is okay in normal size. this is quite usual for emojis, i mean, they are not that important or special here.

linebreaks will maybe be avoided by #588, so that alone is not a reason - and the emojis in normal size are "good enough" with simple linebreaks.

but if we can have the same effect by _only_ css, that would be easy to merge without the disadvantages mentioned above (apart of the 3rd).

so, maybe just remove the now superfluous linebreak (was needed when the logo was shown on the page) and use the nicer chat bubble.

rosano:
>     * all text needs to be retranslated
> 
>     * the translations need to contain lots of special characters, after all experience, this will fail and we have to do a lot of refinishing - and/or the homepage will look broken in some locales (which is not immediately noticed by us)
> 
>     * complicated css that needs to be understood and maybe maintained - or may become unneeded but forgotten completely with the next text update

I think these three points have to more do with the text being written in markdown where more sophisticated parsing is needed to extract the 'source strings'. The tradeoffs between affordances for:

- writing (by hand)
- extraction of source strings
- eventual translation by humans with a separate system
- rendering by Jekyll
- adjusting the presentation / design

means that we're likely to miss something. In this case to, I would have preferred to structure the text with HTML so there is a cleaner separation between content and presentation, and so that the translations would never have conflicts again, but it would have resulted in maybe something less elegant for writing by hand. I also used that 'complicated' CSS as a way to maintain the Markdown (sorry I documented this eventually in the CSS but forgot to explain this on your suggestion earlier)—I don't like it either, it was about balancing various tradeoffs.

Generally, I'd rather have extensive and unbreakable translations + more control over presentation than something optimized for writing by hand.

>     * there are platforms where the emojis are not _that_ great, having them enlarged, well, maybe it is okay in normal size. this is quite usual for emojis, i mean, they are not that important or special here.
> 
> linebreaks will maybe be avoided by #588, so that alone is not a reason - and the emojis in normal size are "good enough" with simple linebreaks.

I agree they're not important but as a front page it looks subdued when the 'main' points are supported by something not so easy to see. They serve the purpose of icons that might illustrate the text in a feature list, which would normally be larger than the size of the text. Not sure I would say it's 'good enough' if people have to squint, but it's also not a primary thing. 

> but if we can have the same effect by _only_ css, that would be easy to merge without the disadvantages mentioned above (apart of the 3rd).
> 
> so, maybe just remove the now superfluous linebreak (was needed when the logo was shown on the page) and use the nicer chat bubble.

I restored the original markdown, removed emojis, and added them via pure CSS. Ultimately I think any change here might cause a translation mess, so we could just close the PR and move on, but for sure worth thinking about how to structure the source strings to avoid this in general as I'd say it's a larger problem.

r10s:
> Ultimately I think any change here might cause a translation mess, so we could just close the PR and move on, but for sure worth thinking about how to structure the source strings to avoid this in general as I'd say it's a larger problem.

yeah, let's go for lower hanging fruits :) there are enough :)

rosano:
> Ultimately I think any change here might cause a translation mess

Actually I just realized that it might be within reach to batch search/remove the three emoji characters from all translations, as that's the only diff. If you're open to collaborating on that I'm happy to do some grunt work.

hpk42:
On Fri, Dec 16, 2022 at 00:12 -0800, Rosano wrote:
> Generally, I'd rather have extensive and unbreakable translations + more control over presentation than something optimized for writing by hand.

+1

rosano:
> Ultimately I think any change here might cause a translation mess

@r10s Actually I just realized that it might be within reach to batch search/remove the three emoji characters from all translations, as that's the only diff. If you're open to collaborating on that I'm happy to do some grunt work. And maybe even somehow editing the source on transifex could have some affordances to automatically find and change corresponding texts. Could be worth learning if that's possible for other contexts.

r10s:
> Actually I just realized that it might be within reach to batch search/remove the three emoji characters from all translations, as that's the only diff. 

yeah, it was not really complicated, there is even a description in the README i noticed :)

>  Generally, I'd rather have extensive and unbreakable translations + more control over presentation than something optimized for writing by hand.

what do you suggest concretely? in general, the support of markdown in transifex is poor - therefore the things with the .po files, if _that_ part is smarter somehow, that'd be great, maybe weblate does a better job there, we never really tested that.

we could also move away from markdown, having html-layout and strings inserted, as done eg. for the download page -- however, that would limit what writers can do, resp. make that part more complicated, adding more burdens and make maintenance even harder. so, i am a bit sceptical if that would hit the correct set of tradeoffs.

rosano:
> having html-layout and strings inserted, as done eg. for the download page

Wow I didn't know this was there, that's how I normally do it.

> what do you suggest concretely? in general, the support of markdown in transifex is poor - therefore the things with the .po files, if _that_ part is smarter somehow, that'd be great, maybe weblate does a better job there, we never really tested that.
> 
> we could also move away from markdown, having html-layout and strings inserted, as done eg. for the download page -- however, that would limit what writers can do, resp. make that part more complicated, adding more burdens and make maintenance even harder. so, i am a bit sceptical if that would hit the correct set of tradeoffs.

I agree with this analysis. I might correct what I said earlier as alas the site will mostly be written by hand, and thus the representation might be better optimized for writing by hand, but it's still possible to gradually move more and more into inserted strings, especially if it's not paragraph text (like the FAQ). When I made my comment about 'extensive and unbreakable translations', I think I meant we could try overdoing the inserted strings, but not necessarily using them 100% of the time.

Practically speaking, I would approach it on a string-by-string basis. I'm not sure how the strings are extracted from Markdown/HTML, but my reflex would be to maintain single, uninterrupted lines (as one would do in YAML) even when they're in the code: so aiming for pure text in Markdown, or if it needs to be wrapped in HTML then put the enclosing tags on their own lines; we'd have to experiment to see if this would make any difference with Transifex, or even Weblate. In specific cases like this pull request, it might be possible to start this process by at least 'rescuing' these three lines, as they're really more presentation than documentation.

PR #594, opened 2022-12-23
IRC and mailing list are not used really -- link to groupsbot instead

hpk42:
this strikes the link to IRC and (non-dc) mailing list -- we don't really use it anyway.  
It attempts to link to the groupsbot .... maybe someone could test and make sure the link to the groupsbot makes sense? Feel free to directly fix/commit to this branch.  thanks!

### DIFF
--- a/en/contribute.md
+++ b/en/contribute.md
@@ -7,22 +7,18 @@ lang: en
 
 For community interactions around Delta Chat please read our [Community Standards](community-standards).
 
-- [Fediverse](https://chaos.social/web/@delta) for announcements;
-  [leaving twitter](https://twitter.com/delta_chat/status/1603771336060436483).
+- [Fediverse](https://chaos.social/web/@delta) for announcements and feedback.
+
+- With your Delta Chat app you can send a `/list` message to [a community
+  groups bot (groupsbot@hispanilandia.net)](mailto:groupsbot@hispanilandia.net) to discover and join various chat groups.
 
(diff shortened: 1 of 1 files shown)


en/contribute.md
```diff
@@ -7,22 +7,18 @@ lang: en
 
 For community interactions around Delta Chat please read our [Community Standards](community-standards).
 
-- [Fediverse](https://chaos.social/web/@delta) for announcements;
-  [leaving twitter](https://twitter.com/delta_chat/status/1603771336060436483).
+- [Fediverse](https://chaos.social/web/@delta) for announcements and feedback.
+
+- With your Delta Chat app you can send a "/list" message to [the community
```
r10s:
```suggestion
- With your Delta Chat app you can send a `/list` message to [the community
```

that way one does not have to explain not to enter the quotes :)

r10s (approved):
nice, thanks for cleaning up!

en/contribute.md
```diff
@@ -7,22 +7,18 @@ lang: en
...
+  groups bot](mailto:groupsbot@hispanilandia.net) to discover and join various groups.
```
r10s: on android, the links should be tappable and open dc, on desktop/ios, it will in most cases open the system mail app, so, maybe additionally show the address? we could maybe also use an invite-link, that, however, currently always initiates a secure-join. 

however, maybe it is good enough to be merged, we can iterate from there ...

PR #595, opened 2022-12-23
bump android version to 1.34.8

r10s:


missytake (approved):
Lacking a proper device to try out the apk, but the download link works.

PR #597, opened 2022-12-25
[WIP] Margin and float correction on small screens

### LINKED ISSUE #596: Image and text alignment fails for screen widths up to 640 pixels
- Expected behavior:
For screen widths up to 640 pixels, the text of a blog article is aligned to the bottom margin of an image, but not positioned in the same line, and the image itself is aligned to the left margin:
grafik

- Actual behavior:
For screen widths up to 640 pixels, the text of a blog article is aligned to the right margin of an image, and the image itself is not aligned to the left margin:
grafik

-- Load a blog article that includes an image in your browser, e.g. https://delta.chat/en/2022-12-15-uidevjob .

- Screenshots:
See above.

- Logs:
N/A.

gerryfrancis:
Correction of left/right margin and float of image on smaller screens up to 640px.

Closes https://github.com/deltachat/deltachat-pages/issues/596 .

### DIFF
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -47,11 +47,13 @@ body {
     max-width: 100% !important;
     max-height: 600px; /*avoid portrait smartphone screenshots becoming too big*/
     height: auto;
+    display: block;
 }
 
 @media only screen and (max-width: 640px) {
 	#content p img, iframe {
 	    float: none !important;
+	    margin: 0 0 1em !important;
 	}
 	.post-box {
 	    margin-top: 4em;


gerryfrancis:
I set this PR to `[WIP]`, as I recognized that the changes have influenced the display of images that are positioned on the landing page, `delta.chat`, a lot due to an incoherent insertion. (E.g., one time two `` HTML tags exist in one `<p>` tag, then two times one `` tag in two `<p>` tags each. The four `` tags of the screenshot gallery should be placed in one single `<p>` tag.)

Therefore I suggest to edit the landing page adequately first, before this PR is going to be merged. @hpk42 @r10s As I do not have access to the CMS and cannot edit the page by myself, someone else might be entitled to do that work. (Alternatively you may want to let me know how it is done, and I can do the necessary changes there.) Thank you! :)

r10s:
> As I do not have access to the CMS and cannot edit the page

there is no CMS other than github where you already suggesting changes :)

sorry, for the very late reply.

however, in general, i do not see that we adapt the landing page to fix a minor things with the blog entry - but we can maybe do this image-block-thingie with fewer side effects and more on-point, so only for the blog-page? but i am also no expert in  this area ...

gerryfrancis:
> sorry, for the very late reply.

@r10s Never mind, we all have been very busy! :)

> however, in general, i do not see that we adapt the landing page to fix a minor things with the blog entry

@r10s AFAICR, some images on the landing page are already framed with a white margin, and some are not, which is bad, because you cannot put them in a row easily and would have to work around it. However, ...

> there is no CMS other than github where you already suggesting changes :)

@r10s ... now that I know, I can start getting into it by myself. :) Will try out some things in the Firefox Inspector, and do some suggestions based on my findings later on.

hpk42:
not sure but i think that #673 already helped quite a bit.  If you have other findings, please re-open or create a new PR. thanks!

gerryfrancis:
@hpk42 Thank you, I totally forgot about this one, sorry! Very happy to see a solution, will check it out! :)

r10s:
ftr: see #672 and updated README for more detailed explanations

PR #599, opened 2022-12-29
update translations

r10s:


r10s:
there are issues with the redirect, at least in the preview, eg. https://staging.delta.chat/599/ru/donate redirects to https://staging.delta.chat/599/en/en/en/en/en/en/en/en/en/en/en/en/en/en/en/en/en/en/en/en/ru/donate - we had that in the past several times, so, not sure if that is preview-only or a real issue.   we should really try to get a grip on these redirects at some point, there is some weird loop, eg. also the redirect for a non-existing file on master looks wrong (https://delta.chat/this-does-not-exist - two "en" seems one too much - and that seem to escalate somehow, we've seen this "en/en/en/en/en/en/en/en/en/en/en/en/en/en/en/en/en/en/" also on master)

tl;dr - maybe just merge that in and hotfix issues on master (so, the one that merges this, should be able to act :)

zh_CN/index.md
```diff
@@ -23,7 +23,7 @@ style="float: left; margin: 10px;display: block;box-shadow: 5px 5px 2px #777;" a
 
 <img src="../assets/blog/screenshots/2020-01-09-delta-chat-iOS-weekend-group-chat.png" width="110" style="margin: 10px" alt="A screenshot of Delta Chat on IOS" /> 
 
-<a class="download-button" href="https://get.delta.chat">下载</a>
+[Download](https://get.delta.chat){: .cta-button}
```
r10s, line 26: sidenote, no blocker:

retranslations aside - not sure, btw. if the markdown syntax is better and more widely understood than  html here. maybe it is just me, for whom this is new :)

r10s:
also the footer is missing the "Donate" link for translations:

Screenshot 2023-02-01 at 14 19 52

r10s:
will do a new run, with some changes wrt donate first

PR #608, opened 2023-02-01
clarified that only mails Delta has seen get deleted

missytake:
related https://github.com/deltachat/deltachat-core-rust/issues/4003

@r10s this is now a bit sharper; do we want to merge this or are we unsure what we can guarantee? OTOH, we can't *guarantee* that we delete all mails from server - someone could also rely on that.

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -228,7 +228,9 @@ to mute a chat, use the chat's menu (Android/Desktop) or the chat's profile (iOS
   delete them there, too.
 - To turn it on, go to "delete old messages from server" in the "Chats & Media"
   settings. You can set a timeframe between "at once" and "after a year"; this
-  way, *all* e-mails will be deleted from the server after that timeframe. 
+  way, all e-mails Delta Chat has seen will be deleted from the server after
+  that timeframe. If you have turned on "No, chats only" under "Settings > 
+  Chats and Media > Show Classic E-Mails", normal e-mails won't be deleted.
 - Note that if you use Delta Chat on more than one device, you should leave the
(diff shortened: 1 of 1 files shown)


en/help.md
```diff
@@ -228,7 +228,9 @@ to mute a chat, use the chat's menu (Android/Desktop) or the chat's profile (iOS
   delete them there, too.
 - To turn it on, go to "delete old messages from server" in the "Chats & Media"
   settings. You can set a timeframe between "at once" and "after a year"; this
-  way, *all* e-mails will be deleted from the server after that timeframe. 
+  way, all e-mails Delta Chat has seen will be deleted from the server after
+  that timeframe. If you have turned on "No, chats only" under "Settings > 
+  Chats and Media > Show Classic E-Mails", normal e-mails won't be deleted.
```
r10s, line 233: this last sentence may make things more unclear, what are "normal e-mails"? if the user has a different interpretation than we (replies ...), things are lost.

the "seen by Delta Chat" is maybe also vague, but here we say, things are deleted, so misinterpretation would not lead to data loss. otoh, this seems to be clear enough that not _all_ emails are deleted, if one looks from that perspective.

maybe also use shorter sentences:

```
- To turn it on, go to "Delete old messages from server" in the "Chats & Media" settings.
  You can set a timeframe between "at once" and "after a year". 
  All e-mails seen by Delta Chat will be deleted from the server after that timeframe. 
```

maybe we should sync the faq also a bit more to what we say in the app (which is still pretty good, there was loooooots of discussions about every word of the confirmation dialog :) - but that is a larger refactoring.

PR #609, opened 2023-02-02
add winget command to download page

Simon-Laux:
this command installs the official release from the Microsoft Windows store


r10s:
reminds me to the 90ies and `regedit.exe` where you also needed to pass weird numbers often :)

r10s (approved):

PR #612, opened 2023-02-09
made the part about server_name better readable

missytake:
We received this feedback a lot during our OFFDEM workshop.

Anything else to fix?

### DIFF
--- a/en/serverguide.md
+++ b/en/serverguide.md
@@ -122,8 +122,11 @@ After that we need to run `printf "#\n" > data/conf/dovecot/global_sieve_before`
 ### Mailadm NGINX config
 
 `mailadm.example.org/new_email` needs to be reachable for HTTP requests to
-work. So add `mailadm.example.org` to `data/conf/nginx/server_name.active`,
-then add the following block to `data/conf/nginx/site.mailadm.custom`:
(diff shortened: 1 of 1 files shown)


missytake:
These changes should be tried out before they can be merged - best on Saturday.

hpk42 (commented):
lgtm, thanks!

en/serverguide.md
```diff
@@ -293,13 +299,61 @@ You can setup ufw as a firewall to protect mailadm against DoS attacks:
...
+you can disable POP3.
+
+To do this, add the following to `mailcow.conf`:
+
+```
+POP_PORT=127.0.0.1:110
+POPS_PORT=127.0.0.1:995
+```
+
+Then apply the changes with `sudo docker compose up -d`.
```
ralphtheninja, line 307: Maybe use `sudo docker compose restart` instead? Assuming the service has been started. I don't know if you can just restart a single service like below or if you need to restart everything.

missytake: In this scenario, `docker compose up` will re-run all containers whose definition was changed by editing those variables. Which is pretty much exactly what we want. `docker compose restart` would just re-start all existing containers, maybe even without changing their configuration, it's not what we want here.

PR #621, opened 2023-03-15
Add comments via the Fediverse

missytake:
Using this guide: https://joelchrono12.netlify.app/blog/how-to-add-mastodon-comments-to-jekyll-blog/

What's still missing:

- [x] make "load comments" button look nice
- [x] make "reply on fediverse" dropdown menu look nice
- [ ] optional: move reply-toots a bit to the right so you can see who is replying to whom
- [x] use most recent version of https://github.com/cure53/DOMPurify


missytake:
hm, line 14 in `_includes/comments.html`:

```
            document.getElementById('comments-list').innerHTML += "<h3>Mastodon Comments</h3>";`
```

apparently gets translated by jekyll to:

```
            document.getElementById('comments-list').innerHTML += "
    
      <h3>
        
        
          Mastodon Comments
        
        
      </h3>";
```

This throws an error in the javascript console:

```
Uncaught SyntaxError: "" string literal contains an unescaped line break
```

Any idea on how to deal with this?

missytake:
https://stackoverflow.com/questions/66217925/uncaught-syntaxerror-string-literal-contains-an-unescaped-line-break suggests that we might need to escape a character in `"<h3>Mastodon Comments</h3>"` - I'll try with the hashes.

r10s:
nice! for:

> make "load comments" button look nice

i would not do much there, we're not using much styling for these things at other places, so i would not start here.

i'd use the following in a line, with quite some whitespace before / after:

```
MASTODON-ICON   Show Comments
```

the "how to reply" instructions/link can be added below the last comment on expandion. the  MASTODON-ICON can be the same as in the footer.

_includes/comments.html
```diff
@@ -0,0 +1,69 @@
...
+<script src="../assets/js/purify.js"></script>
+<script>
+    document.getElementById('load-comment').addEventListener("click", async () => {
+        document.getElementById('load-comment').remove();
+		    const response = await fetch('https://{{ mastodon_instance }}/api/v1/statuses/{{ page.com_id }}/context');
+        const data = await response.json();
+        let commentsList = document.getElementById('comments-list');
+        if (data.descendants && data.descendants.length > 0) {
+	    let commentsHeadline = document.createElement('h2');
+	    commentsHeadline.innerText = 'Mastodon Comments';
```
r10s: not sure about the wording common in the fediverse, but it seems a bit confusing to have "mastodon" as one example in the "reply" hint - and talk about "mastodon" instead of fediverse here. 

maybe just say "Comments" as in the original?

_includes/comments.html
```diff
@@ -0,0 +1,69 @@
+{% assign mastodon_instance = "chaos.social" %}
+{% assign mastodon_user = "delta" %}
+<link rel="stylesheet" href="../assets/css/comments.css">
+<div class="article-content">
+<br>
+<div id="comments-list" style="border-top: 1px solid #BBB;"></div>
```
r10s, line 6: i think, the lines, borders and spacings needs some more love, but it is good enough, no blocker. i also have no "quick" hint, that would need time of playing with the layout

r10s (approved):
nice functionality and a great improvement! thanks a lot!

i styled the "comment button" with the last commit a little, however, i probably do not find the time these days to improve styling further (whitespace, lines, borders) - but it is good enough to get merged, i'd say - esp. as the "link cluttering" below the post was also not nice before.

PR #623, opened 2023-03-24
Mention multiple accounts

### LINKED ISSUE #568: FAQ: drop some words about multi account
afaik, it is not even mentioned. we should at least say, that it exist (it is not obvious on all systems, which is fine, as not needed from the very start. but if one eventually reads trough FAQ, one should get the information that there is sth. as multi account :)

rosano:
Closes #568.

Would be helpful if someone can verify that the instructions are suitable for Android as I don't have one.

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -36,6 +36,7 @@ recipient to install the same app as yours, as with other messengers.
 - _Reliable_ - safe for professional use.
 - _Trustworthy_ - can even be used for business messages.
 - _Copyleft_ and _Standards_ based libre software.
+- _Flexible identity_ with built-in support for [multiple accounts](#multiple-accounts)
 
 
 ### Which messages do appear in Delta Chat?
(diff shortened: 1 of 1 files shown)


rosano:
I would also eventually change/hardcode all the headings to use shorter English anchors in every language as I believe the current approach requires writing a different link (often quite long) in each language.

brabo:
Confirmed for Android!
I like that it also links to multiple devices. Nice!

en/help.md
```diff
@@ -76,6 +77,13 @@ Media -> Show Classic E-Mails", you can change this. You have these options:
 
 - Yes. Beside the plain text, all e-mail attachments are displayed as separate messages. Outgoing messages get attachments as needed automatically.
 
+<h3 id="multiple-accounts">How can I add or switch between multiple accounts?</h3>
```
missytake, line 80:
```suggestion
### How can I add or switch between multiple accounts?
```

This adds an anchor link automatically - let's keep it consistent :) We could see in the staging preview which link is created and fix it above.

But it's a minor thing, we can also just merge it as it is.

missytake (approved):

PR #625, opened 2023-03-24
Document how to send original images

### LINKED ISSUE #619: update help: describe how to send images as original
send images scales them down to a reasonable bytesize.

to send images as original, including metatdata, one has to send them as "file".

rosano:
Closes #619


brabo:
The addition seems to imply that sending as an image does affect the metadata. Is this the case? If so, in which way(s)?
Might be nice to be explicit there IMHO.

en/help.md
```diff
@@ -76,6 +76,8 @@ Media -> Show Classic E-Mails", you can change this. You have these options:
 
 - Yes. Beside the plain text, all e-mail attachments are displayed as separate messages. Outgoing messages get attachments as needed automatically.
 
+- For performance, images are optimized and sent at a smaller size by default, but you can send it as a "file" to preserve the original dimensions and metadata.
```
```suggestion
- For performance, images are optimized and sent at a smaller size by default, but you can send it as a "file" to preserve the original.
```

this seems to imply, that otherwise metadata are _always_ removed, which is currently only the case for _most_ images.

also, not only "dimensions and metadata" are preserved - but also the file format. the whole thing will be left as is, no bit is changed.

r10s (dismissed):

r10s:
> The addition seems to imply that sending as an image does affect the metadata. Is this the case? If so, in which way(s)?
Might be nice to be explicit there IMHO

it is the case for most images, but there is no guarantee - small jpg may be sent as is, as well as images in other formats. this may change in the future, https://github.com/deltachat/deltachat-core-rust/pull/4037 , but just now, we cannot be concrete wrt metadata, so i'd just leave that aspect out.

rosano:
@r10s I added your suggestion and resolved a merge conflict. Still needs approval.

r10s (approved):

r10s:
@rosano thanks a lot, i directly merged it to move forward

PR #627, opened 2023-03-25
Update default for Show Classic E-Mails

rosano:
If there's a better way to arrange these changes to avoid triggering translations let me know.

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -40,20 +40,17 @@ recipient to install the same app as yours, as with other messengers.
 
 ### Which messages do appear in Delta Chat?
 
-By default, Delta Chat shows:
+By default, Delta Chat shows all e-mails.
 
-- Messages sent by other Delta Chat users
-- Replies to your Delta Chat messages, even if they are "normal e-mails".
-
-Other e-mails don't appear in your app by default. At "Settings -> Chats &
+At "Settings -> Chats &
 Media -> Show Classic E-Mails", you can change this. You have these options:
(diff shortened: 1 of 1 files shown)


en/help.md
```diff
@@ -40,20 +40,17 @@ recipient to install the same app as yours, as with other messengers.
...
 Media -> Show Classic E-Mails", you can change this. You have these options:
 
 - "No, chats only": Only messages sent by other Delta Chat users and replies to
   your Delta Chat messages are shown. This makes most sense if you use the same
-  e-mail account for normal e-mails as well. This is the default setting.
+  e-mail account for normal e-mails as well.
 - "All": Delta Chat shows all e-mails that are sent to your email address. This
   makes sense if you want to use Delta Chat for all your e-mails, so no message
-  gets lost.
+  gets lost. This is the default setting.
```
missytake, line 53:
```suggestion
  gets lost. This is the default setting since Delta Chat 1.36.
```

missytake (approved):
This default changes at 1.36, I think? Then we should mention this (at least for a while), people might wonder why they have "no, chats only" turned on.

But if you disagree, just merge it as it is (as soon as 1.36 is published).

en/help.md
hpk42: i think with desktop 1.34.4 or so it's already the default so not sure we need to be too precise here.

hpk42 (approved):

PR #645, opened 2023-04-05
offline first: add a hint that the help is also available for relaxed, in-app reading when being offline

r10s:
just came to my mind, pointing out a bit that the help is really available offline - which is not natural in maaaany other apps:

Screenshot 2023-04-06 at 12 56 08

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -8,8 +8,6 @@ header: Frequently Asked Questions
 
 ## What is Delta Chat?
 
-> In a hurry? You can read this help relaxed at anytime within the app, also without internet.
-
 Delta Chat is a new chat app that sends messages via e-mails, encrypted if possible, with Autocrypt. 
 **You do not have to sign up anywhere, just use your existing e-mail account with Delta Chat.** 
 
--- a/en/help.md
+++ b/en/help.md
@@ -8,6 +8,8 @@ header: Frequently Asked Questions
 
 ## What is Delta Chat?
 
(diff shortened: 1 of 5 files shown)


en/help.md
```diff
@@ -8,6 +8,8 @@ header: Frequently Asked Questions
 
 ## What is Delta Chat?
 
+> In a hurry? You can read this help relaxed at anytime within the app, also without internet.
```
rosano:
```suggestion
> You can also read this help relaxed at anytime within the app, and it works without internet.
```

I would remove the 'hurry', and slightly reword.

rosano (approved):
I think it would be more helpful to move it before the table of contents so that scrolling isn't necessary.

r10s:
> I think it would be more helpful to move it before the table of contents so that scrolling isn't necessary.

+1 - however, there was an issue with that,  but i also did not try hard. wanted some general feedback first, if the hint makes sense at all :)

en/help.md
r10s: hm, the gist was also to be less formal than the rest of the page

rosano: Not a blocker

r10s:
found a way to move it up, via strings-yaml-file - this has also the advantage that no adaptions are needed for the offline help creation

missytake (approved):
Love it! Was wondering about whether you want that sentence in the in-app-help as well :upside_down_face:

adbenitez (approved):

PR #647, opened 2023-04-05
add Brazilian Portuguese

r10s:
this was missing, although significantly translated meanwhile

### DIFF
--- a/_data/lang/pt_BR.yml
+++ b/_data/lang/pt_BR.yml
@@ -1,0 +1,45 @@
+pt_BR:
+  
+  meta:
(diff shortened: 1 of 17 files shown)


missytake (approved):
There is still a lot untranslated, but probably actually publishing the translations encourages translators :) And it's still helpful I guess

missytake:
I prepared the change for the language redirect, will report here as soon as it works.

missytake:
language redirect seems to work, test case exists :)

r10s:
cool, thanks a lot!

Issue #649, opened 2023-04-06
avoid relative links in the offline help


r10s:
relative links in the help needs to be converted to absolute links, this is easily done, but also easily overseen or done wrong (even on delta.chat, see #648).

we should error out in create-local-help.py if there is any link at the end not startiing with `https` or `#` or are otherwise whitelisted.

missytake:
It has advantages to use relative links in the *online* help, because then the link actually directs to the translated page instead of the english page, if the link is not translated correctly as well (and we can absolutely not rely on that).

Maby the script can convert the relative links to absolute links?

r10s:
> Maby the script can convert the relative links to absolute links?

yip, the script is already doing that, eg. for `serverguide`.

but the blog links are not present in other languages, so, for them, an absolute path does not really hurt (and an relative one is even wrong unless using `../en/blog-url`

however, i also aim to check the links in the script more carefully, so that we do not manifest errors in the offline help. the help should just be able to use all kind of _working_ links, the script will convert them.

r10s:
fixed by #651

PR #650, opened 2023-04-07
Fix blog link in help

### LINKED ISSUE #648: some links from help to blogposts are 404 in translations
the reason is that the links are relative - but there is no blog in the translations.

example: from https://delta.chat/de/help#wurde-delta-chat-unabhängig-auf-sicherheitslücken-geprüft: "You can read more about the audit [on our blog](https://delta.chat/de/2023-03-23-third-independent-security-audit)" <- this "on our blog" does not work

same for the link to `2023-01-26-upcoming-mail-server-workshops` - and maybe more. 

at best, the links should be full absolute, incl. https - this also make it easier for the offline help.

EDIT: both examples are 404 even in english, due to wrong dates in the link

missytake:
closes #648

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -714,7 +714,7 @@ it](https://delta.chat/en/2022-09-14-aeap).
 
 - Most mail servers will work well. But what we personally recommend is a
   combination of mailcow and mailadm, as described [in this
-  blogpost](2023-01-26-upcoming-mail-server-workshops).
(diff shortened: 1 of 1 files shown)


r10s:
ah sorry, you were on that as well :)

i aim to check links more carefully in #649 , so that we will notice such errors at least before they get manifested in the offline help.

PR #658, opened 2023-04-20
Tradução index.md para Portugues do Brazil

marcianovc:
Estou começando a traduzir o site para portugues do Brazil


r10s:
hi @marcianovc thanks you so much - translation are done over transifex, however, i assume you notices that - i have just approved an account with the same name on transifex.

as i was on that, i also copied the strings from here to transifex, so that the work is not lost :)

PR #660, opened 2023-04-22
flathub verification

Simon-Laux:
I hope that this works already without help from our sysadmins


Simon-Laux:
just ran it in a codespace, Jekyll already does the right thing, let's merge it, if it doesn't work then, then it's probably some nginx configuration issue

Simon-Laux:
somehow worked in production :) 🎉

PR #662, opened 2023-04-26
update public rate limits

missytake:
after updating the table, it needs to be changed on the website as well.


missytake:
hm, the links are not clickable so far...

hpk42:
while this PR is fine because it's already there i wonder
if we really want to keep updating the limits. 
do you thoughts on this?

missytake:
> while this PR is fine because it's already there i wonder if we really want to keep updating the limits. do you thoughts on this?

Only this last time, because the data was already collected. Other than that, the eppdperf repository is now archived.

PR #666, opened 2023-05-19
docs: tell in the readme that Jekyll should be run inside bundler

link2xt:


link2xt:
Would be nice if you also confirm that it now works for you and website is built and can be browsed locally.

adbenitez (approved):

PR #665, opened 2023-05-19
add fourth-security-audit blogpost

adbenitez:

PR #668, opened 2023-05-20
ci: build preview without Docker

link2xt:
This way we do not ignore the error if the build fails. Otherwise we only check that Docker succeeded to run the container, while the build itself may fail.


r10s (approved):
the gist of this or is to make errors easier to recognize, so that things as  https://github.com/deltachat/deltachat-pages/pull/665/commits/2e0f6319c22b9dc6935c474de239bddc852ab94a are easier fixable?

a message "Check out the page preview at ..." is missing, not sure if this is releated to this pr, however. we can always improve in subsequent pr :)

all in all, lgtm, thanks a lot for taking care!

link2xt:
> a message "Check out the page preview at ..." is missing, not sure if this is releated to this pr, however. we can always improve in subsequent pr :)

This seems to be disabled by default if posting a link to run details succeeded. We can make it unconditional, but I would do this in another PR.

PR #669, opened 2023-05-20
ci: post preview link to comments unconditionally

link2xt:


r10s (approved):
nice!

i find that useful, esp. for ppl looking at a prs more occasional, the link somewhere in the ci output is easily overseen. i also like that one can edit the post and point eg. directly to the affected page.

missytake:
Before, the github actions bot only posted the link to the PR thread once. Now it does it once for every push, afaict - was this the purpose of this PR?

Do we really need that? I mean the several bot comments in this thread have all the same text. Just having one comment as we had before sounds more reasonable.

Especially blog post PRs tend to have dozens of commits, and a lot of back and forth. Having the bot writing a comment for each of them, and sending out a lot of emails because of that is spammy imho.

link2xt:
The problem was that the bot did not post the link at all, it only updated the details without posting anything. Posting only once would be nice.

PR #670, opened 2023-05-20
ci: upload preview using rsync

link2xt:
rsync is faster than scp.


link2xt:
13 seconds instead of 1m 48 seconds on reupload, e.g. small change to a PR.

link2xt:
I also replaced broken deletion script, but I still don't think it is reliable enough: better delete old previews with a cron script rather than when PR is merged.

r10s (approved):

PR #672, opened 2023-05-22
improve mobile-layout: do not start text right of image

r10s:
this simple line break ensures, the text on mobile is not starting directly after the image: before/after:

 

on non-mobile (with larger screen widths), things are floating as expected:

<a name=gist> 

as there is always some confusion about images in blogs:

# this is the gist of the current CSS wrt images

- create images with reasonable sizes, they can be large (say, up to 400px width or so)  to look good on desktop. let them float left/right or not at all (so ``, `` or just `` (the latter is recommended for images > 400px)

- unless you are forcing things using `!important` (which is not recommended at scale) on smaller screens, this floating turns to "own paragraph for the image" - but for that, an **empty line** is important

that's it, that way, images are not _that_ bad on desktop as well as on mobile. the approach of turning floating to non-floating is used in many blogs afaik. of course, things can be improved :)

i think, we should add this information also to the readme

### DIFF
--- a/_posts/2023-05-22-webxdc-security.md
+++ b/_posts/2023-05-22-webxdc-security.md
@@ -227,6 +227,7 @@ that try to identify VPN users.
 ### Maybe using Firefox engines would help us? 
 
 <img src="../assets/blog/2023-05-20-allchrome.png" width="270" style="float:right; margin-left:0.5em;" />
+
 Delta Chat apps do not use Firefox webviews which can be directly configured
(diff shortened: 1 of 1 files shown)


adbenitez (approved):
fixing a bug with an empty line, like a PRO :sunglasses:

hpk42:
Thanks.  I usually look at the last post for layout inspiration so hopefully get it right next time

PR #682, opened 2023-06-06
Update minimum desktop os versions

Simon-Laux:
could be that older versions of ubuntu still work, but this is the LTS release. I also guessed a bit for fedora based on release dates. min macOS comes from https://www.electronjs.org/de/docs/latest/tutorial/unterstützung

### DIFF
--- a/en/download.md
+++ b/en/download.md
@@ -16,7 +16,7 @@ The desktop versions do not require Delta Chat to be installed on a phone.
 Minimal requirements:
 Android 4.1 Jelly Bean
 or iOS 11, iPhone 5s or iPad 5/Air/Mini
-or Windows 7, macOS 10.11 El Capitan, Ubuntu 20.04, Fedora 29 or Debian 10
+or Windows 7, macOS 10.11 El Capitan, Ubuntu 18.04, Fedora 29 or Debian 10
 or compatible systems.
 
 ## Links
--- a/en/download.md
+++ b/en/download.md
@@ -16,7 +16,7 @@ The desktop versions do not require Delta Chat to be installed on a phone.
 Minimal requirements:
 Android 4.1 Jelly Bean
 or iOS 11, iPhone 5s or iPad 5/Air/Mini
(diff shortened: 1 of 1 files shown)


r10s:
i'd check ubuntu in more detail, requiring a barely 6 month old OS as a "minimum requirement" and skipping 10 years might confuse people

iirc @hpk42 is using always quite some old ubuntu stuff, or ask in the dev channel.

the link says ubuntu 14.04, Fedora 24 and Debian 8 are confirmed to work as well - unless we get reports that this is not true, we can also stay with that. that looks to me also as more reasonable jumps.

Simon-Laux:
it is mainly about glibc https://github.com/deltachat/deltachat-core-rust/pull/4448 -> so we know for certain that debian 8 won't work, because we just fixed debian 10.

sadly those fedora releases (20-30) are long not supported anymore, I didn't find a list of what glibc versions they use, so I went a bit by date, I don't plan to install all of them just to check what glibc version they use.

same problem for ubuntu, I did not find the information what glibc they use.

Simon-Laux:
with flatpak you can probably go older, those are for appimage and installers.

Simon-Laux:
ok fedora 29 seems to be at 2.28: https://bodhi.fedoraproject.org/updates/?search=2.28&packages=glibc seems like I just needed dig deeper.

Simon-Laux:
the oldest info for ubuntu seems to be 20.04 https://packages.ubuntu.com/source/focal/glibc (I don't know where to find older info)

r10s:
if debian10 works, according to https://askubuntu.com/questions/445487/what-debian-version-are-the-different-ubuntu-versions-based-on , ubuntu18.04 might work as well as that already uses debian10.

debian10 as well as ubuntu18.04 are some years old, so that would be another story

r10s (approved):
lgtm now from a broader perspective, thanks a lot for taking care.

maybe ppl actually using linux have some more detailed information, however, we can also iterate on that as neeed.

PR #688, opened 2023-06-22
webxdc import/export blogpost

### LINKED ISSUE #681: Publish a blog post about the new import/export webxdc API
- main: new Import/Export API, link to specs
- thank NLNET
- mention Calendar app as example use case (mention that notifications of events is not currently possible, would need new API to be added in the future?)
- mention Cheogram as another implementation of WebXDC

adbenitez:
close #681


r10s (approved):
great post :)

PR #689, opened 2023-06-23
update desktop to 1.38.1

Simon-Laux:


r10s (approved):

r10s:
is there anything that blocks merging this? seems 1.38.1 is out also in windows etc, so it should be available on our homepage as well :)

PR #690, opened 2023-06-27
Clarification 001

brabo:
Small clarification update!


hpk42 (approved):
ok to merge but not sure how it clarifies much: my main question would be if `server_name.active` is a single-line file with `mailadm.example.org` as content?  
If os, then maybe 

    echo "mailadm.example.org" >data/conf/nginx/server_name.active

is a more succint way to express it.

en/serverguide.md
```diff
@@ -122,8 +122,8 @@ After that we need to run `printf "#\n" > data/conf/dovecot/global_sieve_before`
 ### Mailadm NGINX config
 
 `mailadm.example.org/new_email` needs to be reachable for HTTP requests to
-work. So add `mailadm.example.org` to `data/conf/nginx/server_name.active`,
-then add the following block to `data/conf/nginx/site.mailadm.custom`:
+work. So create `data/conf/nginx/server_name.active` and add `mailadm.example.org` to this file,
```
r10s: maybe say waht is a file and what is a line more directly?

```suggestion
work. So create the file `data/conf/nginx/server_name.active` and add the line `mailadm.example.org` to it,
```

tbh, on first read, i thought about a folder :)

r10s (approved):

en/serverguide.md
```diff
@@ -122,8 +122,8 @@ After that we need to run `printf "#\n" > data/conf/dovecot/global_sieve_before`
...
+then create `data/conf/nginx/site.mailadm.custom` and add the following block to it:
```
r10s:
```suggestion
then create the file `data/conf/nginx/site.mailadm.custom` and add the following block to it:
```

en/serverguide.md
```diff
@@ -318,11 +318,11 @@ Add the following lines to each container in
...
-    logging:
-      driver: "syslog"
-      options:
-        syslog-address: "udp://127.0.0.1:514"
-        syslog-facility: "local3"
+      logging:
+        driver: "syslog"
+        options:
+          syslog-address: "udp://127.0.0.1:514"
+          syslog-facility: "local3"
```
r10s: is the indentation of two additional spaces on purpose or an edit accident?

brabo: the two spaces were missing, now it can be copy-pasted directly and have correct indentation in the end result!

brabo:
> ok to merge but not sure how it clarifies much: my main question would be if `server_name.active` is a single-line file with `mailadm.example.org` as content? If os, then maybe
> 
> ```
> echo "mailadm.example.org" >data/conf/nginx/server_name.active
> ```
> 
> is a more succint way to express it.

The clarification is that the file(s) do not exist, and have to be created. previously it gave to workshop attendees the impression the file existed, and some tried locating it and got confused.
We can indeed opt for piping content into all new files.
If we do so, probably echo without flags is the most portable way to do it. There is a bash-centric printf statement for the dovecot global_sieve_before higher up that perhaps should become an echo too then?

en/serverguide.md
r10s: great, thanks for the fix!

missytake:
> The clarification is that the file(s) do not exist, and have to be created. previously it gave to workshop attendees the impression the file existed, and some tried locating it and got confused. We can indeed opt for piping content into all new files.

Piping sounds good :)

> If we do so, probably echo without flags is the most portable way to do it. There is a bash-centric printf statement for the dovecot global_sieve_before higher up that perhaps should become an echo too then?

I think we can assume bash. ubuntu/debian with bash is what I would expect to run in such a scenario. But if it's no hassle... just use whatever you think makes sense.

brabo:
> > The clarification is that the file(s) do not exist, and have to be created. previously it gave to workshop attendees the impression the file existed, and some tried locating it and got confused. We can indeed opt for piping content into all new files.
> 
> Piping sounds good :)
> 
> > If we do so, probably echo without flags is the most portable way to do it. There is a bash-centric printf statement for the dovecot global_sieve_before higher up that perhaps should become an echo too then?
> 
> I think we can assume bash. ubuntu/debian with bash is what I would expect to run in such a scenario. But if it's no hassle... just use whatever you think makes sense.

It is not so much hassle. Bash-centric usage would use printf as it is a builtin and always identical behaviour, whereas echo has some implementation differences depending on which nix one likes to use. About assuming, we already have people doing serverguide on Alpine Linux, some people do prefer other shells.. So for very simple usage where we do not need any argument flags, echo will work for every use case.

missytake (approved):
Okay, thanks for explaining :)

r10s:
there is a merge conflict, the pr seems to need a rebase

hpk42:
i tried rebasing and merging and both gives problems in the gh interface.  Instead of trying to figure it out i just created a #694 -- it's the same change like here but a) in a single commit b) using https://sembr.org (it's a pity that the serverguide doesn't it mostly)

PR #694, opened 2023-07-03
re-do #690 in a single commit

hpk42:
supersedes and fixes #690


brabo (approved):
Good!

r10s:
i also meanwhile pushed translations and pulled them at https://github.com/deltachat/deltachat-pages/pull/700 - this is also maybe interesting to get a feeling about the impact of a few changed lines :) (in this case, serverguide is not translated to so many languages yet, however)

PR #696, opened 2023-07-05
full language names

### LINKED ISSUE #695: Use names of languages in their native script
For someone who knows English alphabets and language codes(or can guess), it's easy to find the version of DeltaChat website for their own language. But for someone who knows only their mother tongue, it's impossible. Therefore, I suggest writing name of languages in their native script in the bottom of the website.

r10s:
use full language names in the footer, the ISO codes are indeed partly unknown and weird.

names and ordering taken from https://github.com/deltachat/deltachat-android/blob/master/res/values/arrays.xml#L7 (they are mostly confirmed by the corresponding translators, not that easy always :)

moreover, the pr does a bit for formatting

closes #695 

before / after:

Screenshot 2023-07-05 at 12 59 33
Screenshot 2023-07-05 at 12 59 24


adbenitez (approved):
nice improvement!

PR #699, opened 2023-07-06
set an explicit #changelogs anchor

### LINKED ISSUE #698: HTML anchor part "#changelogs" mentioned in device message does not work for non-English websites
- Operating System (Linux/Mac/Windows/iOS/Android):
Android 11.

- Delta Chat Version:
1.38.2.

- Expected behavior:
The link `https://get.delta.chat/#changelogs` works for localized (non-English) web content, too.

- Actual behavior:
The link `https://get.delta.chat/#changelogs` does not work for localized (non-English) web content.

-- Set the language to German in Delta Chat as well as your web browser app on top for your favourite web content.
-- Open the device message ("Systemnachricht") "1.38 ist da".
-- Tap on the link `https://get.delta.chat/#changelogs`. (Result: The web browser app opens and displays the website well, but does not jump to the section "Änderungsprotokolle", the German changelog, due to the missing anchor part `#changelogs` on the site. Instead, the respective German anchor part is named `#änderungsprotokolle`.)

- Screenshots:
N/A.

- Logs:
N/A.

r10s:
that way, the anchor don't get translated and we can use sth. as get.delta.chat#changelogs that redirects to the correct headline in all languages

closes #698

### DIFF
--- a/en/download.md
+++ b/en/download.md
@@ -4,7 +4,7 @@ lang: en
 downloads: true
 ---
 
-## Changelogs
+## Changelogs {#changelogs}
 
 * [Desktop](https://github.com/deltachat/deltachat-desktop/blob/master/CHANGELOG.md)
 * [Android](https://github.com/deltachat/deltachat-android/blob/master/CHANGELOG.md)


r10s:
> Check out the page preview at https://staging.delta.chat/699

hm, does not work, EDIT: ahh, probably because translations need to up updated; this will be done outside this pr on master on a regular base, after that, things work with some chance :)

en/download.md
```diff
@@ -4,7 +4,7 @@ lang: en
 downloads: true
 ---
 
-## Changelogs
+## Changelogs {#changelogs}
```
adbenitez, line 7: isn't `[#changelogs]` the correct syntax?

adbenitez: ah this is inserting an anchor ID not a link!?

adbenitez (approved):

en/download.md
r10s: yes, it inserts an id, cmp https://github.com/deltachat/deltachat-pages/blob/master/en/help.md?plain=1#L446 and see that the link also works eg. with https://delta.chat/es/help#multiclient - even though the title is "Múltiples dispositivos" and not "Multi-client"

PR #702, opened 2023-07-27
Optimize SVG images with `scour`

link2xt:
E.g.
scour w3c_home.svg --shorten-ids --no-line-breaks --strip-xml-space --remove-title --strip-xml-prolog --enable-id-stripping


missytake (approved):
I don't see a difference, and if it's optimized now, great :)

PR #703, opened 2023-07-28
Defer loading of banner.css

link2xt:
TLS setup to cosmos.delta.chat sometimes takes too long, postponing rendering.

Defer loadin of CSS as described on
https://web.dev/defer-non-critical-css/

There is no <noscript> tag because the banner is
not created without JavaScript anyway.


rosano (approved):
I suppose we could also add the stylesheet via the original javascript? Not sure if there was a reason I didn't do that.

link2xt:
> I suppose we could also add the stylesheet via the original javascript? Not sure if there was a reason I didn't do that.

Yes, that would be nice if you do this in the cosmos repository and make CSS file blank. Then we can remove the link to CSS file here.

PR #709, opened 2023-08-09
xstore@testrun.org blogpost

missytake:
This PR publishes the blogpost on the website (I accidentally pushed it to master already, but reverted it directly; unfortunately the RSS bot already posted it to the community mailing list).

Things which are missing:

- ~~The QR verification code for messaging xstore@testrun.org~~ -> later with M3, on webxdc.org
- [x] Does interaction between @try.webxdc.org accounts and xstore@testrun.org work?
- ~~Better screenshots? preferably from Android~~
- [x] Consistent upper case for "Private Apps"
- [x] hello@try.webxdc.org should recommend xstore: https://github.com/missytake/greeterbot/commit/90a601f50c09b723dd2ca5553722064136057b81 is already committed, but deployment didn't work yet
- [x] store.xdc -> store app
- [x] private apps -> webxdc apps
- [ ] what happens if I open the store in dc 1.36?
  - downloading apps works, but clicking on the "share" button just expands the details of the app: https://codeberg.org/webxdc/store/issues/231
- [x] new featured image: 5x5 app icon tile
- [x] rework "how does it work" section
- [x] thank NLNet
- review of course
- [ ] After publishing: a toot mocking super apps by billionaires


hpk42:
- maybe we can just forget about the QR code for this blog post -- it's anyway a bit tricky if you have delta chat on the same device where you are reading the post ... so the single possible way to use it is sending "hi" to xstore@testrun.org (a clickable mailto-link that could include the "hi" body, maybe?) 
- let's not use "private apps" as the central term but "webxdc apps" and sometimes "chat-shared apps" or just "apps"

adbenitez:
> * Does the bot already send the store.xdc just after verifying?

the status of that bug is tracked here: https://codeberg.org/webxdc/store/issues/228

_posts/2023-08-10-xstore.md
```diff
@@ -0,0 +1,145 @@
...
+---
+
+<img src="../assets/blog/gordon_superapp.jpg" width="340" style="float:right; margin-left:1em;" />  
+As [Gordon Brander](https://mastodon.social/@gordon) already stated -
+with [Private Apps](help#webxdc),
+Delta Chat is more than a messenger.
+You can share Private Apps with any other Delta Chat user
+and use them collaboratively;
+for example to play chess
+or to edit your household's shopping list.
```
hpk42: - highlighting a game (after all most apps are games these days) makes sense already here 
- the first sentence if not the first paragraph should not have or require a link -- and it should rather contain the word "super app" rather than just having it in the attached image
- i'll see to provide a PR for the first para in a minute :)

missytake:
> maybe we can just forget about the QR code for this blog post -- it's anyway a bit tricky if you have delta chat on the same device where you are reading the post ... so the single possible way to use it is sending "hi" to [xstore@testrun.org](mailto:xstore@testrun.org) (a clickable mailto-link that could include the "hi" body, maybe?)

fine with me :)

> let's not use "private apps" as the central term but "webxdc apps" and sometimes "chat-shared apps" or just "apps"

hmm... that's what we are using on the FAQ page for example, I think it makes sense to use webxdc as a technical term and private apps for users. What would be your reasoning?

_posts/2023-08-10-xstore.md
```diff
@@ -0,0 +1,145 @@
+---
+title: A place where you can get Private Apps - xstore is ready!
+author: missytake
+image: ../assets/blog/2023-08-xstore-applist.png
+---
+
+<img src="../assets/blog/gordon_superapp.jpg" width="340" style="float:right; margin-left:1em;" />  
+As [Gordon Brander](https://mastodon.social/@gordon) already stated -
```
adbenitez: IMHO is better to point to the toot directly rather than the account

missytake: It was a tweet, and twitter is too broken for me right now to find it... that's why I pointed to Mastodon instead. But if you find the tweet let's include it :)

adbenitez:
> * maybe we can just forget about the QR code for this blog post -- it's anyway a bit tricky if you have delta chat on the same device where you are reading the post

meanwhile I created a PR to fix the issue in the bot so we can include the QR, about in-same-device, we can also include a link with the QR data that is just an URI supported by delta chat, also mentioning to send "hi" and the bot address with mailto link would not hurt tho, but better to suggest the qr workflow because user starts with a verified chat up-front from the very first message which is nice and without needing to type anything, which feels more magical/cool :)

missytake:
> meanwhile I created a PR to fix the issue in the bot so we can include the QR, about in-same-device, we can also include a link with the QR data that is just an URI supported by delta chat, also mentioning to send "hi" and the bot address with mailto link would not hurt tho, but better to suggest the qr workflow because user starts with a verified chat up-front from the very first message which is nice and without needing to type anything, which feels more magical/cool :)

Holger and I just discussed that maybe it makes more sense to just use a normal mailto: link, as there is much less which can go wrong, and at least typing is something users can always fall back to.

adbenitez:
> Holger and I just discussed that maybe it makes more sense to just use a normal mailto: link, as there is much less which can go wrong, and at least typing is something users can always fall back to.

the PR got merged and it works nice without any user manual interaction, you scan the QR and get the app store and protected chat, as I said in my comment I am not against the manual "write hi" option, but I think it will not hurt to also include the QR instead of that random icon of the app store (that I hope we change it at some point), just say "scan this qr" and below "or write to ... `<the manual way here>`" 

because the manual way alone sounds a bit of several steps and in fact the mailto link will cause more issues than the QR, several uses have set a classic email app to always open by default mailto links, users use to hit "remeber my decision" and then it is hard to change

link2xt:
> Or maybe also use the “add second device” feature to try out a cheap method of safely syncing notes, checklists or calendars between your devices without any servers or additional accounts.

Actually there is an email server, maybe just remove "without any servers or additional accounts".

_posts/2023-08-10-xstore.md
```diff
@@ -0,0 +1,158 @@
...
+The xstore chat bot
+collects webxdc apps from a list of “git” repository sources
+and will continously offer the latest released app versions.
+You can also self-host the store bot:
+checkout installation instructions
+in the [bot’s codeberg repository](https://codeberg.org/webxdc/store/).
+It works well together with [xdcget](https://codeberg.org/webxdc/xdcget/),
+a command line tool
+which downloads webxdc releases
+from git public repositories.
```
adbenitez, line 166: only codeberg and github at the moment tho

adbenitez (approved):

_posts/2023-08-10-xstore.md
```diff
@@ -0,0 +1,158 @@
...
+[xstore@testrun.org](mailto:xstore@testrun.org?body=hi) is a chat bot instance.
+After you send “Hi” to the bot,
+you receive and start the store app,
+which continues to communicate with the bot
+for selecting and downloading apps.
+You can not share the store app
+to another chat.
+That’s because
+webxdc apps can not access the Internet,
+only the bot.
```
adbenitez: this sentence is a bit hard to understand, maybe:

> That's beacuse the store app can not access the internet, so it is the bot the one that puts the data inside the store using special messages in your 1:1 chat with it

or something like that

missytake: better?

_posts/2023-08-10-xstore.md
missytake: true, but if someone wants to add a repository from a different server... we'd probably make it possible in the very same PR.

_posts/2023-08-10-xstore.md
```diff
@@ -58,7 +57,7 @@ scan this invitation code to get an instant e-mail test account:
 The test e-mail account will only exist 90 days
 and it does not allow to send chat mails to most "outside" e-mail addresses. 
 But if you want to test things for yourself,
-just share apps with your "Saved Messages" chat to play
+you may share apps with your "Saved Messages" chat to play
```
hpk42: we say "just" a bit too often

PR #710, opened 2023-08-09
Mention using same network when setting up a new device

ralphtheninja:


missytake:
That's already bullet point no. 1 of https://delta.chat/en/help#troubleshooting, directly below this FAQ item - I think it fits better there. We can not include everything people need to get right in the step-by-step, that's what a troubleshooting section is for.

r10s (approved):

ralphtheninja:
> That's already bullet point no. 1 of https://delta.chat/en/help#troubleshooting, directly below this FAQ item - I think it fits better there. We can not include everything people need to get right in the step-by-step, that's what a troubleshooting section is for.

If you include this step you essentially don't need it in the troubleshooting section. If you don't include it people will very likely end up in the troubleshooting section.

I really don't see the harm here.

missytake:
Point.

PR #713, opened 2023-08-10
Explicitly set the size of download badges

link2xt:
This prevents layout from shifting while the images are loaded.


r10s (approved):
thanks a lot!

Issue #719, opened 2023-08-22
Dark theme environment support


d47081:
On my environment the website looks like this

![image](https://github.com/deltachat/deltachat-pages/assets/108541346/7157ba93-8b5f-4f29-ac20-f8ac0a0d8d93)

I think, at least we need to define `background-color` as white by default or create adaptive template to make website more useful for geeks and nightly visitors

PR #720, opened 2023-08-22
Set default background color

d47081:
Dark theme environment support
Issue #719


missytake (approved):
I can't reproduce the issue, but this change doesn't make it worse, so if it works for you, great.

d47081:
For example, you can reproduce it on Ubuntu with dark theme, using Firefox.

Just checked - now it looks, fine!

PR #727, opened 2023-09-12
Revise "Encryption" section of the FAQ

link2xt:


hpk42:
Do whitespace changes trigger translations? If so we should only.fix formatting if we are changing entries.

link2xt:
> Do whitespace changes trigger translations? If so we should only.fix formatting if we are changing entries.

We can squash everything and revert whitespace changes for unchanged sections in the end.

en/help.md
```diff
@@ -376,49 +405,29 @@ not their account, you can still communicate in the 1:1 chat. ([Read more](#verd
 
 ### Does Delta Chat support Perfect Forward Secrecy?
 
-- No, OpenPGP doesn't support Perfect Forward Secrecy. Perfect Forward Secrecy
```
iequidoo, line 379: Maybe leave this explanation as is, otherwise it could be misread as Delta Chat not providing Forward Secrecy while it is possible?

link2xt: It is possible to implement some sort of key rotation on top of OpenPGP. OpenPGP is just a container for encrypted messages, you can still organize key management however you want, e.g. implement a double ratchet on top of it: https://gitlab.com/sequoia-pgp/openpgp-dr

iequidoo (approved):

en/help.md
```diff
@@ -263,163 +263,239 @@ it is recommended to leave any group chat before uninstalling Delta Chat.
...
+they might also use a non-Autocrypt e-mail app on that account
+which then leads to intermittently unencrypted communications
+because replying to an unencrypted message typically happens without E2E-encryption.
+This "reply-unencrypted" to "un-encrypted" messages is mandated by Autocrypt 
+to prevent unreadable messages on the side of your chat partners 
+and their non-Autocrypt compliant e-mail app. 
+
+If many of your chat partners use non-Autocrypt e-mail apps often (including web e-mail)
+you may toggle the Autocrypt setting in "Advanced settings"
+to rather default to "not prefering encryption"
```
link2xt: I would rather not suggest changing this setting at all. It can be used as a reason why the messages are unencrypted (because your contacts do not prefer encryption), but not as a suggestion. How do I even know if "many of my chat partners use non-Autocrypt e-mail apps often", they should know better than me what they use.

en/help.md
hpk42: integrated some info into PFS section.

en/help.md
```diff
@@ -613,20 +687,26 @@ like experiments? Register through "Sign up -> with Delta Chat"!)
...
-  continuously improved and many bugs have been fixed since the original
-  introduction in 2018. However, there remain cases, especially with large
-  groups where inconsistencies can occur, or messages become unreadable.
+Note that "1:1" chats are currently only opportunistically encrypted (Autocrypt). 
+You need to create a verified group with your contact in order to ensure
+that all messages will be safely E2E-encrypted between you two. 
+However, we plan to introduce verified 1:1 chats soon, 
+simplifying and extending the guarantees of verified E2E-encryption for all chat types. 
+Until then verified groups will remain classified as an experimental feature 
+although they are successfully used from many repression-facing groups,
```
```suggestion
although they are successfully used by many repression-facing groups, 
```

link2xt:
Could also add a question like "What encryption algorithms Delta Chat uses?", citing AES-128, SHA-256 and Ed25519/X25519, but this may become outdated after upcoming OpenPGP encryption upgrade, e.g. if SHA-256 is replaced with SHA-3.

en/help.md
```diff
@@ -263,163 +263,237 @@ it is recommended to leave any group chat before uninstalling Delta Chat.
...
+Nothing.
+
+Delta Chat apps (and other [Autocrypt](https://autocrypt.org)-compliant e-mail apps) 
+share the keys required for E2E-encryption encryption automatically.
+After a first message was received from a contact, 
+Delta Chat defaults to using E2E-encryption with that contact. 
+With [verified chat groups](#verifiedchats) 
+this initial encryption-handshake is automated and 
+all messages are guaranteed to be E2E-encrypted without any human interaction
+and even when servers are compromised.
```
missytake: Hmmm. Usually people don't ask me "what do I have to do to activate E2E", but "when will my messages be encrypted"... and I tend to reply with "after you exchanged messages", not with "Nothing".

I do see why we would want to frame the question like this, especially because E2E is just one small part of the security model, only protects against some threats, and gets too much focus anyway. The padlock ist already comparatively more focus than e.g. device encryption gets.

But if we think about "frequently asked questions", and want to answer the questions we receive... it feels more honest to me to say "to establish that you can just exchange messages".
But then you need to talk about group chats and gossip and suddenly you do a lot of explaining, which is somehow TMI.

You probably already considered all of this... I'm unsure myself on what to do here.

en/help.md
missytake: Maybe "groups under threat" would be a more broad term, but I also really like this one.

missytake (approved):

en/help.md
link2xt: The question was previously titled "What do I have to do to activate the end-to-end-encryption?" and started with "Nothing", so this part was just not properly revised. I would say it is worth rephrasing it.

hpk42: Am open to rephrasings , can review tomorrow morning.

hpk42: k, i rephrased and took up @missytake's suggestion. 
also revised the verified chats section a little 
see https://github.com/deltachat/deltachat-pages/pull/727/commits/ba7ebd29cf00b52b559e1bed23a3cbbcb6b402b5

missytake: Great!

PR #684, opened 2023-06-19
Encryption FAQ improvements, adapt to verified 1:1 chats

Hocuri:
Adapt FAQ for https://github.com/deltachat/deltachat-core-rust/issues/4188

I want to push this forward already now because when you tap on "Learn More", then the FAQ opens. But for now, the relevant FAQ section isn't written yet. When we settled for a rough version, holger had the idea to put it into some hidden place on the website, so that it can be linked from the Android app, but won't be translated yet since we may still want to change things.

PR #733, opened 2023-09-17
[superseded by #746, do not merge] Adapt FAQ to verified chats

Hocuri:
_Follow-up for #684. Draft since it's only to be merged when we release verified 1:1 chats._

_Feel free to edit this PR (I'm still behind with PRs/reviews, so I'll only start a discussion if I feel it's important :)). For big changes, it could make sense to quickly ask or make an a/v, but if you feel like it, you're free to directly make big changes into the PR here, too._

**FAQ Requirements for verified 1:1 chats are:**

- Talk about verified chats instead of groups - this made some changes necessary to feel consistent
- We need a #verifiedchats and a #verificationbroken section, which is shown if the user clicks on "Learn more" after the verification was established / broken.
- Both of them need to be understandable by non-technical users, especially the #verificationbroken section.
- The #verificationbroken section should give some advise how to repair the verification.

**What's also in this PR, without it being a requirement:**
- Break the instructions how to scan a QR Code into a step-by-step tutorial with bullet points
  - Include an image of the QR Code icon (in one of my user testings there was the problem that the user didn't recognize this icon as a QR Code, therefore failing to scan the QR Code. Not sure if that's a frequent problem, though, so we can also remove this change again).

**Other improvements in https://github.com/deltachat/deltachat-pages/pull/684/ that I didn't include in the PR here**, since I first wanted to check: Do y'all think that we should do this? (also, it's not necessary to be in the same PR)
- Starting with the easy questions. Right now, the first paragraph is "Which standards are used for end-to-end encryption?". I think that we should answer "What is end-to-end encryption?" etc somewhere close to the top.
  - Link to this "What is end-to-end encryption?" question from the "What is a verified chat?" question
- Use the wording "guaranteed end-to-end encryption" instead of "verified E2E" in one question, since that's what we're doing more in the UIs, too

**Questions:**
- We should probably always use "End-to-End" instead of "E2E", because the former is more widely known?

**To remember after merging (for me):**
- Let DC Android point to the new FAQ.
  - `whatdoesverifiedmean` was renamed to `verifiedchats`


adbenitez (approved):

en/help.md
```diff
@@ -433,6 +438,60 @@ Delta Chat shows two fingerprints there.
 If the same fingerprints appear on your own and your contact's device,
 the connection is safe.
 
+**In group chats:** {#createverifiedgroup}
```
r10s: these links do work only within headlines, see https://staging.delta.chat/733/en/help#howtoe2ee that downs `{#createverifiedgroup}` in plain text

en/help.md
```diff
@@ -433,6 +438,60 @@ Delta Chat shows two fingerprints there.
...
+     Most users don't need to worry about this, but if you are in a high-risk situation,
+     keep in mind that a powerful attacker with access to the email provider
+     could theoretically read and/or modify the messages.
+
+- **Why don't other messengers like Signal show such warnings?**
+
+  They do! The warnings just are a lot easier to overlook. Signal calls them
+  ["Your safety number changed"](https://support.signal.org/hc/en-us/articles/360007060632-What-is-a-safety-number-and-why-do-I-see-that-it-changed-).
+
+  Plus, Signal is not compatible with unencrypted emails, so, in Signal there is never the problem that your chat partner can't handle encryption anymore - all messages in Signal are encrypted.
```
r10s, line 494: lot's of "Signal" here :) 

unless we really know, these questions raise a lot for users, i do not think, we should compare in detail to Signal in our faq; including using their terms and link to pages explaining how to understand them. 

also, it may sound as if it is an advantage of signal to _not_ support email.

if one does not want to strike the paragraph alltogether, i  would shorten it to.

```suggestion
- **Why don't other messengers show such warnings?**

  They do! The warnings just are a lot easier to overlook and may read "Your safety number changed"
```

Hocuri: We can also completely remove this paragraph - while it would be a question I personally would have, we don't have evidence anyone else has it.

r10s: i think, saying that "it's not only us" but that it is a general problem, is fine

en/help.md
Hocuri: Hm. Let's try if deeper nesting works

hpk42:
closing this PR in favor of #746

PR #734, opened 2023-09-18
add a new entry about attachment encryption + two minor edits

hpk42:
lutindiscret asked about E2E-encryption of attachment on the fediverse. https://mastodon.libre-entreprise.com/@lutindiscret/111068632045634191 
also added two minor edits to keep the form of  FAQ entries to first say yes/no/... if it fits.

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -293,9 +293,17 @@ use [verified chat groups](#verifiedchats)
 which additionally protects against compromised or malfeasant e-mail servers. 
 
 
+### Are attachments (pictures, files, audio etc.) E2E-encrypted? 
+
+Yes. 
+When we talk about an "E2E-encrypted message" 
+we always mean a text message along with an (optional) attachment. 
+The whole message, including any attachment, is then E2E-encrypted. 
+
+
 ### Is OpenPGP secure? {#openpgp-secure}
 
-Delta Chat uses a secure subset of OpenPGP
+Yes, Delta Chat uses a secure subset of OpenPGP
 and only displays a padlock security indicator on a message
 if the whole message is properly encrypted and signed.
 For example, "Detached signatures" are not treated as secure.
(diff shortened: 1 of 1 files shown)


missytake (approved):
makes sense :)

en/help.md
```diff
@@ -293,9 +293,17 @@ use [verified chat groups](#verifiedchats)
 which additionally protects against compromised or malfeasant e-mail servers. 
 
 
+### Are attachments (pictures, files, audio etc.) E2E-encrypted? 
+
+Yes. 
+When we talk about an "E2E-encrypted message" 
+we always mean a text message along with an (optional) attachment. 
+The whole message, including any attachment, is then E2E-encrypted.
```
link2xt, line 301: What is missing is that attachments are not encrypted separately. Filenames, number of attachments etc. are not exposed.

en/help.md
```diff
@@ -351,6 +359,8 @@ as defined by the Autocrypt Level 1 specification.
 
 ### Is a message exposed in cleartext if E2E-encryption is not available? {#tls}
 
+No, probably your message is not exposed as cleartext to the network
```
link2xt, line 362: This "probably" summary sounds very bad for a security question.

For server-to-server encryption the situation is still not so good, Gmail has 90-95% of server-to-server traffic encrypted: https://transparencyreport.google.com/safer-email/overview

I think it is important to highlight that connection to the provider is always encrypted unless you explicitly configure otherwise, and messages are not exposed to your ISP. What your email provider does with the mail and whether it send and receive messages securely, is outside of our control.

en/help.md
hpk42: Not sure this clarification is neccessary but also can not hurt.

en/help.md
hpk42: apart from the last sentence i agree.  I'd leave the MTA-STS bit in even if it's not "our" responsibility.

PR #735, opened 2023-09-20
Follow-up for #734

link2xt:


en/help.md
```diff
@@ -359,12 +360,14 @@ as defined by the Autocrypt Level 1 specification.
...
-No, probably your message is not exposed as cleartext to the network 
-but it depends on the configuration of your e-mail provider. 
-With almost all mail servers, Delta Chat establishes strict _transport encryption_
-([TLS](https://en.wikipedia.org/wiki/Transport_Layer_Security))
-which secures connections between your device and your e-mail server. 
+No,
+this does not necessarily mean that the message is exposed in cleartext.
+
+Delta Chat always uses ([TLS](https://en.wikipedia.org/wiki/Transport_Layer_Security)) encryption
+which secures connections between your device and your e-mail server
```
r10s: we're most of the time saying "e-mail provider" when it comes to who's responsible for the user's email address. not sure which fits better, but i think, the term "provider" is more known in this context. morevoer, later on, we also say "your e-mail provider"

```suggestion
Delta Chat always uses ([TLS](https://en.wikipedia.org/wiki/Transport_Layer_Security)) encryption
which secures connections between your device and your e-mail provider
```

nb: is "always" correct here? so if a provider does not support TLS, you can only log in until manually disabled? if in doubt, just strike the "always".

link2xt: Yes, if you want no TLS at all, you have to explicitly select Plaintext. We try various ports, TLS/STARTTLS and strict/opportunistic automatically, but never try to use Plaintext unless it is explicitly chosen.

r10s: great, thanks for clarification!

hpk42 (approved):

PR #736, opened 2023-09-21
Provide webp version of thumbnails for the front page

link2xt:
They are more than twice as small compared to PNG versions and should be supported by all browsers nowadays,
but there is still a fallback.


en/index.md
```diff
@@ -14,23 +14,39 @@ lang: en
 # Available on mobile and desktop
 
 
+<picture>
+<source srcset="../assets/blog/screenshots/2019-12-17-delta-chat-google-play-release-chat-list-light-thumbnail.webp" type="image/webp" />
+<source srcset="../assets/blog/screenshots/2019-12-17-delta-chat-google-play-release-chat-list-light-thumbnail.png" type="image/png" />
 <a href="../assets/blog/screenshots/2019-12-17-delta-chat-google-play-release-chat-list-light.png">
 <img src="../assets/blog/screenshots/2019-12-17-delta-chat-google-play-release-chat-list-light-thumbnail.png" width="120" height="213"
 style="float: left; margin: 10px;display: block;box-shadow: 5px 5px 2px #777;" alt="A screenshot of Delta Chat on Android showing chat list" />
 </a>
```
r10s, line 23: i have seen your first, more logical but not working, try with  `<a>...</a>` around `<picture>`. really wondering why the variation with  `<a>...</a>`  inside `<picture>` is working :)

but tested on safari, chrome, firefox - things are displayed correctly now

however, chrome and safari seems not to pick up the webp variation for me (firefox does, checked in the "network" page of the dev tools)

probably still good enough

r10s (approved):
thanks a lot for taking care of details

link2xt:
No idea why it does not work for chromium, but at least it is an improvement for firefox.

PR #737, opened 2023-09-21
Fix loading of webp images

link2xt:
Follow-up for #736

Had to move `<a>` tag out of `<picture>` and wrap everything into `<div>`. `<div>` is needed because kramdown parser does not start an HTML block when `<a>` is encountered as `<a>` tag is a span-level HTML tag: https://kramdown.gettalong.org/syntax.html#html-blocks

### DIFF
--- a/en/index.md
+++ b/en/index.md
@@ -14,39 +14,45 @@ lang: en
 # Available on mobile and desktop
 
 
+<div>
+<a href="../assets/blog/screenshots/2019-12-17-delta-chat-google-play-release-chat-list-light.png">
 <picture>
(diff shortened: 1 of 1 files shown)


r10s:
is kramdown used by Jekyll? or why is it important?

EDIT: answering myself: probably because of the previous issues where `<a>` was just written out and not recognised as a tag - and that issue is probably similar in Jekyll and in kramdown

r10s (approved):
lgtm - and also fits more into the common understanding of `<a> sth on screen </a>`

link2xt:
> is kramdown used by Jekyll?

Yes, it is a dependency in `Gemfile.lock`.

PR #738, opened 2023-09-22
Compress download badges to WebP

link2xt:
This reduces total size of images from 164K to 104K.

Files were converted with

  `for x in *.png; do cwebp -lossless -q 100 -z 9 $x -o ${x%.png}.webp; done`


r10s:
the gplay badge is a bit out of date anyways - and the current version is 5k only

google-play-badge

(via https://play.google.com/intl/en_us/badges/ )

i think, in this case, we should better go for that.

link2xt:
I updated the badge, had to trim the version Google provides with `convert input.png -trim output.png`, then recompressed PNG version which got large after imagemagick with `optipng -Z` and converted to webp again.

Google Play beta badge is removed instead of updating the logo because we are not using it anyway.

link2xt:
Hmm, there is something wrong with the layout, it has shifted if you compare https://delta.chat/en/download and https://staging.delta.chat/738/en/download by switching the tab.

Somehow picture tag around the badge is 56 px in height, while the image is 48 px.

Found an answer: https://stackoverflow.com/questions/6664460/line-height-affects-images
`img` height is affected by the `line-height`, we should set `display: block` on images, at least the ones inside `a.img-badge`.

EDIT: fixed it, now the layout is exactly as before.

link2xt:
Also updated apklis PNG from 20K to 3.7K, webp also got 14.7 KB smaller.

link2xt:
Overall the whole page is reduced from 154K to 65K now.

_sass/_downloadpage.scss
```diff
@@ -50,8 +50,9 @@ div.download-content {
         display: flex;
         align-content: center;
         margin: 3px;
-        &>img{
+        img{
             height: 48px;
+            display: block;
```
r10s, line 55: nb: another lever to reduce the bytesize would have been to make use of  this height - so, scaling all non-vector images to 2x48px ("2x" to look better on retina/hi-res). that would easily half the number of pixels that need compressions (current images around 150px-250px in height

but this is no blocker for this pr, just an idea to optimize further

r10s (approved):
nice! thanks a lot for saving 90k on each opening of the download page!

_sass/_downloadpage.scss
link2xt: Probably the correct way is to have different versions of images for different DPI, but I do not plan to improve this further at the moment.

PR #740, opened 2023-10-04
wording: renamed private apps to webxdc apps again

### LINKED ISSUE #712: wording in the FAQ: private apps -> webxdc apps
We decided to change the wording.

In hardly any current chat or current writings do we talk about webxdc apps as  "private apps". We also had private apps in the media tabs of delta chat and now it's just apps.

web apps, chat-shared apps, webxdc apps or just apps are all fine but "Private App" as if it would be a standing central term we don't think it's a good idea anymore.

missytake:
closes #712


hpk42 (approved):
thanks :)

PR #741, opened 2023-10-05
cleaner changelog

r10s:
by the "sidebar" github added recently,
there is even more cluttering on the screen.

the changelogs are meant for end-users,
so better keep things simple.
also, one of the first buttons on github is "Pricing", which may be confusing := on mobile, you see barely a single change before scrolling.

before/after:

Screenshot 2023-10-05 at 20 50 13 Screenshot 2023-10-05 at 21 25 57


adbenitez:
makes sense to remove the core changelogs link, it is not interesting for end users but for developers using core in their 3rd party product/bot

adbenitez (approved):

PR #747, opened 2023-11-06
hide windows store button because it is broken

Simon-Laux:
See https://github.com/deltachat/deltachat-desktop/issues/3292 for details.


missytake (approved):
Nice, easy to revert later.

PR #746, opened 2023-11-04
revise FAQ to reflect 1.42 release terminology 

hpk42:
based on https://github.com/deltachat/deltachat-pages/pull/733

PR #748, opened 2023-11-08
ansuz/faq e2e amendments

ansuz:
I don't intend for this patch to be merged as-is. some of them are corrections to the FAQ itself, but others are concerned with policies in various delta chat implementations that are scattered across other repos. it was simply easier to annotate this text with suggestions rather than find all the appropriate places to file issues.

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -420,21 +420,25 @@ This will automatically resolve the warning for yourself
 and for all members of the green-checkmarked chat. 
 
 **Your contact sent a mail through a webmail interface or another e-mail app
-and will get back to use Delta Chat soon again.**
+and will get back to using Delta Chat soon again.**
 
 Simply accept the warning.
 As soon as your contact uses Delta Chat again, 
 guaranteed end-to-end encryption will be automatically re-established. 
+<!-- two things here:
+1. I'd avoid using "simply" in this paragraph for the reasons outlined in https://justsimply.dev/
+2. in general this is a dangerous habit to encourage (dismiss the possibly important security warning). I don't have a better suggestion in mind as far as the FAQ is concerned, but alert fatigue can be disastrous and it might be worth rethinking the current approach
+-->
 
-**Your contact stopped using Delta Chat at all**
+**Your contact stopped using Delta Chat entirely**
 
 Accept the warning and remove the contact from any active green-checkmarked group 
-(which you can find in "Shared chats" in the Contact profile). 
+(which you can find in "Shared chats" in the Contact profile). <!-- it would be interesting to make this possible from a single button in their Contact profile -->
 
-Don't fret too much about the "May not be end-to-end encrypted anymore" warning.
+Don't fret too much about the "May not be end-to-end encrypted anymore" warning. <!-- in principle I agree that there are situations where I'd rather fall back to regular mail rather than insist on maintaining encryption and losing contact, but I'd probably rephrase this paragraph to exclude "Don't fret too much" which could be seen as dismissive to people who really do require e2e -->
 Sometimes remaining in contact is more important than end-to-end encryption.
 Chances are that ["Transport Layer Encryption" (TLS)](#tls) still meaningfully protects 
-the confidentiality of your messages between your device and the e-mail server. 
+the confidentiality of your messages between your device and the e-mail server. <!-- I think I'd feel better about this conclusion if it said "_may_ still meaningfully protect". ideally we don't want anyone to panic, but it is worth reminding people to reflect on their use-case and threat model (without necessarily using that jargon) -->
 
 
 ### Are attachments (pictures, files, audio etc.) end-to-end encrypted?
@@ -840,7 +844,7 @@ extendable messenger.
 
 ## Experimental Features
 
-We are very grateful about feedback on these features - do you want to share <!-- grateful _for_ feedback -->
+We are very grateful for feedback on these features - do you want to share 
 your ideas? Join the [Forum](https://support.delta.chat) to contribute. (You
 like experiments? Register through "Sign up -> with Delta Chat"!)
 
--- a/en/help.md
+++ b/en/help.md
@@ -284,6 +284,8 @@ guarantee end-to-end encrypted messages.
 
 ### When will messages be end-to-end encrypted? {#whene2e}
 
+<!-- this almost sounds like a roadmap question. it might be better to phrase it as "how can I tell if..." -->
+
 All end-to-end encrypted messages carry a padlock
 <img style="vertical-align:middle; width:1.5em; margin:1px" src="../assets/home/padlock.png" alt="padlock"/>.
 
@@ -306,6 +308,8 @@ automatically spreading guaranteed end-to-end encryption among all chat members.
 
 
 ### How can I get guaranteed end-to-end encryption and green checkmarks? {#howtoe2ee}
+
+<!-- perhaps worthwhile to recommend that the secondary channel be e2e-encrypted, if possible -->
 
 Meet your chat partner outside Delta Chat, preferably in person
 but a second channel like a video chat
@@ -371,6 +375,8 @@ Whoever adds a contact to a green-checkmarked group becomes an introducer
 to those members who didn't yet know about the added contact.
 In a Contact profile you can tap on the "Introduced by ..." text repeatedly
 until you get to the one with whom you directly did a [QR-scan](#howtoe2ee).
+
+<!-- a diagram depicting the web of trust could probably help people process the introducer chain. I'm primarily thinking of an example in the faq to demonstrate the generic concept of a web of trust, but perhaps in the future this could be an interactive feature of delta chat accessible via the contact profile UI? -->
 
 Note that in a Contact profile you may see and tap introducers 
 but there is no green checkmark in the profile title. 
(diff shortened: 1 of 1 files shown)


ansuz:
oops. somehow I ended up with merge conflicts with the upstream force-push :shrug:

I can revisit this later and resolve if necessary

en/help.md
```diff
@@ -284,6 +284,8 @@ guarantee end-to-end encrypted messages.
 
 ### When will messages be end-to-end encrypted? {#whene2e}
 
+<!-- this almost sounds like a roadmap question. it might be better to phrase it as "how can I tell if..." -->
+
```
hpk42, line 288: changed

en/help.md
```diff
@@ -307,6 +309,8 @@ automatically spreading guaranteed end-to-end encryption among all chat members.
 
 ### How can I get guaranteed end-to-end encryption and green checkmarks? {#howtoe2ee}
 
+<!-- perhaps worthwhile to recommend that the secondary channel be e2e-encrypted, if possible -->
+
```
hpk42, line 313: i don't think it's neccessary -- in general everything you do / collab with other people should be e2e-encrypted ;)
but here, having a separate second channel, even if non-e2ee jitsi or telegram, raises the costs for the attacker considerably -- the attacker needs to a) control the mail server b) control the random jitsi instance + have some AI running to replace QR codes seemlessly -- not impossible but a very sophisticated costly attack and it's hard to guess what second channel people will use -- the "preferably in person" already indicates that anything else is less safe.

en/help.md
```diff
@@ -372,6 +376,8 @@ to those members who didn't yet know about the added contact.
 In a Contact profile you can tap on the "Introduced by ..." text repeatedly
 until you get to the one with whom you directly did a [QR-scan](#howtoe2ee).
 
+<!-- a diagram depicting the web of trust could probably help people process the introducer chain. I'm primarily thinking of an example in the faq to demonstrate the generic concept of a web of trust, but perhaps in the future this could be an interactive feature of delta chat accessible via the contact profile UI? -->
```
hpk42, line 379: yes, that would be nice but out of scope for the current iteration of FAQ and 1.42 release.

en/help.md
```diff
@@ -414,21 +420,25 @@ This will automatically resolve the warning for yourself
 and for all members of the green-checkmarked chat. 
 
 **Your contact sent a mail through a webmail interface or another e-mail app
-and will get back to use Delta Chat soon again.**
+and will get back to using Delta Chat soon again.**
```
hpk42, line 423: applied.

en/help.md
```diff
@@ -414,21 +420,25 @@ This will automatically resolve the warning for yourself
...
 
 Simply accept the warning.
 As soon as your contact uses Delta Chat again, 
 guaranteed end-to-end encryption will be automatically re-established. 
+<!-- two things here:
+1. I'd avoid using "simply" in this paragraph for the reasons outlined in https://justsimply.dev/
+2. in general this is a dangerous habit to encourage (dismiss the possibly important security warning). I don't have a better suggestion in mind as far as the FAQ is concerned, but alert fatigue can be disastrous and it might be worth rethinking the current approach
+-->
```
hpk42, line 431: thanks for the valid point! i revised the whole FAQ entry about "sent a message from another device" to reflect your valid criticism.  In fact, the original text was clearly missing the case of "you are under attack", as unlikely as it may be.   The entry now starts like this (but also i modified some other bits according to your comments) 

### A contact "sent a message from another device", what can i do? {#nocryptanymore}

Your chat with a contact lost guaranteed end-to-end encryption. 
The green checkmark was removed for this chat and contact when you see this warning. 
**If you find the sudden drop of guaranteed end-to-end encryption 
surprising for this contact then don't accept the warning!**
Instead check with your contact through a second channel 
like a video call, other messenger or a phone call,
to find out what happened. 

If your contact actually caused the drop of guaranteed end-to-end encryption
please see the next paragraphs for common reasons and their mitigations.
Regardless, all other green-checkmarked chats remain guaranteed end-to-end encrypted
even if the contact is a member there.  
[...]

en/help.md
```diff
@@ -414,21 +420,25 @@ This will automatically resolve the warning for yourself
...
 
-**Your contact stopped using Delta Chat at all**
+**Your contact stopped using Delta Chat entirely**
 
 Accept the warning and remove the contact from any active green-checkmarked group 
-(which you can find in "Shared chats" in the Contact profile). 
+(which you can find in "Shared chats" in the Contact profile). <!-- it would be interesting to make this possible from a single button in their Contact profile -->
```
hpk42, line 436: sure, but not in 1.42 ;)

en/help.md
```diff
@@ -414,21 +420,25 @@ This will automatically resolve the warning for yourself
...
 
-Don't fret too much about the "May not be end-to-end encrypted anymore" warning.
+Don't fret too much about the "May not be end-to-end encrypted anymore" warning. <!-- in principle I agree that there are situations where I'd rather fall back to regular mail rather than insist on maintaining encryption and losing contact, but I'd probably rephrase this paragraph to exclude "Don't fret too much" which could be seen as dismissive to people who really do require e2e -->
 Sometimes remaining in contact is more important than end-to-end encryption.
 Chances are that ["Transport Layer Encryption" (TLS)](#tls) still meaningfully protects 
-the confidentiality of your messages between your device and the e-mail server. 
+the confidentiality of your messages between your device and the e-mail server. <!-- I think I'd feel better about this conclusion if it said "_may_ still meaningfully protect". ideally we don't want anyone to panic, but it is worth reminding people to reflect on their use-case and threat model (without necessarily using that jargon) -->
```
hpk42, line 442: thanks, godo points.  I revised the whole sub-entry: 

**Your contact stopped using Delta Chat entirely**

Sometimes remaining in contact is more important than end-to-end encryption.
Chances are that ["Transport Layer Encryption" (TLS)](#tls) still meaningfully protects
the confidentiality of your messages between your device and the e-mail server. 
However, if you have reasons to distrust your or the contact's e-mail server
then it's better to not use this chat anymore for sensitive information. 

There is not much you can do for now than
accepting the warning and removing the contact from any active green-checkmarked group
(which you can find in "Shared chats" in the Contact profile). 
If the contact removed the Delta Chat app because of buggy or undesirable behaviour,
please consider posting a note to our [support forum](https://support.delta.chat)
to help us identify wider problems that need addressing. Thanks!

en/help.md
```diff
@@ -499,6 +509,10 @@ as defined by the Autocrypt Level 1 specification.
 
 ### Is a message exposed in cleartext if end-to-end encryption is not available? {#tls}
 
+<!-- this is another place where I think it would be really helpful to have a good diagram to indicate what parties could be involved in sending an email: (you, the recipient(s), your email provider, that of all a message's recipients, and then various other infrastructure providers (telecom, wifi access points, etc.).
```
hpk42, line 512: yes

en/help.md
```diff
@@ -499,6 +509,10 @@ as defined by the Autocrypt Level 1 specification.
...
+
+As it stands this section requires a considerable amount of external reading to make sense of whether the reader can safely proceed. "No, not necessarily" and "will today typically be transport-encrypted as well" are generally correct, but are premised on assumptions that most readers won't be capable of verifying. In particular I think most readers will not understand the concrete implications of being exposed to "the network" or what it means for an e-mail server to be compromised (hackers? nation states? ad-tech? insider threats?). -->
+
```
hpk42, line 515: good point.  I modified the faq entry to start differently:  

### Is a message exposed in cleartext if end-to-end encryption is not available? {#tls}
Even if your messages are not guaranteed to be end-to-end encrypted, 
they are still protected from your Internet providers like cell or cable companies.  
However, your and your contact's e-mail providers
may read, analyze or even modify your messages, 
including any attachments,
if they are not end-to-end encrypted. (...)

I also modified the whole entry a little to accomodate your comment and valid criticism.

hpk42 (commented):
thanks a lot, ansuz!
i mostly applied/addressed your comments in https://github.com/deltachat/deltachat-pages/pull/746/commits/d355e4b20065ce7a26c8b4b1fcb6cf00976cfb20
will be closing this PR here as i think all that can be addressed is addressed now. 
Would be grateful if you can check the commit/diff to see if i missed something, or addressed it poorly.

en/help.md
ansuz: I commented on two minor typos, otherwise it looks good!

en/help.md
ansuz: I've made a note to find or put together a diagram for a later release for this and the e2e network topology.

en/help.md
ansuz: I wasn't sure where to file an issue since this would presumably need to be implemented across multiple platforms. would a forum post be most appropriate?

hpk42:
On Thu, Nov 09, 2023 at 02:21 -0800, ansuz wrote:

> @ansuz commented on this pull request.
>  Accept the warning and remove the contact from any active green-checkmarked group 
> -(which you can find in "Shared chats" in the Contact profile). 
> +(which you can find in "Shared chats" in the Contact profile). 
> 
> I wasn't sure where to file an issue since this would presumably need to be implemented across multiple platforms. would a forum post be most appropriate?

Not sure an issue makes much sense. There are many things we can improve
and it's unclear how much this is really an issue in the real-world. 
It's not an everyday occurence that people suddenly stop using DC
when they also have many chats that are still active, 
and then re-use the same e-mail address. 
More typically, they only have very few contacts (which is the reason they drop it),
and then it's not a big issue at all.

PR #755, opened 2023-11-16
make sure, deep links in the help are working

r10s:
we're using some deep link anchors meanwhile in the apps, make sure, these links are always working -
and not get lost on updates or translations.

### DIFF
--- a/tools/create-local-help.py
+++ b/tools/create-local-help.py
@@ -25,6 +25,14 @@ linked_files = [
     "assets/home/delta-what-optim.png"
 ]
 
+# list all anchors (fragments) that are used from outside to access the help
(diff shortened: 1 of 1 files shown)


adbenitez (approved):
I was just wondering that!!! and was manually testing if they are the same for the Spanish page of the FAQ

r10s:
> I was just wondering that!!! and was manually testing if they are the same for the Spanish page of the FAQ

:)

in fact, the script also checks all internal links (they are partly even translated)

and also pings all external links, checking that this is 200 OK or so

but still good to have a manual look on that, i also do that for german (and a bit randomly and roughly on other languages)

Simon-Laux (approved):

PR #758, opened 2023-11-17
replace all links to countermitm with securejoin urls and names

hpk42:
i only did the english version, not sure if it makes sense to also do all other languages in this PR? (i ran a script to produce the PR, so easy to adapt).


r10s:
it does not make sense to change other languages here; they are overwritten on next pull from transifex.

to mitigate burden to translators i can change lnks there (if not yet done meanwhile by translators)

tl;dr it is fine to only change english

r10s (approved):

PR #765, opened 2023-11-23
FAQ item to explain disappearing messages

### LINKED ISSUE #762: Clarify how disappearing messages work
In https://delta.chat/en/help#how-can-i-ensure-message-end-to-end-encryption-and-deletion, we should add a small explanation on how the messages are deleted. Maybe an FAQ item of its own does make sense.

missytake:
It doesn't entirely fit into the "encryption" section, but as it relates to the FAQ item below it, I thought it's probably the best spot. Alternatively, "miscellaneous" would also work.

fix #762

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -183,7 +183,7 @@ to mute a chat, use the chat's menu (Android/Desktop) or the chat's profile (iOS
   message them or write to a group they're in as well.
 
 
-### How do disappearing messages work?
+### How do disappearing messages work? {#ephemeralmsgs}
 
 You can turn on "disappearing messages"
 in the settings of a chat,
@@ -642,7 +642,7 @@ is [using chats with guaranteed end-to-end encryption](#howtoe2ee)
 and turning on disappearing messages.
 
 Guranteed end-to-end encrypted chats protect against [MITM attacks](https://en.wikipedia.org/wiki/Man-in-the-middle_attack)
-and turning on ["disappearing messages"](#how-do-disappearing-messages-work) deletes the messages
+and turning on ["disappearing messages"](#ephemeralmsgs) deletes the messages
 on the server after a user-configured time.
 
 If you don't need a longer-lived copy of your messages on the server, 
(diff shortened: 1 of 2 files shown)


en/help.md
```diff
@@ -603,6 +603,37 @@ even if the direct chat between you two has a
...
+both in each email account on the server,
+and in the app itself.
+
+Note that you can rely on disappearing messages
+only as long as you trust your chat partners;
+malicious chat partners can screenshot
+or otherwise save a message before it is deleted.
+
+Apart from that,
+if one chat partner silently uninstalls Delta Chat,
```
```suggestion
if one chat partner uninstalls Delta Chat,
```

i think, "silently" does not really matter here

r10s (commented):
lgtm, i mean we had zero information before :)

but i would indeed move that out of encryption, it does not really fit and breaks the flow when reading the whole encryption things "beginning to end"

"miscellaneous" seems to be okay - or the first section where where also explain pinning, muting, archiving, online-dot etc

en/help.md
```diff
@@ -183,6 +183,37 @@ to mute a chat, use the chat's menu (Android/Desktop) or the chat's profile (iOS
...
+The time span begins when the receiver's Delta Chat app
+first sees the message.
+The messages are deleted
+both in each email account on the server,
+and in the app itself.
+
+Note that you can rely on disappearing messages
+only as long as you trust your chat partners;
+malicious chat partners can screenshot
+or otherwise save a message before it is deleted.
```
```suggestion
malicious chat partners can take photos,
or otherwise save, copy or forward messages before deletion.
```

i'd avoid the term "screenshot" here as this immediately results in the statement "why don't you disable screenshots as other apps?" - and we should not dive deeper in this help :)

moreover, i'd pick up at least the terms we have in the UI.

r10s (approved):
i added a commit that introduces a static anchor; this helps adding a "learn more" button to the UIs at some point.

moreover, i added a minor wording suggestions, so, all in all lgtm now :)

missytake:
I added one more commit because I realized that "if DC gets uninstalled it can't delete messages from the device anymore" doesn't make sense, we can probably assume the data is deleted anyway and the messages don't arrive anymore.

PR #768, opened 2023-11-24
clarify when 'ephemeral timer' really starts

r10s:
the passive wording
'begins when the receiver’s Delta Chat app first sees the message' is quite passive and could be interpreted
such as there is no user action needed at all to start the timer.

in fact, the timer starts when the _user_ first "sees" the message.

/me came over that during translating


missytake (approved):
That's a much nicer wording.

PR #772, opened 2023-11-24
Update PGP import documentation to match behavior

snan:
Alternatively, someone could fix this behavior in the app so it could
autodetect keys regardless of file endings. But then it'd have to
scan every file.

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -711,10 +711,6 @@ This is not the main goal of Delta Chat. In fact, the majority of new
 users will not have any key prior to using Delta Chat. We do, however,
 try to support private keys from as many sources as possible.
 
-If you want to keep your passphrase you'll have to create an e-mail
-alias for use with Delta Chat such that Delta Chat's key is tied to
-this e-mail alias.
-
 Removing the password from the private key will depend on the
 software you use to manage your PGP keys. With Enigmail, you can set your
 password to an empty value in the Key Management window. With GnuPG you can set
--- a/en/help.md
+++ b/en/help.md
@@ -704,8 +704,8 @@ passphrase encryption and the password and try the import again.  If you want
 to keep your passphrase you'll have to create an e-mail alias for use
 with Delta Chat such that Delta Chat's key is tied to this e-mail alias.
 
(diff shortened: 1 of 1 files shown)


en/help.md
```diff
@@ -704,6 +704,9 @@ passphrase encryption and the password and try the import again.  If you want
 to keep your passphrase you'll have to create an e-mail alias for use
 with Delta Chat such that Delta Chat's key is tied to this e-mail alias.
 
+Another common error is if your file is in the ascii armored format,
+the file must have an `.asc` file ending.
```
```suggestion
Another common error: if your file is in the ascii armored format,
the file must have an `.asc` file ending.
If you struggle with getting Delta Chat to recognize your keys,
a good workaround is to export the PGP keys Delta Chat generated when you logged in,
and manually edit the exported PGP key .asc files so they contain the key you want to use.
If the keys are in the same format in which Delta Chat just exported them,
it will be the easiest.
```

missytake (approved):
Good start. Consider adding this suggestion below - I'm not entirely sure with the wording though, as it's already quite late.

en/help.md
r10s: @missytake manually editing key files is getting really special here, this seems not to be the main target group for the faq in its current form, not to speak about support issues with that. adding tricks and tweaks for pro-users to the help seems to be a non-goal.

better shift that to a [forum HOWTO-entry](https://support.delta.chat/c/howto/8) and add a link to the help.

that said, i find the short version of @snan better fitting to overall level of depth

snan:
There's no guarantee that editing an exported file will get it right; that might lead to a mismatch of unarmored parts in an armored file or vice versa.

snan:
But recommending to use the armored format is probably a good idea, @missytake.

Those are the only ones I've managed to get it working with.

missytake (approved):

r10s:
thanks, very nice now!

i just wanted to comment wrt linebreaks, however, as this was already merged, i've done a follow-up at https://github.com/deltachat/deltachat-pages/pull/773

PR #774, opened 2023-11-25
the minimum for 'disappearing messages' seems to be 1 minute

r10s:
i checked recent android/desktop/ios - all "1 minute" as minimum.

@missytake where do the "30 seconds" come from?

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -189,7 +189,7 @@ You can turn on "disappearing messages"
 in the settings of a chat,
 at the top right of the chat window,
 by selecting a time span
-between 30 seconds and 5 weeks.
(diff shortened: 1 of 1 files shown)


r10s:
k, i researched a bit, until 2021, we had a "30 seconds" option, it is changed after user testing at https://github.com/deltachat/deltachat-android/pull/1780 - the change to "5 weeks" came a bit later at https://github.com/deltachat/deltachat-android/pull/1786

in fact, android and desktop have these options while ios is behind here, i filed an issue for the latter at https://github.com/deltachat/deltachat-ios/issues/2000

merging that in to avoid more wrong translations :)

PR #784, opened 2023-12-27
add page for 37c3

Simon-Laux:


Simon-Laux:
preview at: https://staging.delta.chat/784/37c3

missytake (approved):
Good for now, but why not write it in markdown and make use of our css? I first thought that you'd copy-pasted parts of en/index.md but didn't include the css properly... but it seems to be taken from the generated index.html?

PR #785, opened 2024-01-02
update 1.42 blogpost: 1.42 is out meanwhile

Simon-Laux:


r10s (approved):
i'd also be fine to just delete the corresponding lines. but strike-through is fine as well, most important is the info that waiting is no longer required, thanks for fixing!

PR #787, opened 2024-01-07
security txt append year

Simon-Laux:
extend expiry date for another year


r10s (approved):
thanks for taking care!

ftr, this pr seems to refer to the docs at https://www.rfc-editor.org/rfc/rfc9116#section-2.5.5 , saying "It is RECOMMENDED that the value of this field be less than a year into the future to avoid staleness"

PR #789, opened 2024-01-12
Add nine.testrun.org QR invite to download page

missytake:
In other languages than english it doesn't work yet; the new strings on the download page remain empty, I don't understand why. Any ideas? I tried to do transifex magic, but tx init fails for me because api.transifex.com isn't reachable o.0

This could be prettier, but good enough for now. I'm not a designer.

### DIFF
--- a/_data/lang/ca.yml
+++ b/_data/lang/ca.yml
@@ -22,8 +22,12 @@ ca:
   language: Idioma
 
   download:
-    recommendation: Recomanat per a tu
-    other: Altres plataformes
+    install: ""
+    get_account: ""
+    get_account_explanation: ""
+    add_friends: ""
+    add_friends_explanation: ""
+    other: ""
     source_code: Codi font
     install_via_packagemanager: Instal·la'l via Gestor de Paquets
     outdated: desactualitzat
--- a/_data/lang/cs.yml
+++ b/_data/lang/cs.yml
@@ -22,8 +22,12 @@ cs:
   language: Jazyk
 
   download:
-    recommendation: Doporučujeme
-    other: Jíné technologie
+    install: ""
+    get_account: ""
+    get_account_explanation: ""
+    add_friends: ""
+    add_friends_explanation: ""
+    other: ""
     source_code: Zdrojový kód
     install_via_packagemanager: Instalace ze správce balíčků
     outdated: zastaralé
--- a/_data/lang/de.yml
+++ b/_data/lang/de.yml
@@ -22,8 +22,12 @@ de:
   language: Sprache
 
   download:
-    recommendation: Für Sie empfohlen
-    other: Andere Plattformen
+    install: Delta Chat installieren
+    get_account: Ein Konto einrichten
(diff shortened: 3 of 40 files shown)


hpk42 (approved):
Very nice improvement already.  (But language needs to be fixed)

missytake:
We could also make the `<details>` tag nicer, I didn't try that so far: https://www.sitepoint.com/style-html-details-element/

r10s:
first of all, yeah, this is a huge improvement; i just did not came to fix translations so far, but if will care for it asap.

for the new "Download" link in the menu - i am not sure about it. one the one hand it was emphasised by border and bold - on the other hand de-emphasised my moving position. all in all, this results to quite some noise with few effect.  
also before, as the second item, "Download" was quite visible.

maybe this part was good enough before - or we add an _additional_ get started badge in the upper right corner, sth. rotated by 45 degrees:

Screenshot 2024-01-16 at 14 42 15

but maybe move that out of this pr, it seems to be an independent improvement

missytake:
> for the new "Download" link in the menu - i am not sure about it. one the one hand it is emphasised by border and bold - on the other hand de-emphasised by moving position. all in all, this results to quite some noise with few effect. also before, as the second item, "Download" was quite visible.

I would say the new position *emphasizes* it, actually.

(I'm not sure if I understood your comment correctly, I fixed some typos in the quote to indicate how I understood it)

r10s:
it is overall stronger by the bold button style, but being the last element in the menu weakens it a little, imo. but maybe it is also that i find the big button a bit misplaced in the menu, it does not fit well optically, it is even lager than the icon. also, not sure, there are much issues with the current button (therefore, the "Get Started" idea, in case, ppl do not understand the wording "Download")

however, another idea: maybe we can make it the _first_ element by striking the text "Home"? ppl can click the logo (this is probably common knowledge) or use the "back button" - and if really ppl do not find the way back to the home page, and instead go for "Download", then it's that. maybe we can chat tomorrow a bit over that and find a solution :)

r10s:
tbh, i think we should split the menu bar discussions of this pr - it is unclear, and the current approach breaks with some known UI guidelines  (some mentioned above, plus, after some one-to-one discussions: an action button in the menu bar. it is a menu item, but it does not look like a menu item. tapping first "download" on download page does not work. being the item in the menu bar is easily confused with "this is how selected pages look like)

so, i do not think, the change on the button  is an improvement. if we think, or even have reports, that the "Download" menu item is not bold enough or needs to be renamed, we should do that in another pr

missytake:
> tbh, i think we should split the menu bar discussions of this pr - it is unclear, and the current approach breaks with some known UI guidelines (some mentioned above, plus, after some one-to-one discussions: an action button in the menu bar. it is a menu item, but it does not look like a menu item. tapping first "download" on download page does not work. being the item in the menu bar is easily confused with "this is how selected pages look like)

Point, didn't think of that. I reverted the commit.

missytake:
Hm, we could hide it as long as we're on the download page. But if, then maybe in another PR.

r10s:
> But language needs to be fixed

fixed (a `./tools/t-dance push; ./tools/t-dance pull` was missing)

r10s (approved):

PR #791, opened 2024-01-17
fix non-js download

r10s:
fixes #789 in case javascript is disabled


missytake (approved):
In lynx, I still see `boxes summary`, but lynx users are used to a lot of weird stuff :shrug:

r10s:
i added a commit that fixes the cursor for the expand link, and yes, lynx is probably still good enough :)

PR #793, opened 2024-01-18
FAQ: explain setting e-mail subject and signature

### LINKED ISSUE #790: FAQ item: how to add subject line
Sure, for now creating a group with 2 members and setting the group name as subject is just a workaround, but it's nice enough that it makes sense to add it to the FAQ.

missytake:
fix #790 

I got a bit carried away and also wrote about signatures. For users who want to know more about using Delta Chat as a normal e-mail client this is valuable information, and it helps shaping expectations about how Delta Chat deals with the balancing act of being both a messenger and an e-mail client.

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -96,6 +96,38 @@ You may also wish to learn [how to add accounts to multiple devices](#multiclien
   device.
 
 
+### Can I set the E-Mail Subject with Delta Chat?
+
+For most messages, Delta Chat sets the subject to
+"New Message from Alice" automatically,
+or "..." if it is encrypted.
+The exception are group chats,
+where the subject is the Chat Title.
+
+So if you want to set the subject yourself,
+for example for a formal e-mail to a business contact,
+you can create a group
+with only yourself and the recipient(s),
+set the Chat Title to the Subject you want,
+and then send your e-mail as the first message.
+If the recipients don't use Delta Chat,
+it will look like a normal,
+formal e-mail to them.
+You will get bonus professionalism points
+if you set a [signature](#signature).
+
+
+### Can I set an E-Mail Signature with Delta Chat? {#signature}
+
+Yes,
+you can do so under "Settings > Edit Profile > Signature Text".
+Your contacts who use Delta Chat will see it
+when they view your contact details.
+To everyone else,
+It will appear as an e-mail signature
+below the text of all your messages.
+
+
 ### Does Delta Chat support HTML e-mails?
 
 Yes, incoming HTML messages come with a "Show full message" button. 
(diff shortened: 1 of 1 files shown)


en/help.md
```diff
@@ -96,6 +96,38 @@ You may also wish to learn [how to add accounts to multiple devices](#multiclien
...
+it will look like a normal,
+formal e-mail to them.
+You will get bonus professionalism points
+if you set a [signature](#signature).
+
+
+### Can I set an E-Mail Signature with Delta Chat? {#signature}
+
+Yes,
+you can do so under "Settings > Edit Profile > Signature Text".
```
```suggestion
you can do so under "Settings > Profile > Signature Text".
```

it is not "Edit Profile" on Android or iOS (you just tap the avatar/name under the headline "Your Profile" there). so, general "Profile" should work on all platforms, if ppl are already in settings, they will find it.

en/help.md
```diff
@@ -96,6 +96,38 @@ You may also wish to learn [how to add accounts to multiple devices](#multiclien
   device.
 
 
+### Can I set the E-Mail Subject with Delta Chat?
```
r10s: i would move "Can I set the E-Mail Subject with Delta Chat?" down after "If Delta Chat uses E-Mail, is it really an Instant Messenger?"

it is more like a advanced tip, and in general i have the idea to have the first part readable from top to bottom, more focusing on features and good parts. also the "E-Mail compat" does not fit to our current engagements.

the signature part can stay where it is.

en/help.md
```diff
@@ -96,6 +96,38 @@ You may also wish to learn [how to add accounts to multiple devices](#multiclien
...
+
+For most messages, Delta Chat sets the subject to
+"New Message from Alice" automatically,
+or "..." if it is encrypted.
+The exception are group chats,
+where the subject is the Chat Title.
```
```suggestion
Delta Chat sets (and [encrypts](#message-metadata)!) the classic e-mail subject
automatically to  the group name or to the sender name.
```

r10s (approved):
some comments, thanks a lot for pushing that forward - and sorry for huge delays :)

en/help.md
```diff
@@ -96,6 +96,38 @@ You may also wish to learn [how to add accounts to multiple devices](#multiclien
...
+set the Chat Title to the Subject you want,
+and then send your e-mail as the first message.
+If the recipients don't use Delta Chat,
+it will look like a normal,
+formal e-mail to them.
+You will get bonus professionalism points
+if you set a [signature](#signature).
+
+
+### Can I set an E-Mail Signature with Delta Chat? {#signature}
```
```suggestion
### Can I set a Signature (Status, Motto) with Delta Chat? {#signature}
```

i would leave out "E-Mail" in the question, it is explained below.

not sure about the brackets, though. but maybe it makes things clearer in case ppl do not know the term "Signature". for nerds, it also makes clear, that this is not a crypto thing :) (which is called "signature" as well at other places of the help)

PR #794, opened 2024-01-19
explain chatmail unencrypted restriction on get.delta.chat

missytake:
Until the error is explained nicely in the app, we should have some explanation on the download page - many users will probably still miss it (many users will probably not realize they can click on the QR code, there are still some things you need to know so this works)

### DIFF
--- a/en/download.md
+++ b/en/download.md
@@ -4,6 +4,26 @@ lang: en
 downloads: true
 ---
 
+## ❷ Get an Account
+
+You can use (almost) any email account with Delta Chat.
+If you need a new one,
+tap or scan this QR code
+to get a random [@nine.testrun.org](https://nine.testrun.org) email address:
+
+[![dcaccount:https://nine.testrun.org/cgi-bin/newemail.py](../assets/blog/nine-invite-qrcode.png)](dcaccount:https://nine.testrun.org/cgi-bin/newemail.py)
+
+## ❸ Enter a Username and Meet Your Friends
+
+Now you can enter a username,
+select a profile picture,
+and chat with your friends.
+**NOTE:** with a [@nine.testrun.org](https://nine.testrun.org/info.html) account
+you can't send unencrypted messages
+to users on other servers;
+meet them and verify them by [scanning QR codes](help#howtoe2ee)
+or ask them to send a message to you first.
+
 ![An iOS user scanning a QR code on someone else's phone.](../assets/blog/2023-11-qr-scan.jpg)
 
 ## Changelogs {#changelogs}
--- a/en/download.md
+++ b/en/download.md
@@ -6,7 +6,7 @@ downloads: true
 
 ## ❷ Get an Account
 
-You can use (almost) any email account with Delta Chat.
+You can use [(almost)](https://providers.delta.chat/) any email account with Delta Chat.
 If you need a new one,
 tap or scan this QR code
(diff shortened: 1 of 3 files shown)


r10s:
- the reasoning "cannot send unencrypted", therefore do a qr scan  is a bit complicated
- even if the contact sends a message first, this won't solve the issue

nb: can we shift that to markdown part, the yml is not really meant for long texts 
then we can also split the paragraphs, making translations easier

nb: it is planned for 1.44 that we show a concrete error; but mentioning the issue on pages still makes sense

( /me afk, so i cannot easily suggest line-wise :)

missytake:
>     * even if the contact sends a message first, this won't solve the issue

It does, as long as the contact uses an Autocrypt-capable client...

> nb: can we shift that to markdown part, the yml is not really meant for long texts then we can also split the paragraphs, making translations easier

good point, I added a commit. that would probably mean that some translations need to be re-done...

en/download.md
```diff
@@ -4,6 +4,26 @@ lang: en
 downloads: true
 ---
 
+## ❷ Get an Account
+
+You can use (almost) any email account with Delta Chat.
```
Simon-Laux: maybe a good place to link the provider db?

en/download.md
```diff
@@ -4,6 +4,26 @@ lang: en
...
+Now you can enter a username,
+select a profile picture,
+and chat with your friends.
+**NOTE:** with a [@nine.testrun.org](https://nine.testrun.org/info.html) account
+you can't send unencrypted messages
+to users on other servers;
+meet them and verify them by [scanning QR codes](help#howtoe2ee)
+or ask them to send a message to you first.
+
 ![An iOS user scanning a QR code on someone else's phone.](../assets/blog/2023-11-qr-scan.jpg)
```
r10s, line 27: i would tune down the limitation and make them less negative. also, by having a separate paragraph, we can easily remove if if 1.44 is released and a good error message is everywhere.

```suggestion
and connect with your friends by by [scanning QR codes](help#howtoe2ee).

![An iOS user scanning a QR code on someone else's phone.](../assets/blog/2023-11-qr-scan.jpg)

Note: @nine.testrun.org accounts block un-encrypted messages to recipients on other servers
but setting up contact via [QR codes](help#howtoe2ee) allows your messages to pass freely to any outside recipients.
```

r10s, line 27: i would tune down the limitation and make them less negative. also, by having a separate paragraph, we can easily remove if if 1.44 is released and a good error message is everywhere.

```suggestion
and connect with your friends by by [scanning QR codes](help#howtoe2ee).

![An iOS user scanning a QR code on someone else's phone.](../assets/blog/2023-11-qr-scan.jpg)

Note: @nine.testrun.org accounts block un-encrypted messages to recipients on other servers
but setting up contact via [QR codes](help#howtoe2ee) allows your messages to pass freely to any outside recipients.
```

r10s (commented):
> Until the error is explained nicely in the app

well, meanwhile it is explained nicely: https://github.com/deltachat/deltachat-android/pull/2940, https://github.com/deltachat/deltachat-ios/pull/2062 - of course, 1.44 is not shipped yet, but i would still tune the message down

hpk42:
this is now much better in the app, we can close this issue here i think. 

image

PR #796, opened 2024-02-05
Update help.md

jidanni:
For [#2953](https://github.com/deltachat/deltachat-android/pull/2953)


r10s (approved):
makes sense, thanks a lot!

i'll merge that in and will also care for translations

PR #799, opened 2024-02-13
improve add-second-device troubleshooting for windows

### LINKED ISSUE #3218: "Add Second Device" doesn't work on Windows because of firewall sometimes
- Operating System (Linux/Mac/Windows/iOS/Android): Windows 10
- Delta Chat Version: 1.36.1
- Expected behavior: The second device manages to connect to the sending instance of DC
- Actual behavior: Connection times out on the second device
  1. Install Delta Chat desktop (if you already have one, you may download the portable version)
  2. Install Delta Chat on another device (say, Android) (if you already have one, you can install an extra [nightly one](https://download.delta.chat/android/nightly/))
  3. On desktop, make sure you're logged in.
  4. On desktop, click "Settings" -> "Add Second Device" -> "Continue"
  5. On the second device, on the log in screen tap "Add as second device" and scan the QR code from the first device.
- Screenshots:
- Logs:

## Debugging

The connection fails, unless you have added DC desktop in "Allow an app through Windows Firewall", or unless it was added there in another way, say through the Windows Firewall alert that can pop up when an app tries to listen on a non-localhost network interface.

The "Add Second Device" feature was added here https://github.com/deltachat/deltachat-core-rust/pull/4007, and it uses `iroh`. The binding address is specified [here](https://github.com/deltachat/deltachat-core-rust/blob/edfdbbdc90e97de92b5f86a74ecddb3be6a8aa9a/src/imex/transfer.rs#L182), down to [here](https://docs.rs/iroh/0.4.1/src/iroh/provider/mod.rs.html#191), but it doesn't trigger the Windows Firewall alert popup, like it usually does with, say, a simple HTTP server app.

I have a suspicion that it may be related to the fact that DC listens for UDP packets, not TCP. [`iroh`](https://docs.rs/iroh/latest/iroh/index.html) uses the QUIC protocol, which is UDP-based. I just tried

```rust
std::net::UdpSocket::bind(("0.0.0.0", 0))?;
```

And it did not trigger the alert, but with a concrete port `("0.0.0.0", 59999)` the alert _is_ triggered. And, again, yes, if you set up an TCP listener, say, with

```rs
std::net::TcpListener::bind(("0.0.0.0", 0))?
```

The alert _does_ get triggered.

## Possible solutions

* A simple hack would be to create another dummy (TCP) interface binding at the same time we create the real (QUIC) one, in order to trigger the Windows Firewall alert and let the user decide whether to allow DC through firewall (see https://github.com/deltachat/deltachat-core-rust/pull/4351)
* @Simon-Laux and @flub pointed me to the `windows-rs` library that DC already depends on, namely these two functions [1](https://github.com/microsoft/windows-rs/blob/a672f5f450b16c947840ab89fdc5ddd16d1fd0d0/crates/libs/windows/src/Windows/Win32/NetworkManagement/WindowsFirewall/impl.rs#L1471), [2](https://github.com/microsoft/windows-rs/blob/a672f5f450b16c947840ab89fdc5ddd16d1fd0d0/crates/libs/windows/src/Windows/Win32/NetworkManagement/WindowsFirewall/impl.rs#L950). But I myself am not sure how they work. I heard they may require elevation ("Run as administrator"). But not only it is a proper solution, I also suspect that it _might_ actually be better because it _might_ (based on my intuition) cause the Firewall alert to pop up a second time if network access was previously denied for the application.

r10s:
explicitly mention to switch to 'Private Network' on windows add-second-device troubleshooting, could not check on an english windows, but searching for the terms, these seems to be the wordings used by windows

i moved the point to the second-option, directly after same-network, as it seems to be a pretty concrete issue on windows.

closes https://support.delta.chat/t/add-hint-select-on-windows-desktop-private-network-to-add-a-second-device/2921
closes https://github.com/deltachat/deltachat-desktop/issues/3218

this will also go to the offline helps once merged.

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -753,6 +753,9 @@ One device is not needed for the other to work.
   which is known to cause problems (especially on Windows).
   **Disable the personal firewall** for Delta Chat on both ends and try again
 
+- **Guest Networks** may not allow devices to communicate with each other.
+  If possible, use a non-guest network.
+
 - Ensure there is **enough storage** on the destination device
 
 - If transfer started, make sure, the devices **stay active** and do not fall asleep.
--- a/en/help.md
+++ b/en/help.md
@@ -747,6 +747,7 @@ One device is not needed for the other to work.
 
 - On **Windows**, go to **Control Panel / Network and Internet**
   and make sure, **Private Network** is selected as "Network profile type"
+  (after transfer, you can change back to the original value)
 
 - Your system might have a "personal firewall",
   which is known to cause problems (especially on Windows).
--- a/en/help.md
+++ b/en/help.md
@@ -745,6 +745,9 @@ One device is not needed for the other to work.
 
 - Double-check both devices are in the **same Wi-Fi or network**
 
+- On **Windows**, go to **Control Panel / Network and Internet**
+  and make sure, **Private Network** is selected as "Network profile type"
+
 - Your system might have a "personal firewall",
   which is known to cause problems (especially on Windows).
   **Disable the personal firewall** for Delta Chat on both ends and try again


dev24126:
> explicitly mention to switch to 'Private Network' on windows add-second-device troubleshooting, could not check on an english windows, but searching for the terms, these seems to be the wordings used by windows
> 
> i moved the point to the second-option, directly after same-network, as it seems to be a pretty concrete issue on windows.
> 
> closes https://support.delta.chat/t/add-hint-select-on-windows-desktop-private-network-to-add-a-second-device/2921 closes [deltachat/deltachat-desktop#3218](https://github.com/deltachat/deltachat-desktop/issues/3218)
> 
> this will also go to the offline helps once merged.

I think it would also be helpful to add the comment that the Network profile can be set back to the default or "Public Network" after the second device for DeltaChat is added.

Reason: "Public Network" setting is more secure in public networks, sometimes also in Private Networks ;)

r10s:
i added a sentence, _that_ the user can reset "Network profile type" after transfer - _why_ and reasoning is better explained in the windows dialog itself and not part of troubleshooting (also we do not want to maintain other windows network profile types and explain them)

r10s:
... and i dropped another line for guest-networks as popular eg. in germany with the "fritz box": these routers have an option for a "guest network" - which then has the option that the devices do not see each other - i just stumbled upon  that on testing the windows issue :)

Simon-Laux (approved):

PR #804, opened 2024-03-04
update `keytool` example command

r10s:
using `-list` and `-printcert` at the same time just prints "keytool error: java.lang.Exception: Only one command is allowed: both -list and -printcert were specified."

and also according to
https://docs.oracle.com/javase/6/docs/technotes/tools/solaris/keytool.html , these commands seem mutually exclusive.

came over that when checking https://support.delta.chat/t/allow-the-users-to-verify-the-downloads-for-desktop/2953


missytake (approved):
thanks for noticing.

PR #811, opened 2024-03-12
download: mention invite links

missytake:
This is much easier for most users than scanning QR codes.

Maybe we could also leave out scanning QR codes here, and change the image so users can see how they can create an invite link.

### DIFF
--- a/_data/lang/en.yaml
+++ b/_data/lang/en.yaml
@@ -26,7 +26,7 @@ en:
     get_account: Get an Account
     get_account_explanation: You can use (almost) any email account with Delta Chat. If you need a new one, tap or scan this QR code to get a random <a href="https://nine.testrun.org" target="_blank">@nine.testrun.org</a> email address
     add_friends: Enter a Username and Meet Your Friends
-    add_friends_explanation: Now you can enter a username, select a profile picture, and chat with your friends. The best way is to meet them and verify them by <a href="help#howtoe2ee">scanning QR codes</a>.
+    add_friends_explanation: Now you can enter a username, select a profile picture, and chat with your friends. The best way is to <a href="../en/2024-03-12-jumbo44#sharing-invite-links-for-chats-and-chat-groups">send them an invite link</a> or meet them and verify them by <a href="help#howtoe2ee">scanning QR codes</a>.
     other: Download options for other platforms
     source_code: Source Code
     install_via_packagemanager: Install via Package Manager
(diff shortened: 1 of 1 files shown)


hpk42 (approved):

_data/lang/en.yaml
```diff
@@ -26,7 +26,7 @@ en:
     get_account: Get an Account
     get_account_explanation: You can use (almost) any email account with Delta Chat. If you need a new one, tap or scan this QR code to get a random <a href="https://nine.testrun.org" target="_blank">@nine.testrun.org</a> email address
     add_friends: Enter a Username and Meet Your Friends
-    add_friends_explanation: Now you can enter a username, select a profile picture, and chat with your friends. The best way is to meet them and verify them by <a href="help#howtoe2ee">scanning QR codes</a>.
+    add_friends_explanation: Now you can enter a username, select a profile picture, and chat with your friends. The best way is to <a href="../en/2024-03-12-jumbo44#sharing-invite-links-for-chats-and-chat-groups">send them an invite link</a> or meet them and verify them by <a href="help#howtoe2ee">scanning QR codes</a>.
```
r10s: lots of "them" and a relatively long sentence. i suggest the following, that is shorter and  also avoids the "best" wording, see below

```suggestion
    add_friends_explanation: Now you can enter a username, select a profile picture, and chat with your friends. Meet and verify them by <a href="help#howtoe2ee">scanning QR codes</a> or <a href="../en/2024-03-12-jumbo44#sharing-invite-links-for-chats-and-chat-groups">send them an invite link</a>.
```

r10s:
> This is much easier for most users than scanning QR codes.

thinking it over, not sure, if that is true in general.

i think, it depends strongly on the situation, what is more _useful_.

but _easier_ is scanning each other's QR code - and on initial onboarding, there is also some chance ppl are really sitting beside each other. QR codes also have a good existing understanding meanwhile, and you can do everything in-app.

invite-links involve 3rd party software, so it adds more fraction, is less straight-forward, involves context switches and may end up in a mess. it is not so easy as "do that, and you're done", it is more advanced.

this is much better, if one just _receives_ an invite link through some channel ppl are using anyways, for sure.

that said, i totally agree with mentioning invite links, but we should not go all-in. both are reasonable options, there is no "best", at least not currently

missytake:
> but _easier_ is scanning each other's QR code

The "easier" wording was not optimal. QR scanning is often easier, especially if you know the steps and know what you are doing. But the most **familiar** way is clicking on an invite link, that is the behavior people know from other apps.

> and on initial onboarding, there is also some chance ppl are really sitting beside each other. QR codes also have a good existing understanding meanwhile, and you can do everything in-app.

We can guide users to use QR codes in more situations, we've been trying to nudge users towards that for a couple releases now. Whether this has success needs to be shown in user testing imho, I don't see this working well yet in my daily life.

For power users, QR code scanning is often useful, especially if you are setting up multiple devices at the same time. But during onboarding, users are in any case much less familiar with QR code scanning than with invite links.

> that said, i totally agree with mentioning invite links, but we should not go all-in. both are reasonable options, there is no "best", at least not currently

With that I agree.

PR #812, opened 2024-03-12
fix typos and CHANGLOG links

r10s:


r10s:
preview works, will merge that in

PR #817, opened 2024-03-22
add AppGallery download option

### LINKED ISSUE #816: add appgallery to get.delta.chat
add a logo and link to https://appgallery.huawei.com/app/C108252361 , similar to all the other stores

r10s:
this PR adds the option to download Delta Chat via Huawei's AppGallery.

button is not localised currently, but that is true also for many other's, not sure if that is really needed, maybe chinese etc. - but for now, it seems good enough and a huge improvement to be in the store at all

closes #816


link2xt:
I think it should be somewhere around Amazon store, on the second row, not on the second place even before F-Droid.

link2xt (approved):

r10s:
> I think it should be somewhere around Amazon store, on the second row, not on the second place even before F-Droid.

makes sense at current stage, we can reconsider at any time. i changed that and force-pushed.

PR #819, opened 2024-03-25
Build website with nix

link2xt:
`nix build` builds website in `result/`. `nix develop` starts a shell where you can run `jekyll serve`.

### DIFF
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -225,7 +225,7 @@ blockquote { font-style: italic; }
 h1,h2,h3,h4,h5,h6 {
     &:hover {
         .anchor::before {
-            content: "🔗"
+            content: "\1F517"
         }
     }
 }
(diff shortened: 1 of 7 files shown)


_sass/_base.scss
```diff
@@ -225,7 +225,7 @@ blockquote { font-style: italic; }
 h1,h2,h3,h4,h5,h6 {
     &:hover {
         .anchor::before {
-            content: "🔗"
+            content: "\1F517"
```
link2xt, line 228: Version of `jekyll` I have is unhappy about emojis in .scss and failed with `Encoding::UndefinedConversionError`. Using nix also ensures I am using exactly the same version of Jekyll that runs in CI so such problems can be avoided.

missytake (approved):
seems to work? didn't try out locally because I still don't have the courage to let this installer loose on my xubuntu :upside_down_face:

PR #824, opened 2024-04-01
download: recommend to install flatpak with --user

missytake:
The issue was brought up at https://support.delta.chat/t/deltachat-flatpak-not-opening-file-manager-to-allow-attaching-images-or-files/3001 - until it's fixed maybe we should recommend `--user` instead.


r10s (approved):
`--user` probably installs the app only for the current user, not for all users, right? i assume, there is no consent anyway here, so, lgtm, esp. as it mitigates a concrete issue.

however, in addition, we should file an issue for desktop, just pointing to https://support.delta.chat/t/deltachat-flatpak-not-opening-file-manager-to-allow-attaching-images-or-files/3001 as a minimum

missytake:
I think it makes more sense if the issue is opened by the original reporter herself, I asked her to report it on GitHub.

PR #825, opened 2024-04-03
Minor text changes

gerryfrancis:


en/help.md
```diff
@@ -86,7 +86,7 @@ You may also wish to learn [how to add accounts to multiple devices](#multiclien
   or add them via QR code, they automatically see it as your profile picture.
 
 - Contacts who don't use Delta Chat do not see the profile picture
-  (however, of course, they can install Delta Chat :)
+  (however, of course, they can install Delta Chat). :)
```
r10s, line 89: https://xkcd.com/541/

r10s:
thanks a lot, however, i am closing this here as most changes only refer to casing and punctuation that is discussable anyways.

even if we would assume your suggestions are better: if we would merge this PR, this would result in lots of load for translators, and the translations would become removed for weeks, maybe for much longer (unfortunately, [in contrast to app-strings](https://github.com/deltachat/deltachat-android/blob/main/scripts/tx-update-changed-sources.sh), for $reasons, there is currently no way to change the english source without triggering translations)

so, merging this PR would overall worsen things.

therefore, in general, we avoid such reformatting where possible and correct only "hard" errors. reformatting is better done when larger parts are changed anyways for other reasons.

gerryfrancis:
> merging this PR would overall worsen things.

@r10s I did not know that, but hey, no problem, thank you for the detailed explanation! :)

PR #830, opened 2024-04-05
Scratch remark about QR icon in help.md

gerryfrancis:
Fixes #829

PR #831, opened 2024-04-05
FAQ: mention invite links

missytake:


r10s:
this PR needs a rebase or needs to be redone

en/help.md
```diff
@@ -350,7 +350,11 @@ One of you is the "Inviter", the other is the "Joiner".
 - Direct 1:1 chat invitation: 
   Tap the QR Code icon <img style="vertical-align:middle; width:1.8em; margin:1px" src="../assets/help/qr-icon.png" />
   on the Delta Chat app main screen.
-  Share the QR image with the other side either in person or through a second channel.
+  Show the QR code to the other person for scanning.
```
r10s: better leave this as is, the "second channel" is a reference to above  and the same wording is also used for "Group Invitations". it is good as is, no need to force retranslations here.

en/help.md
```diff
@@ -350,7 +350,11 @@ One of you is the "Inviter", the other is the "Joiner".
...
+
+If the QR code scanning doesn't work in your situation,
+you can also copy an <https://i.delta.chat> invite link to clipboard
+and share it with the other person through a second channel.
```
r10s: at this position, the paragraphs disturbs a bit the flow and the idea of a "QR scan-show checklist" as it is currently.

(if ppl go for the invite-link-alternative, the paragraph "Joiner side:" is no longer valid/needed)

therefore, i'd first describe "QR scan-show checklist" as it is currently (it is short enough), and then before "Both Inviter and Joiner:" (which is valid again also for invite links) go for invite links.

moreover, i'd leave out the https://i.delta.chat  link as it is not helpful in most cases and mention **Share** as well (it is the primary option on android / ios even)

```suggestion
If the QR code scanning doesn't work in your situation,
you can also **Share** (or **Copy to Clipboard**) an equivalent invite link
to the other person through a second channel.
```

en/help.md
r10s: i added a commit to avoid this PR getting stale

en/help.md
r10s: i added a commit to avoid this PR getting stale

r10s:
i rebased and added a commit that leaves the QR checklist alone and inserts a pointer to "Invite Links" afterwards

the QR code icon position description for desktop was already changed at https://github.com/deltachat/deltachat-pages/pull/830

missytake:
Great :)

PR #834, opened 2024-04-17
download: hide non-automatic download options behind <details>

### LINKED ISSUE #823: demote downloads of Android and Desktop-release-files because of missing automatic upgrades
We should not recommend as primary options any downloads that do not provide automatic updates.
So i suggest to demote

- [ ] .exe 
- [ ] .rpm
- [ ] .deb 
- [ ] .tar.gz
- [ ] .AppImage
- [ ] apk 
- [ ] apklis (because it's outdated) 
- [ ] dmg

to not have images but be listed in the fine-print.

missytake:
fix #823 

With these `<details>` tags I always wonder if people understand that they can click on it... maybe a sub-headline would be better in this case.

To give examples, I doubt a cuban user would find apklis now (which is maybe fine, as it's outdated there), and a ubuntu user who is overwhelmed by flatpak (friend of mine) would have a hard time finding .deb (but then again .deb failed for her because of an oudated ubuntu version).

I'm looking forward to an apt repository so much^^

### DIFF
--- a/_sass/_downloadpage.scss
+++ b/_sass/_downloadpage.scss
@@ -77,8 +77,7 @@ div.download-content {
 }
 
 summary.noupgrades {
-    color: #1e6bb8;
-    text-decoration: none;
+    font-weight: bold;
     cursor: pointer;
 }
 
--- a/_sass/_downloadpage.scss
+++ b/_sass/_downloadpage.scss
@@ -82,8 +82,8 @@ summary.noupgrades {
 }
 
 summary {
-  list-style-type: '➡ ';
+  list-style-type: '> ';
 }
 details[open] > summary {
-  list-style-type: '⬆ ';
+  list-style-type: '^ ';
 }
--- a/_sass/_downloadpage.scss
+++ b/_sass/_downloadpage.scss
@@ -75,3 +75,9 @@ div.download-content {
     font-weight: bold;
     cursor: pointer;
 }
+
+summary.noupgrades {
+    color: #1e6bb8;
+    text-decoration: none;
+    cursor: pointer;
+}
--- a/_sass/_downloadpage.scss
+++ b/_sass/_downloadpage.scss
@@ -81,3 +81,10 @@ summary.noupgrades {
     text-decoration: none;
     cursor: pointer;
 }
+
+summary {
+  list-style-type: '➡ ';
+}
+details[open] > summary {
+  list-style-type: '⬆ ';
+}
--- a/_sass/_downloadpage.scss
+++ b/_sass/_downloadpage.scss
@@ -77,13 +77,10 @@ div.download-content {
 }
 
 summary.noupgrades {
-    font-weight: bold;
+    color: #1e6bb8;
     cursor: pointer;
+    list-style: none;
 }
-
-summary {
-  list-style-type: '> ';
-}
-details[open] > summary {
-  list-style-type: '^ ';
+summary.noupgrades:hover {
+    text-decoration: underline;
 }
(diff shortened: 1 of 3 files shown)


r10s:
> With these `<details>` tags I always wonder if people understand that they can click on it.

i think, ppl will find that. but we can vastly improve that by letting it looks as a link, so exactly similar to the "Source Code" link.

that would also clutter UI less, as there is no other style introduces.

wrt less cluttering: we can short the link to "Options without automatic updates"

i would not make the link larger - the idea is to tune it down, ppl shall use one of the other options, only if that does not fit,  they can look for sth. else

btw. tbh, when i only read the headline of the PR, i was first sceptical about the `<details>` tag, i was more thinking about tuning the size down and moving down - but i think, it's not that bad

missytake:
> > With these `<details>` tags I always wonder if people understand that they can click on it.
> 
> i think, ppl will find that. but we can vastly improve that by letting it looks as a link, so exactly similar to the "Source Code" link.

I tried that, but then it looked weird because it was so different from "Download options for other platforms". I changed the arrow and the cursor when hovering, I hope that helps.

In general, this guide has more interesting ideas about styling a `<details>` tag: https://www.sitepoint.com/style-html-details-element/

> wrt less cluttering: we can short the link to "Options without automatic updates"

not a fan, now it looks similare to "Download options for other platforms", that is aesthetic and does make sense - and saying explicitly that the options are *download* options makes sense in both cases.

missytake:
Unfortunately this commit breaks the CI: https://github.com/deltachat/deltachat-pages/pull/834/commits/bcad7caa007393bd1befd3213a1a112606ff19f4

locally it works, but we use the CI for deploying, so... :/

Simon-Laux:
Looks good

hpk42 (approved):
Lgtm but don't understand why ci is unhappy

hpk42:
> > With these `<details>` tags I always wonder if people understand that they can click on it.
> 
> i think, ppl will find that. but we can vastly improve that by letting it looks as a link, so exactly similar to the "Source Code" link.

It's worth a try without? .... Let's see how many people really don't notice they can't click the details?

_sass/_downloadpage.scss
```diff
@@ -74,4 +74,16 @@ div.download-content {
...
+    font-weight: bold;
+    cursor: pointer;
+}
+
+summary {
+  list-style-type: '> ';
+}
+details[open] > summary {
+  list-style-type: '^ ';
+}
```
r10s: i think, better remove these lines, they seem not to work on at least some browsers (safari) - and the default arrow is just fine and known enough

_sass/_downloadpage.scss
```diff
@@ -74,4 +74,16 @@ div.download-content {
...
     font-weight: bold;
     cursor: pointer;
-}
\ No newline at end of file
+}
+
+summary.noupgrades {
+    font-weight: bold;
+    cursor: pointer;
+}
```
r10s: i would reuse the style of links here:

- it looks nicer, less cluttered, we already have so many different fonts and sizes on the download page, i'd not add another 

- it makes clear that the thing is tappable - even if one does not know what the triangle means 

- still, it is less bold

- imo, it looks somehow less "nested" and complicated, when otherwise looking at the two triangles with different levels

```suggestion
summary.noupgrades {
    font-weight: normal;
    color: #1e6bb8;
    cursor: pointer;
}

summary.noupgrades:hover {
    text-decoration: underline;
}
```

we can also consider to move "Source Code" to the bottom of each box, then the thing would look even more sorted, as links with the same color are together. also, it makes some sens to have source code further down as it does not help for downloading but is more a footnote

r10s (approved):
this is what i would change, minors, it is already without a large improvement

missytake:
You are right, as links but without a `<details>` list item it looks quite nice, but clickable. It is very small now, but well. I like it.

PR #840, opened 2024-04-24
update translations, make chatmail.md translatable

r10s:
this PR follows the instructions in the README.md to make a page (here: the new `en/chatmail.md`) translatable


r10s:
postponed, the text is subject to changed. pushing to translators is a little bit too soon.

~~in the ui PR, we should just link to the english page for now the (so, https://delta.chat/en/chatmail (with /en/) instead of https://delta.chat/chatmail (without /en/)~~ EDIT: no, we use the "final" link `https://delta.chat/chatmail` - it is just some nightlies and  betas that may get 404

PR #843, opened 2024-04-26
We neither accept nor ignore warnings ;-)

gerryfrancis:


r10s:
both are probably fine - but is "don't ignore" really so much better than "don't accept" to justify retranslations?

note that the text was pretty much reviewed already by native speakers, so i cannot really judge the english part - but in german, also both would be fine "akzeptieren Sie die Warnung nicht!" vs. "ignorieren Sie die Warnung nicht!".

the "don't accept" wording looks even better to me as it has a  more concrete instruction for action, _not_ to tap "ok".

gerryfrancis:
@r10s Alright, no problem! :) Closing here...

PR #846, opened 2024-04-27
reorder security section and put dates in front -- no text changes

hpk42:


r10s (approved):
yes, with newest-atop, the order is much nicer.

ftr: all changed paragraphs will be triggered for retranslation as it is not only ordering but content has changed (by reformatting date). but as this is an important part of the help, this seems worth the effort - also, translators can see the history and get a hint that only few has changed (and what exactly)

PR #847, opened 2024-05-01
add anchor for multi client troubleshooting

Simon-Laux:
so that this can be linked in the local inApp help

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -797,7 +797,7 @@ both **devices are completely independent.**
 One device is not needed for the other to work.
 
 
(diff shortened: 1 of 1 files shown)


r10s:
i do not think that skipping the introduction is necessarily an improvement or worth the noise of dealing with retranslations and/or untranslated headlines next week, i commented in more detail at https://github.com/deltachat/deltachat-desktop/pull/3801#issuecomment-2089093981

Simon-Laux:
closing for now then

PR #853, opened 2024-05-04
move 'Advanced' topcis down, no text changes

r10s:
'Advanced' topics should be tuned down,
so that ppl reading the help
are not immediately confronted
with questions that do not tangent recommended workflows.

The one or the other wording or presence is probably questionable, but this PR is not about rewording, just about resorting, which avoids retranslations at all and make things already much nicer.


missytake (approved):
good idea :)

PR #856, opened 2024-05-05
fix switch/add account help for desktop

r10s:
switch/add account help for desktop was wrong; but also for android/ios it could be easier.

we have "account controls" at the same place on all platforms now, make use of that :)


missytake (approved):
Great :)

PR #867, opened 2024-05-12
revise home page messaging

hpk42:
after some testing of wordings in the fediverse, and discussions with bjoern and adb, this is my suggestion in line with the discussions on  1.46 release messaging. Can be merged before it's out because it's pretty true already :)

- emphasizes instantness, reliability, decentralization, interactive apps, audited security
- overall de-emphasizes e-mail, now contained in "Internet Standards" 
- reduces visual clutter on the site 

The Screenshot-section should be improved separately, needs updates and more spirit -- see https://github.com/deltachat/deltachat-pages/issues/868 -- and overall layout question are also out of scope for this PR, please :)

### DIFF
--- a/en/index.md
+++ b/en/index.md
@@ -1,9 +1,9 @@
 ---
-title: Delta Chat Homepage
+title: Delta Chat, decentralized private messenger 
 lang: en
 ---
 
-# Delta Chat is a standards-based private messenger 
+# Delta Chat is a decentralized private messenger app
 
 <br>
 
--- a/en/index.md
+++ b/en/index.md
@@ -3,7 +3,7 @@ title: Delta Chat, decentralized secure messenger
 lang: en
 ---
 
-# Delta Chat is a decentralized & secure messenger app
+# Delta Chat is a decentralized and secure messenger app
 
 <br>
 
@@ -11,11 +11,11 @@ lang: en
     💬 Reliable instant messaging with
        multi-profile and multi-device support
    <br/><br/>
-   💨 Sign up to [secure & fast chatmail](chatmail) or use [traditional e-mail](https://providers.delta.chat/) servers
+   💨 Sign up to [secure fast chatmail servers](chatmail) or use [traditional e-mail servers](https://providers.delta.chat/) 
    <br/><br/>
    🥳 Interactive [web apps in chats](https://webxdc.org/) for gaming and collaboration
    <br/><br/>
-   🔒 [Audited end-to-end encryption](https://delta.chat/en/2024-03-25-crypto-analysis-securejoin) safe against network & server attacks
+   🔒 [Audited end-to-end encryption](https://delta.chat/en/2024-03-25-crypto-analysis-securejoin) safe against network and server attacks
    <br/><br/>
    👉 [F](https://en.wikipedia.org/wiki/Free_software)[OSS](https://en.wikipedia.org/wiki/Open-source_software) software, built on [Internet Standards](https://github.com/deltachat/deltachat-core-rust/blob/master/standards.md), avoiding [xkcd927](https://xkcd.com/927/) :)
 </span>
--- a/en/index.md
+++ b/en/index.md
@@ -3,15 +3,23 @@ title: The e-mail messenger
 lang: en
 ---
 
-# Delta Chat is a messaging app that works over e-mail 
+# Delta Chat is a unique decentralized messenger 
 
-💬 Message anyone with an e-mail address [even if they don't use Delta Chat](https://www.youtube-nocookie.com/embed/8LbrGXKZN70).
+<br>
 
-🥳 Enjoy interactive chat experiences through [webxdc apps](https://webxdc.org).
+<span style="font-size: 1.5em">
+    💬 Reliable instant messaging with
+       multi-profile and multi-device support
+   <br/><br/>
+   🥳 Interactive [web apps in chats](https://webxdc.org/) for gaming and collaboration
+   <br/><br/>
+   🔒 Audited end-to-end encryption safe against network & server attacks
+   <br/><br/>
+   👉 [FO](https://en.wikipedia.org/wiki/Free_software)[SS](https://en.wikipedia.org/wiki/Open-source_software) software, built on [Internet Standards](https://github.com/deltachat/deltachat-core-rust/blob/master/standards.md), avoiding [xkcd927](https://xkcd.com/927/) :)
+</span>
 
-🔒 End-to-End Encryption using [Autocrypt](https://autocrypt.org) and [SecureJoin](https://securejoin.delta.chat/en/latest/new.html) protocols, with [multiple security audits](https://delta.chat/en/2023-03-27-third-independent-security-audit). 
 
-# Available on mobile and desktop
+# Available on mobiles and desktop
 
 
 <div>
@@ -56,7 +64,5 @@ lang: en
 
 [Download](https://get.delta.chat){: .cta-button}
 
-[Open Source](https://en.wikipedia.org/wiki/Open-source_software)
-and [Free Software](https://en.wikipedia.org/wiki/Free_software), built on [Internet Standards](https://github.com/deltachat/deltachat-core-rust/blob/master/standards.md). 
 
-Check out [user stories and voices](user-voices).
+
--- a/en/index.md
+++ b/en/index.md
@@ -17,7 +17,7 @@ lang: en
    <br/><br/>
    🔒 [Audited end-to-end encryption](https://delta.chat/en/2024-03-25-crypto-analysis-securejoin) safe against network & server attacks
    <br/><br/>
-   👉 [FO](https://en.wikipedia.org/wiki/Free_software)[SS](https://en.wikipedia.org/wiki/Open-source_software) software, built on [Internet Standards](https://github.com/deltachat/deltachat-core-rust/blob/master/standards.md), avoiding [xkcd927](https://xkcd.com/927/) :)
+   👉 [F](https://en.wikipedia.org/wiki/Free_software)[OSS](https://en.wikipedia.org/wiki/Open-source_software) software, built on [Internet Standards](https://github.com/deltachat/deltachat-core-rust/blob/master/standards.md), avoiding [xkcd927](https://xkcd.com/927/) :)
 </span>
 
 
--- a/en/index.md
+++ b/en/index.md
@@ -11,7 +11,7 @@ lang: en
     💬 Reliable instant messaging with
        multi-profile and multi-device support
    <br/><br/>
-   💨 No number, email or data needed to use interoperable [chatmail servers](https://delta.chat/en/chatmail)
+   💨 No number, e-mail or data needed to use interoperable [chatmail servers](https://delta.chat/en/chatmail)
    <br/><br/>
    🥳 Interactive [web apps in chats](https://webxdc.org/) for gaming and collaboration
    <br/><br/>
--- a/en/index.md
+++ b/en/index.md
@@ -1,9 +1,9 @@
 ---
-title: Delta Chat, decentralized private messenger 
+title: Delta Chat, decentralized secure messenger 
 lang: en
 ---
 
-# Delta Chat is a decentralized private messenger app
+# Delta Chat is a decentralized & secure messenger app
 
 <br>
 
@@ -11,7 +11,7 @@ lang: en
     💬 Reliable instant messaging with
        multi-profile and multi-device support
    <br/><br/>
-   💨 No number, e-mail or data needed to use interoperable [chatmail servers](https://delta.chat/en/chatmail)
+   💨 Sign up to [secure & fast chatmail](chatmail) or use [traditional e-mail](https://providers.delta.chat/) servers
    <br/><br/>
    🥳 Interactive [web apps in chats](https://webxdc.org/) for gaming and collaboration
    <br/><br/>
--- a/en/index.md
+++ b/en/index.md
@@ -21,7 +21,7 @@ lang: en
 </span>
 
 
-# Available on mobiles and desktop
+# Available on mobile and desktop
 
 
 <div>
--- a/en/index.md
+++ b/en/index.md
@@ -1,9 +1,9 @@
 ---
-title: The e-mail messenger
+title: Delta Chat Homepage
 lang: en
 ---
 
-# Delta Chat is a unique decentralized messenger 
+# Delta Chat is a standards-based private messenger 
 
 <br>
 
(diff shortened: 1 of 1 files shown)


adbenitez (approved):

en/index.md
```diff
@@ -3,15 +3,23 @@ title: The e-mail messenger
...
+<span style="font-size: 1.5em">
+    💬 Reliable instant messaging with
+       multi-profile and multi-device support
+   <br/><br/>
+   🥳 Interactive [web apps in chats](https://webxdc.org/) for gaming and collaboration
+   <br/><br/>
+   🔒 Audited end-to-end encryption safe against network & server attacks
+   <br/><br/>
+   👉 [FO](https://en.wikipedia.org/wiki/Free_software)[SS](https://en.wikipedia.org/wiki/Open-source_software) software, built on [Internet Standards](https://github.com/deltachat/deltachat-core-rust/blob/master/standards.md), avoiding [xkcd927](https://xkcd.com/927/) :)
+</span>
```
r10s: i think, the following linking pattern makes a little more sense: F - for "Free" and "OSS" for "Open Source Software"

```suggestion
   👉 [F](https://en.wikipedia.org/wiki/Free_software)[OSS](https://en.wikipedia.org/wiki/Open-source_software) software, built on [Internet Standards](https://github.com/deltachat/deltachat-core-rust/blob/master/standards.md), avoiding [xkcd927](https://xkcd.com/927/) :)
</span>
```

r10s (approved):
very nice, big improvement.

we can merge as is, and it is good to focus on the text. technically, before pushing to translators, i like to declutter the HTML block so it is easier translatable as text

en/index.md
```diff
@@ -1,17 +1,27 @@
 ---
-title: The e-mail messenger
+title: Delta Chat Homepage
```
missytake: That it is a homepage is obvious already. What about "Delta Chat, the e-mail messenger" ?

en/index.md
```diff
@@ -1,17 +1,27 @@
...
 lang: en
 ---
 
-# Delta Chat is a messaging app that works over e-mail 
+# Delta Chat is a standards-based private messenger
```
missytake: Every time I explain Delta Chat to someone, I say "it's either a messaging app that works over e-mail, or an e-mail client that looks like a messenger, depending on the perspective". For this, you'd need to understand the role of "standards" in the IT world, and now we don't even say which standards. This is very abstract, I'm not a fan. Maybe if I'd participated in more of the discussions around this change.

en/index.md
```diff
@@ -1,17 +1,27 @@
...
+    💬 Reliable instant messaging with
+       multi-profile and multi-device support
+   <br/><br/>
+   💨 No number, e-mail or data needed to use interoperable [chatmail servers](https://delta.chat/en/chatmail)
+   <br/><br/>
+   🥳 Interactive [web apps in chats](https://webxdc.org/) for gaming and collaboration
+   <br/><br/>
+   🔒 [Audited end-to-end encryption](https://delta.chat/en/2024-03-25-crypto-analysis-securejoin) safe against network & server attacks
+   <br/><br/>
+   👉 [F](https://en.wikipedia.org/wiki/Free_software)[OSS](https://en.wikipedia.org/wiki/Open-source_software) software, built on [Internet Standards](https://github.com/deltachat/deltachat-core-rust/blob/master/standards.md), avoiding [xkcd927](https://xkcd.com/927/) :)
```
missytake, line 20: Hm, this also addresses a very narrow target group; for our main page, maybe too narrow imho.

en/index.md
```diff
@@ -1,17 +1,27 @@
...
+</span>
 
-🔒 End-to-End Encryption using [Autocrypt](https://autocrypt.org) and [SecureJoin](https://securejoin.delta.chat/en/latest/new.html) protocols, with [multiple security audits](https://delta.chat/en/2023-03-27-third-independent-security-audit). 
 
-# Available on mobile and desktop
+# Available on mobiles and desktop
```
missytake: "on mobile phones" maybe?

missytake: "on mobile phones" maybe?

missytake (commented):
What is the target group of this page? It seems to me you mostly want to address a quite nerdy crowd; not sure if this is very valuable. We already have a fediverse presence which is strongly geared towards that audience.

I think for our main page, we should keep the threshold low. We should explain in the easiest way possible what Delta Chat is. This raises the likelihood that people who we don't know yet will like the project and can interact productively with it.

Some of the changes are an improvement for sure though.

en/index.md
kermoshina: Agree that it might be too techy. What about "Free and open source software, based on robust standards"?

en/index.md
kermoshina, line 1: Something in the word "unique" makes me hesitate... Why dont we stay with secure? Secure decentralized messenger?

It is unique indeed, but the word unique is overused...

en/index.md
kermoshina: And desktops? In plural

kermoshina (approved):

en/index.md
kermoshina: Yeah I also mention the "based on email"

en/index.md
hpk42: the word "unique" is gone from the latest draft, maybe you looked at an old one?

en/index.md
hpk42: reverted to old "Available on mobile and desktop" -- i think "mobile" is a standing term but maybe plural is weird.

hpk42: reverted to old form for now.

hpk42:
kermoshina, xenia: thanks for the feedback and thoughts!

**Target group:** intended are multipliers, developers, might-become contributors but also more deeply interested users, who know properties of other messengers.  Few mainstream users will see and study this web page that doesn't even have a video.  They rather go to the store pages if even that -- it's more that they get told to try it and then it depends on our in-app guidance and easy-to-communicate steps how to get started: users need to explain other users how to succeed.  If we want to focus on mainstream users with our communication we need to do and instigate videos. 

**As to "E-mail"**, the lack of focusing on it is intentional and tied to the 1.46 release goal: you should not need to know or care that you are using e-mail in order to succeed.  Better not think first about using your existing e-mail account because that requires a lot of understanding and there are many known issues (rate-limits, double-notifications with shared usage with other e-mail apps, spam/reachability problems, private key management if you are encrypting etc.) that we can't fix.  Also, while there are many e-mail enthusiasts, large factions of expert/multiplyer communities disregard e-mail out of hand, as kind of a showstopper.  True, there are also a sizable number of people who like "e-mail" in some way, and it's the reason they are interested in Delta. 

However, the page does focus on **"chatmail" instead of "e-mail"**, with the first link going to https://delta.chat/en/chatmail which explains clearly how it is realted to e-mail.  Betting on chatmail makes sense: it has changed the situation (with one of our most re-tooted annoucenments ever) as the question of "which e-mail server should i use" is largely gone and you all know how much efforts we spent on supporting providers, doing provider comparisons, fixing corner cases, supporting outlook etc.  Moreover, with chatmail we can do push notifications and many other user high-prio wishes that we knew about for years.  By integrating chatmail onboarding as the new default, users do not need to know it is e-mail.  I am aware this is a a departure from how we talked about things in the last 3-5 years.  But I'd say that "Internet standards" is a better term anyway for what we are doing and trying to communicate, as it also encompasses webxdc, a pretty unique effort in the messenger space. See also some user-testing in the fediverse here: https://chaos.social/deck/@delta/112422007157098940 :) 

By focusing on **Internet standards** and providing a funny dogwhistle with xkcd927, we are making a major point for those who understand its problematics, Matrix and other efforts. Nerdy? Maybe but not in a bad way as you can click it and read the comic :) The rest of the points try to convey key differences meaningful for the target group. 

Hope this clarifies things a bit? Happy to continue the discussion if you see the need but then maybe better in person in a few days then.

en/index.md
hpk42: changed it to match the new title. see below.

kermoshina:
> Few mainstream users will see and study this web page that doesn't even have a video. They rather go to the store pages



I agree! 

 > If we want to focus on mainstream users with our communication we need to do and instigate videos.

Totally! Happy to help with this.

> **As to "E-mail"**, the lack of focusing on it is intentional and tied to the 1.46 release goal: you should not need to know or care that you are using e-mail in order to succeed. 

Ok, makes sense!

> 
> However, the page does focus on **"chatmail" instead of "e-mail"**, with the first link going to https://delta.chat/en/chatmail which explains clearly how it is realted to e-mail. Betting on chatmail makes sense: it has changed the situation (with one of our most re-tooted annoucenments ever) as the question of "which e-mail server should i use" is largely gone

True!


> But I'd say that "Internet standards" is a better term anyway for what we are doing and trying to communicate, as it also encompasses webxdc, a pretty unique effort in the messenger space. See also some user-testing in the fediverse here: https://chaos.social/deck/@delta/112422007157098940 :)

Ok! I am happy with using Internet standards as a term 


> By focusing on **Internet standards** and providing a funny dogwhistle with xkcd927, we are making a major point for those who understand its problematics, Matrix and other efforts

Makes sense
>

en/index.md
```diff
@@ -1,15 +1,25 @@
...
 
-💬 Message anyone with an e-mail address [even if they don't use Delta Chat](https://www.youtube-nocookie.com/embed/8LbrGXKZN70).
+<br>
 
-🥳 Enjoy interactive chat experiences through [webxdc apps](https://webxdc.org).
+<span style="font-size: 1.5em">
+    💬 Reliable instant messaging with
+       multi-profile and multi-device support
+   <br/><br/>
+   💨 No number, e-mail or data needed to use interoperable [chatmail servers](https://delta.chat/en/chatmail)
```
missytake:
```suggestion
   💨 No phone number or e-mail needed to sign up on [chatmail servers](https://delta.chat/en/chatmail)
```

Let's make clear that it is about sign-up. And I think we can leave out "or data" here, people are used to lying with this anyway; email and phone number are the barriers.

And let's talk about interoperability below, in an "e-mail" point.

missytake, line 14:
```suggestion
   💨 No phone number or e-mail needed to sign up on [chatmail servers](https://delta.chat/en/chatmail)
```

Let's make clear that it is about sign-up. And I think we can leave out "or data" here, people are used to lying with this anyway; email and phone number are the barriers.

And let's talk about interoperability below, in an "e-mail" point.

en/index.md
```diff
@@ -1,15 +1,25 @@
...
+   <br/><br/>
+   🥳 Interactive [web apps in chats](https://webxdc.org/) for gaming and collaboration
+   <br/><br/>
+   🔒 [Audited end-to-end encryption](https://delta.chat/en/2024-03-25-crypto-analysis-securejoin) safe against network & server attacks
+   <br/><br/>
+   👉 [F](https://en.wikipedia.org/wiki/Free_software)[OSS](https://en.wikipedia.org/wiki/Open-source_software) software, built on [Internet Standards](https://github.com/deltachat/deltachat-core-rust/blob/master/standards.md), avoiding [xkcd927](https://xkcd.com/927/) :)
```
missytake, line 20:
```suggestion
   👉 [F](https://en.wikipedia.org/wiki/Free_software)[OSS](https://en.wikipedia.org/wiki/Open-source_software) software, built on e-mail interoperability, avoiding [xkcd 927](https://xkcd.com/927/) :)
```

This makes it more clear what exactly our strength is in this regard. I propose linking to https://staging.delta.chat/867/en/help#does-delta-chat-work-with-my-e-mail-provider or https://staging.delta.chat/867/en/help#why-do-i-have-to-enter-my-e-mail-password-into-delta-chat-is-this-secure, and explain there that you can use any e-mail account to chat.

missytake (commented):
> **Target group:** intended are multipliers, developers, might-become contributors but also more deeply interested users, who know properties of other messengers. 

> By focusing on **Internet standards** and providing a funny dogwhistle with xkcd927, we are making a major point for those who understand its problematics, Matrix and other efforts. Nerdy? Maybe but not in a bad way as you can click it and read the comic :) The rest of the points try to convey key differences meaningful for the target group.

I'd argue that it isn't understood by the whole target group, why standards are important to combat the messenger problematic. What standards are and how they are used is mostly clear to developers, who actually look at them to implement them. Others might know about interoperability, maybe they heard of the term network effects. But I don't think it is clear enough to warrant a pole position on the front page. Where the term is right now, as part of point 5, is fine though.

> However, the page does focus on **"chatmail" instead of "e-mail"**, with the first link going to https://delta.chat/en/chatmail which explains clearly how it is realted to e-mail. Betting on chatmail makes sense: it has changed the situation (with one of our most re-tooted annoucenments ever) as the question of "which e-mail server should i use" is largely gone and you all know how much efforts we spent on supporting providers, doing provider comparisons, fixing corner cases, supporting outlook etc. Moreover, with chatmail we can do push notifications and many other user high-prio wishes that we knew about for years. By integrating chatmail onboarding as the new default, users do not need to know it is e-mail. I am aware this is a a departure from how we talked about things in the last 3-5 years. But I'd say that "Internet standards" is a better term anyway for what we are doing and trying to communicate, as it also encompasses webxdc, a pretty unique effort in the messenger space. See also some user-testing in the fediverse here: https://chaos.social/deck/@delta/112422007157098940 :)

I would not hide the fact that you can still use almost any e-mail account to use Delta Chat. For many people and organizations this is still be the best way to use Delta Chat. If we hide that information, users might be surprised why it isn't recommended anymore. But the "classic e-mail" use case didn't become worse. And actually if e-mail interoperability is important to someone, then using a chatmail server is a worse experience (you won't be able to respond to classic e-mails).

The classic e-mail use case might not be the main use case we want in the future; but we should still mention that it is _possible_. It is our largest interoperability advantage after all.

en/index.md
r10s: i see the point, but i'd anyway tune down email here. for reasons @hpk42 pointed out above, eg. _"Better not think first about using your existing e-mail account because that requires a lot of understanding and there are many known issues"_ is very strong

i do not think, that this will confuse existing users - but even if: the whole change is about users we do not have currently, which are the far larger base, hopefully.

and we do not hide "classic email" completely - it is still in the help and at other places, also on onboarding, only two taps away. but it is reasonable to not explicitly mention it on the homepage.

i just give this change a try - we played the "just use classic email" card quite some while :)

en/index.md
hpk42: now changed to "is a decentralized private messenger app"

hpk42:
Nami, I sympathize with that you are skeptical of dropping "e-mail" alltogether at this point. And that you are skeptical that "Standards-based" is a good rallying title.  Here is my next attempt to address these concerns. 

image

missytake (approved):
Yep, that's even better than my suggestion. "or use traditional e-mail servers" is clearly showing that chatmail is the cool new thing, but the other is still possible. It also fits the sentiment/expectation most viewers will have that e-mail is traditional. I like.

en/index.md
missytake: "Aim for new users, not for current users" is a valid point; but note that it is [not without problems](https://nothinghuman.substack.com/p/the-tyranny-of-the-marginal-user).

Our current userbase is also connected with future users; they will most likely keep recommending delta as "the email messenger" for a while. If email isn't even mentioned on the front page, people will wonder if they are at the right address.

hpk's most recent proposal is the best compromise between both concerns I think. As the second option in "or use traditional email servers", users don't *have to think* about their current accounts, but we show that they *can*.

Basically, I think we should provide an entry point for the whole span of considerations touching "wow, it runs on e-mail? that means I can use all the advantages of the e-mail network like wide interoperability, spam protection, sieve rules, BCC, aliases, and all of our members already have an account via SSO?"
Because Meera and Stefanie from https://github.com/deltachat/bg-deltachat/blob/master/ux-feedback/personas.md would think about this for longer, but if it isn't on the front page, they wouldn't dive into the FAQ I think.

r10s (approved):
even nicer now! and indeed nice wrt "new" vs "traditional"

gerryfrancis:
Nice work, guys! :) The only thing I have stumbled upon is how we would like to write `and`; either `&` or `and`. I do not prefer one writing over the other, but I think it should be consistent.

hpk42:
thanks all, it's merged now.

PR #870, opened 2024-05-13
cleanup links

r10s:
- move all links to a single section

- move 'system requirements' up (would be best to have them in the respective sections, but this is already an improvement)

- 'FAQ multiclient' - this is probably not one of the first questions, so it seems okay to have that only in the normal FAQ; also, this comes from a time where our multiclient story was much worse, meanwhile, this is guided in the app as well

- add link to the new 'Alternative Clients' overview (which may be moved from forum to delta.chat)


hpk42 (approved):
nice improvement to put all links to extra things into one section,  reduces the clutter on the download page

PR #872, opened 2024-05-14
use unagitated, more professional looking list

r10s:
while on making the page [easier translatable](https://github.com/deltachat/deltachat-pages/pull/867#pullrequestreview-2051412930)
by keep using markdown and increase the fontsize by some css magic, it turns out that the simple thing looks better.

the increased fontsize
make headlines less outstanding and turns the whole page a bit into a desert of text without structure.
also, it does not look good compared the rest of the site.

i played around with some background or a box, bold text etc. - however, there is not much on the page,
why adding another level of attracting attention.
but i removed the second headline, moving it completely to "Download" makes it even more outstanding, without cluttering.

so it seems better to stay with the simple thing
and better concentrate making the screenshots nice.

my 2ct :)

EDIT: still using slightly larger font, by the moved second headline this still looks good. also, things are driven by CSS now, so we can tweak that without forcing retranslations

### DIFF
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -84,6 +84,12 @@ body {
 	width: 100%;
 }
 
+/* larger font on homepage, may be removed or adapted whenever homepage changes */
+#homepage-heading ~ p {
+  font-size: 1.2em;
+  margin-top: 1.6em !important;
+  margin-bottom: 1.6em !important;
+}
 
 /* rough layout: desktop second 
 *******************************************************************************/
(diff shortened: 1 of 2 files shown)


hpk42 (approved):
i am fine with it -- i generally prefer larger fonts but if you say so :)

r10s:
i added a commit that makes the font larger again, without having everything the same

r10s:
playing around, i think, it is more the second headline than the font size, that make things looking cluttered, i am not yet happy with this PR ...

PR #882, opened 2024-05-22
update screenshot

### LINKED ISSUE #868: New screenshots for home page (and later stores), write some replication machinery to ease that
When releasing 1.46 apps let's do some up-to-date screenshots, first on the home page. 

Let's try to have some interesting conversations and snapshot the accounts, and try to be able to re-generate sscreenshots and the contained dialogues (which should not just be about office lunch, or some party, but about some real drama like the alice/Bob/Mad-hatter dialogues (and there originally also was carol who started an affair with alice or something?) 

maybe we can write a python script that actually creates Alice, Bob, Mad-hatter accounts, sets avatars, sends messages etc. and then provides a number of accounts in a zip file so that one can just import it into Android, ios, desktop and do some snapshots. 
We can evolve the dialogues over time, so something that replicates what we have.  cc @missytake @adbenitez @kermoshina -- could be a fun thing to go for next week at bb :)

r10s:
closes #868

did this PR together with @adbenitez 

we can always iterate over create more fancy things, however, the screenshots are barely readable without having enlarged them, so not so sure about a "story", as this might not show things as reactions or gallery. but i still like the idea if someone picks that up, same for a generation script, maybe it introduces the overall effort

there is also the idea of an carousel or so, still some place for optimisation

before / after (the desktop screenshot was recently updated already):

Screenshot 2024-05-23 at 13 02 43
Screenshot 2024-05-23 at 13 02 49


missytake (approved):
Looks much better already!

PR #890, opened 2024-05-25
FAQ: invite links make everything easier now

### LINKED ISSUE #862: FAQ on invite link is missing?
seems we don't have much about invite links and only talk about QR code scanning in the FAQ. 
Should change around 1.46 -- the feature is already in 1.44.

missytake:
fix #862 

Now that we have invite links, the QR code flow is less important. It's mostly useful for chat partners who just met and don't have a second channel yet.

We also don't really need the "joiner side" anymore. If they are together in person, they already know where the QR code menu is. If it's a link, they do not even need the QR menu.

Unfortunately the invite QR codes are still `OPENPGP4FPR` codes, so if the joiner doesn't have Delta Chat installed, scanning them with any app will not lead to https://i.delta.chat. This is the only thing which makes me hesitate whether we can already delete the whole "Joiner Side" paragraph. But as long as they have Delta Chat installed at all, even a normal scanner app _should_ open the app. The FAQ still implies that the joiner needs to have Delta Chat installed.

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -402,16 +402,17 @@ you can generate an invite link.
 If you are together in person,
 you can show a QR code to your chat partner.
 
-1. For **Group invitations**,
-   tap the chat group title to see its member list,
-   and select "QR Invite code".
-2. For **direct 1:1 chat invitations**,
-   tap the QR Code icon <img style="vertical-align:middle; width:1.8em; margin:1px" src="../assets/help/qr-icon.png" />
-   on the Delta Chat app main screen.
+- For **Group invitations**,
+  tap the chat group title to see its member list,
+  and select "QR Invite code".
+
+- For **direct 1:1 chat invitations**,
+  tap the QR Code icon <img style="vertical-align:middle; width:1.8em; margin:1px" src="../assets/help/qr-icon.png" />
+  on the Delta Chat app main screen.
 
 Let your chat partner scan the QR image
 with their Delta Chat app,
-or click "Copy" to create an invite link
+or click "Copy" or "Share" to create an invite link
 and share it with your chat partner.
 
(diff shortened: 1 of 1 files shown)


hpk42 (approved):
it's better than before even though i am not sure it's good to remove the clear visual "Inviter side" and "Joiner side" separation.

en/help.md
```diff
@@ -395,40 +395,26 @@ End-to-end encryption is guaranteed if there is a green checkmark next to the ch
...
+1. For **Group invitations**,
+   tap the chat group title to see its member list,
+   and select "QR Invite code".
+2. For **direct 1:1 chat invitations**,
+   tap the QR Code icon <img style="vertical-align:middle; width:1.8em; margin:1px" src="../assets/help/qr-icon.png" />
+   on the Delta Chat app main screen.
+
+Let your chat partner scan the QR image
+with their Delta Chat app,
+or click "Copy" to create an invite link
```
```suggestion
or click "Copy" or "Share" to create an invite link
```

en/help.md
```diff
@@ -395,40 +395,26 @@ End-to-end encryption is guaranteed if there is a green checkmark next to the ch
...
+like a video chat or a different messenger,
+you can generate an invite link.
+
+If you are together in person,
+you can show a QR code to your chat partner.
+
+1. For **Group invitations**,
+   tap the chat group title to see its member list,
+   and select "QR Invite code".
+2. For **direct 1:1 chat invitations**,
```
r10s: it is better to add an empty line between; future changes will result in less noise wrt retranslations.

maybe also a unordered translations is better, as it is not "two steps" but "either or"

```suggestion
   and select "QR Invite code".

2. For **direct 1:1 chat invitations**,
```

PR #891, opened 2024-05-25
FAQ: only talk about accounts when it's about interacting with e-mail

### LINKED ISSUE #887: FAQ: talk about profiles consistently
We still talk about accounts in the FAQ; we need to revisit where it makes sense and where we need to change it.

missytake:
fix #887 

I looked through our FAQ where we still need the "account" wording and where we can replace it with "profile".

In all cases where it's about how the e-mail account interacts with the wider e-mail ecosystem, e.g. because of using the same account with non-Autocrypt clients, deleting the e-mail account, e-mail accounts of chat partners which might not use Delta Chat, deletion on the server, etc. I kept using the `e-mail account` wording.

In the below cases I found that the `profile` wording even improves the FAQ:

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -495,7 +495,7 @@ even if the contact is a member there.
 **Your contact is using Delta Chat on a second device (phone or laptop)**
 
 If they have another device with a Delta Chat app running,
-they should remove the account from the new device
+they should remove the profile from the new device
 and add it [as a second device as described here](#multiclient).
 As soon as they message you afterwards, the warning will be gone
 and guaranteed encryption is established with both devices of your contact. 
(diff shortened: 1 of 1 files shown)


Simon-Laux (approved):

en/help.md
```diff
@@ -654,12 +654,12 @@ deliver messages to recipient's devices.
 
 Both for protecting against metadata-collecting e-mail servers 
 as well as against the threat of device seizure
-we recommend to use a Delta Chat optimized [e-mail server instance](https://delta.chat/serverguide)
-to create pseudonymous temporary accounts through QR-code scans.
-Note that Delta Chat apps on all platforms support multiple accounts 
-so you can easily use action-specific "1-week" or "1-month" accounts next to your "main" account
-with the knowledge that all temporary account data, along with all metadata, will be deleted.
-Moreover, if a device is seized then contacts using temporary e-mail accounts
+we recommend to use a [chatmail server](chatmail)
```
r10s: as "serverguide" etc, this needs to be an absolute link as the destination is not part of the offline help later

```suggestion
we recommend to use a [chatmail server](https://delta.chat/chatmail)
```

Issue #902, opened 2024-06-03
Jekyll does not start on Arch Linux


link2xt:
There is no package for jekyll on Arch and jekyll is part of gemfile anyway, so "[install Jekyll](https://jekyllrb.com/docs/installation/) on your local machive" before `bundle install` should not be needed anyway.

After running `bundle install` I get this error when trying to run jekyll:
```
$ bundle exec jekyll
bundler: failed to load command: jekyll (/home/user/src/deltachat/deltachat-pages/vendor/bundle/ruby/3.0.0/bin/jekyll)
/home/user/src/deltachat/deltachat-pages/vendor/bundle/ruby/3.0.0/gems/jekyll-4.3.2/lib/jekyll.rb:29:in `require': cannot load such file -- json (LoadError)
	from /home/user/src/deltachat/deltachat-pages/vendor/bundle/ruby/3.0.0/gems/jekyll-4.3.2/lib/jekyll.rb:29:in `<top (required)>'
	from /home/user/src/deltachat/deltachat-pages/vendor/bundle/ruby/3.0.0/gems/jekyll-4.3.2/exe/jekyll:8:in `require'
	from /home/user/src/deltachat/deltachat-pages/vendor/bundle/ruby/3.0.0/gems/jekyll-4.3.2/exe/jekyll:8:in `<top (required)>'
	from /home/user/src/deltachat/deltachat-pages/vendor/bundle/ruby/3.0.0/bin/jekyll:25:in `load'
	from /home/user/src/deltachat/deltachat-pages/vendor/bundle/ruby/3.0.0/bin/jekyll:25:in `<top (required)>'
	from /home/user/src/deltachat/deltachat-pages/vendor/bundle/ruby/3.0.0/gems/bundler-2.4.13/lib/bundler/cli/exec.rb:58:in `load'
	from /home/user/src/deltachat/deltachat-pages/vendor/bundle/ruby/3.0.0/gems/bundler-2.4.13/lib/bundler/cli/exec.rb:58:in `kernel_load'
	from /home/user/src/deltachat/deltachat-pages/vendor/bundle/ruby/3.0.0/gems/bundler-2.4.13/lib/bundler/cli/exec.rb:23:in `run'
	from /home/user/src/deltachat/deltachat-pages/vendor/bundle/ruby/3.0.0/gems/bundler-2.4.13/lib/bundler/cli.rb:492:in `exec'
[...]

PR #903, opened 2024-06-03
Reapply "Build website with nix"

link2xt:


missytake (approved):
Looks good. I think we can close #902 as well with this?

hagenest (approved):
still bullish about nix, although I'm not 100% what it does better in this specific case :D

link2xt:
#902 is a bit independent, I can use nix instead but the issue is that readme tells you to install jekyll but then does not actually use installed jekyll because bundler installs another copy.

link2xt:
> still bullish about nix, although I'm not 100% what it does better in this specific case :D

I don't really care if nix is used by CI, but want to be able to run `nix build` to get a website copy locally.

PR #909, opened 2024-06-12
download: remove onboarding steps from download page

### LINKED ISSUE #889: remove "step 2" from download page
the step of scanning an account QR code is optional when 1.46 is out on all platforms; the onboarding process also guides to alternative logins

missytake:
closes #889


r10s:
- i would preserve the paragraph "Enter a Username and Meet Your Friends" for now. it is probably one of the larger challenges after getting an account is now dead-simple.  
however, the step should be "❷" then :)

- unused strings should be removed from the .yml

in general, let's wait with merging until 1.46 is out everywhere.

missytake:
> * i would preserve the paragraph "Enter a Username and Meet Your Friends" for now. it is probably one of the larger challenges after getting an account is now dead-simple.
>       however, the step should be "❷" then :)

Setting a profile is already guided through by the app now. Finding contacts can be a challenge, but will a random sentence on the download page make that easier? Now that users directly get an account, there is no reason for them to look for next steps on the download page.

r10s:
> Setting a profile is already guided through by the app now. 

true, however, repetition is usually working.

>  Now that users directly get an account, there is no reason for them to look for next steps on the download page.

that point is more convincing. indeed, the whole "step" thingie is no longer needed.

most ppl also download via stores - and there, the added screenshot showing scanning maybe even has a larger impact.

so, i am fine with removing it. if it turns out, ppl have serious issues to get in contact, we can iterate, but then more in the app probably.

nb: i added a commit that adds some padding as otherwise "requirements" and "more options" are too close to each other (i hope that is syntactically fine, i am not deep into scss

r10s (approved):

PR #910, opened 2024-06-12
download: we have an official arch package now

### LINKED ISSUE #908: add official arch package
link and the command to install it
https://archlinux.org/packages/extra/any/deltachat-desktop/

missytake:
closes #908

@Jikstra or is there still a reason to keep the AUR package there if there is an official package now? I don't use arch, no idea.


missytake:
@dvzrv like this?

Jikstra:
> closes #908
> 
> @Jikstra or is there still a reason to keep the AUR package there if there is an official package now? I don't use arch, no idea.

I think AUR can make sense to keep around, its basically a dev channel. But can also go awayy

Simon-Laux (approved):

PR #911, opened 2024-06-13
update translations

r10s:
this mainly pulls in the russian update, driven by @guland2000


r10s:
super-nice also for the help, that goes to the in-app offline help as well

PR #913, opened 2024-06-17
bump android to 1.46.5

r10s:


missytake (approved):
no-brainer :) is it ready already?

r10s:
yeah, it was tested a bit internally and currently building on f-droid, let's merge. for gplay, i would wait some more days to be more sure there are no worsenings because of the new core

PR #914, opened 2024-06-24
Change privacy-minimizing to privacy-preserving in FAQ

hagenest:
Just a small change, we also have to update translations. I guess this is only possible after merging this PR? Haven'T worked with transiflex before.


missytake (approved):

missytake:
Translations are documented here: https://github.com/deltachat/deltachat-pages/?tab=readme-ov-file#translate-the-website
But it will happen through our normal website updating workflows, no need to worry

PR #922, opened 2024-07-08
bump android to 1.46.7

r10s:


r10s:
merging, gplay is already rolling as well

PR #923, opened 2024-07-16
Improve download page

missytake:
Not sure if the bullet points are worth the reformatting... but I tend to say yes. As it is right now it is basically illegible, just a long sentence with a lot of numbers, nobody reads that.

### DIFF
--- a/en/download.md
+++ b/en/download.md
@@ -5,10 +5,11 @@ downloads: true
 ---
 
 Minimal requirements:
-Android 4.1 Jelly Bean
-or iOS 12, iPhone 5s or iPad 5/Air/Mini 2
-or Windows 10, macOS 10.15 Catalina, Ubuntu 18.04, Fedora 29 or Debian 10
-or compatible systems.
+- Android 4.1 Jelly Bean
+- iOS 12, iPhone 5s or iPad 5/Air/Mini 2
+- Windows 10
+- macOS 10.15 Catalina
+- Ubuntu 18.04, Fedora 29 or Debian 10 or compatible systems.
 
 The desktop versions do not require Delta Chat to be installed on a phone.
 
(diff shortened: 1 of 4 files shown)


ralphtheninja (approved):

en/download.md
```diff
@@ -5,10 +5,11 @@ downloads: true
...
 Minimal requirements:
-Android 4.1 Jelly Bean
-or iOS 12, iPhone 5s or iPad 5/Air/Mini 2
-or Windows 10, macOS 10.15 Catalina, Ubuntu 18.04, Fedora 29 or Debian 10
-or compatible systems.
+- Android 4.1 Jelly Bean
+- iOS 12, iPhone 5s or iPad 5/Air/Mini 2
+- Windows 10
+- macOS 10.15 Catalina
+- Ubuntu 18.04, Fedora 29 or Debian 10 or compatible systems.
```
r10s:
```suggestion
- Ubuntu 18.04, Fedora 29, Debian 10 or compatible systems.
```

r10s (approved):
yeah, this is an improvement in readability :)

i was also wondering from time to time about that paragraph, but never took action - my idea was more about moving the requirements to the belonging "os boxes", however. but keep the things together is of some worth as well

missytake:
> my idea was more about moving the requirements to the belonging "os boxes", 

point, this is much better. Let me come up with something.

r10s:
pretty nice already!

- the "Delta Chat doesn't have to be installed on a phone." feels a bit out of context, however.  
maybe better leave that as a general hint at the old position, this also clutters the boxes a bit less - and the hint is valuable in general, also if you're up to download eg. "android" at the moment.

- for uncluttering the boxes: it will look better when having "other download options" and "source code" below each other as before - but in am fine with the current ordering as well (it is maybe a  little more logical)

PR #926, opened 2024-07-22
make 'system requirements' less language dependent

r10s:
avoid the english word 'or' in requirements and make them less language dependent (the OS names are usually not translated anyways and subject to change often, so it makes sense to have that part more independent)

successor of #923

### DIFF
--- a/_includes/download-boxes.html
+++ b/_includes/download-boxes.html
@@ -85,7 +85,7 @@
             </div>
             <div>
                 {%if tx.download.minimum != ""%}{{ tx.download.minimum }}{%else%}{{ txEn.download.minimum }}{%endif%}:
-                iOS 12, iPhone 5s or iPad 5/Air/Mini 2
+                iOS 12, iPhone 5s, iPad 5/Air/Mini 2
             </div>
             <div class="descr">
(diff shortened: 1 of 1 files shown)


_includes/download-boxes.html
```diff
@@ -167,7 +167,7 @@ <h1>{%if tx.menu.download != ""%}{{ tx.menu.download }}{%else%}{{ txEn.menu.down
             </div>
             <div>
                 {%if tx.download.minimum != ""%}{{ tx.download.minimum }}{%else%}{{ txEn.download.minimum }}{%endif%}:
-                Ubuntu 18.04, Fedora 29, Debian 10 or compatible systems
+                Ubuntu 18.04, Fedora 29, Debian 10, Compatible Systems
```
r10s, line 170: i am wondering if we shouldn't be a bit more specific here.

eg. there are "Arch" and "Nix" downloads as well - don't they have a minimal version? cc @missytake, @Simon-Laux, @nicodh

missytake (approved):

_includes/download-boxes.html
missytake: Arch and Nix are rolling releases if I understood it correctly, they don't really *have* outdated versions in the wild.

PR #931, opened 2024-07-28
Fix broken links to docker/docker-compose installation (in the server guide)

ralphtheninja:
Links to installing `docker` and `docker-compose` are broken and yield a 404 (also in german for some reason)

image

### DIFF
--- a/ca/serverguide.md
+++ b/ca/serverguide.md
@@ -26,7 +26,7 @@ What you need:
 ## Installing Docker
 
 As a prerequisite you need to install [docker and
-docker-compose](https://docs.mailcow.email/i_u_m/i_u_m_install/).
+docker-compose](https://docs.mailcow.email/getstarted/install/#docker-and-docker-compose-installation).
 
 ### If docker.com is Blocked:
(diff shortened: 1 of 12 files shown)


missytake (approved):
Usually our transifex workflow (a non-CI workflow, lol) takes care of porting the fixes to other languages... but for the languages which never have been translated, like in this PR, I think it doesn't make a difference.

ralphtheninja:
> Usually our transifex workflow (a non-CI workflow, lol) takes care of porting the fixes to other languages... but for the languages which never have been translated, like in this PR, I think it doesn't make a difference.

@missytake Ok. So I should only update the `en`-part next time?

missytake:
> > Usually our transifex workflow (a non-CI workflow, lol) takes care of porting the fixes to other languages... but for the languages which never have been translated, like in this PR, I think it doesn't make a difference.
> 
> @missytake Ok. So I should only update the `en`-part next time?

yes :) the rest happens automatically in subsequent PRs @r10s usually does

PR #940, opened 2024-08-10
bump android to 1.46.10

r10s:


r10s:
gplay starts rolling again; as  1.46.10 uses same core and has only few changes compared to 1.46.8, we'll probably rollout  1.46.10 a bit faster

PR #958, opened 2024-09-18
fix link to chatmail blogpost

### LINKED ISSUE #957: Redirection problems in lieu of localization
[Reported on the fediverse](https://social.anoxinon.de/@Lioh/113157780249093805) 

It is probably related to localization and redirects, a kind of known issue that there is some mess on our nginx deployment. 
Should be reproducable with this link: https://delta.chat/2023-12-13-chatmail  and using a non-english locale. 

image

missytake:
fix #957


r10s (approved):
you're faster, i just wanted to do the same :)

PR #959, opened 2024-09-19
Collect most recent iOS releases in a document

zeitschlag:
- Idea war to start _somewhere_. This document gives a simple overview which app-release uses which Core-version.
- I opted for a table in this case as it made sense to me, but I'm also happy to convert it into a list with some headlines in between. Just HMI!
- Feel free to add your releases on other platforms :-)

### DIFF
--- a/releases.md
+++ b/releases.md
@@ -1,0 +1,16 @@
+# Delta Chat Releases
+
+## iOS
+
+| Release                  | Core Version                   | Remarks            |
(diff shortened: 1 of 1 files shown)


missytake (approved):
Great :) maybe a column whether it is public in appstore already? For other platforms it probably makes sense to add all the various appstores separately. Would be cool if we could auto-populate it somehow

PR #971, opened 2024-10-21
Add RELEASES.md, fill in android releases

Hocuri:


r10s:
there is already https://github.com/deltachat/deltachat-pages/blob/main/releases.md - i think, the idea was to have it all at one glance?

Hocuri:
I didn't know about this, will move it. Yes, the idea was to have it at a glance.

r10s (approved):
lgtm, the main usefulness will at the end come from the file being updated (the easier format and the less effort, the better)

there is also amazon, huawei and this site - but no need to add that to this PR

ralphtheninja:
`git rm RELEASES.md`?

Hocuri:
Thanks for catching, fixed: https://github.com/deltachat/deltachat-pages/pull/973

PR #979, opened 2024-11-03
fix docker command in readme

Simon-Laux:


r10s (approved):
bigger numbers always looks better 💪

PR #970, opened 2024-10-17
Deploy LibResilient with GitHub Pages as a backend

link2xt:

PR #989, opened 2024-11-07
show Github download option

r10s:
this is useful, in case delta.chat is unavailable for whatever reason.

with [LibResilient](https://github.com/deltachat/deltachat-pages/pull/970) it is possible that the blocked page is still openable. but even if not, it is useful to show the option,
so that ppl get to know this option.

(we're not linking to the concrete release version, however, as that may be out of sync with the pages-version - and we do not want to add additional maintainance effort here)


link2xt (approved):

link2xt:
delta.chat is just frequently down due to some routing problems on the hoster as visible in our alerts (connection to prometheus goes down from time to time). In this case download.delta.chat is not accessible as we don't mirror it, so GitHub is useful in this case.

Google Play and F-Droid are normally up and accessible anyway, so GitHub is an alternative to download.delta.chat rather than stores.

r10s:
maybe `download.delta.chat` as a shortcut often used and propagated is not that useful at the end? as subdomains are tricky?
maybe we should better start propagating sth. as `delta.chat/dl` (even shorter, so there is some incentive to use that :) - that then has a chance to get cached by LibResilient?

link2xt:
Everything under https://delta.chat/ gets downloaded from https://deltachat.github.io/deltachat-pages/ if main server is not accessible. So as long GitHub is up and user previously visited delta.chat, website is somewhat usable.

Other subdomains are not mirrored anywhere and service worker is installed only for https://delta.chat/ anyway, so not using subdomains is more reliable indeed.

missytake:
> delta.chat is just frequently down due to some routing problems on the hoster as visible in our alerts (connection to prometheus goes down from time to time). 

I disagree that the hoster is the problem here or the website is down; I think it's just wireguard. more details in https://github.com/deltachat/sysadmin/issues/195

Issue #848, opened 2024-05-02
Infinite redirects happen for non existing pages


nicodh:
go to https://delta.chat/foobar and you will be redirected to https://delta.chat/de/de/de/foobar etc.

same for https://delta.chat/chatmail/ instead of https://delta.chat/chatmail 

previous discussion: #618

r10s:
@missytake iirc, we tried fixing that some years ago already, without success, or we decided that there were larger fishes to get. but i agree, it is pretty annoying, i also come over that regularly, would be great if you can have a second look, sth. is wrong with our redirects

nicodh:
On which web server are the pages hosted? The redirects have to be configured on that server I guess...

r10s:
hm, the README points wrt language-redirects to https://github.com/deltachat/sysadmin/blob/master/page/delta.chat/delta.chat - but that seems outdated - @missytake will know better :)

r10s:
clicking around, configuration seems to be at https://github.com/deltachat/sysadmin/blob/master/page/config/nginx/delta.chat now

r10s:
just got another complain:

> grad durch zufall gelernt: `https://delta.chat/enchatmail` leitet weiter zu `https://delta.chat/de/de/de/de/de/de/de/de/de/de/de/de/de/de/de/de/de/de/de/de/enchatmail`
> 
> das sieht für mich nich so ganz richtig aus, soll das so? wer ist dafür verantwortlich?


this is somehow different meanwhile - even if we agree that it is not worse, it is a bad impression and it always costs time to argue that it is not that bad ...

missytake:
There is this test script for the redirects: https://github.com/deltachat/sysadmin/blob/master/page/test-redirects.py

r10s:
closed by #990

Issue #618, opened 2023-02-20
language redirect errors


r10s:
i think, there is somehow an `en` to much in redirects, in the past we had recursive redirects as `en/en/en/en/en`, that is stopped with a hack meanwhile, but https://delta.chat/se/ redirecting to `https://delta.chat/en/en/se/` is still wrong.

the logic seems to live at https://github.com/deltachat/sysadmin/blob/beb0307a873529e9e739f951ce8df672f1a088fe/delta.chat/delta.chat#L81 - on testing, one should be aware of browser caches

iirc, there were other issues filed before but i cannot find them somehow

ralphtheninja:
Ah yes, I forgot to create the issue I promised ya'll. We were trying to install delta chat on iOS on two different devices. Both had problems getting to the website because nginx returned 403 errors. This is really a shame, since getting to a webpage shouldn't be hindered by the server itself ;)

Anyway. My suggestion would be to remove all that weird redirect complexity (or fix it) and have all requests land on `https://delta.chat/en` by default and instead add a language picker to the page so the person can pick their language of choice themselves. The server can assume less and just return what's asked for, like a server should and we can go about our day.

gerryfrancis:
> iirc, there were other issues filed before but i cannot find them somehow

This one is still open: https://github.com/deltachat/deltachat-pages/issues/541

Also maybe check the right order of `RewriteCond` and `RewriteRule` (if they exist) in the `.HTACCESS` file. (It is just a guess.)

r10s:
> Also maybe check the right order of RewriteCond and RewriteRule (if they exist) in the .HTACCESS file. (It is just a guess.)

i think, this is mostly related to apache - where delta.chat is using nginx

the logic is linked above, and i think, that should be simplified. quite some of the lines are hacks to stop endless redirects - if we fix the underlying issue, that can go away. the concatenation of en..request_uri looks suspicious to me, as the old language path is not stripped. i think, the whole language-logic should operate with the last path element only, and/or only when there is no path set at all (so, only when delta.chat is opened)

however, this needs to be tried out by someone with access to a server, it does not make too much sense to judge without having access to the code and being able to try things out :)

gerryfrancis:
> i think, the whole language-logic should operate with the last path element only, and/or only when there is no path set at all (so, only when delta.chat is opened)

@r10s The last path element can also be the name of a site, e.g. https://delta.chat/de/blog .

missytake:
[...]

PR #990, opened 2024-11-08
client-side redirects

r10s:
server-side redirects cause a lot of surprisingly hard to fix issues - as the years-old issue with the `/en/en/en/en/...` redirects (#848, #618 ...), but also `LibResilient` has [issues](https://gitlab.com/rysiekpl/libresilient/-/issues/85) with them.

surely, this is all solvable,
however, another solution are client-side redirects.

this means, to have a little html-file for each redirect,
the file contains the destination (relatively!) as HTML, and also executes a little javascript that does language-corrected redirects, if possible.

nice side-effect is that they can also be created and modified from this repo, by pages authors, eg. the example also works in this PR's page preview

drawbacks are:
- when javascript is disabled, redirecty are in english (one still can change language in the footer)
- while search engined follow client-side-redirects, they may rank pages using server-side redirects better.
- EDIT: catch-all redirects are not working - but can also not create issues :)

the drawbacks seem not to outweight the issues before nor affect a significant number of users or have big imact.

btw, there is already `index.html`, that uses the client-side approach, not sure if that is really in use, however.

if we want to go for client-side-redirects for existing redirects, we need to change server-side redirects at the same time; this can be done in subsequent PRs, however


link2xt:
index.html is used when primary site is down and user types https://delta.chat/ into the browser. If primary server is down there are no HTTP redirects.

dl.html
```diff
@@ -0,0 +1,3 @@
+<!DOCTYPE html>
+<meta http-equiv="refresh" content="1; url=en/download">
```
missytake:
```suggestion
<meta http-equiv="refresh" content="1; url=https://delta.chat/en/download">
```

Can we use an absolute URL here? it kinda breaks the preview, but e.g. lynx (non-js command line browser) only uses these refresh URLs for absolute URLs.

missytake (approved):
I'm not sure if worse SEO (maybe needs to be tested before/after with a tool) really justifies replacing the slightly buggy server-side redirects, as they are not buggy with *normal daily* usage (clicking on links), but only with developer usage (manually typing URLs).

 I clearly see the advantage of both closing #848 and moving these redirects into the repository, so I won't block it.

dl.html
r10s: it is not only about the preview, it is also about having redirects that also works, when eg. used from another domain, another subdirectory etc. - when the original domain, used in the path, is eg. blocked.

eg. being able to use a mirror directly (LibResilient does magic there, but it would  be nice of one can test the github-pages mirror also directly)

that will simplify things - but yeah, if significant number of browser do not support that, we can go for absolutes, it's a pity, still :(

missytake: no, it's not so bad. You can just click the link in lynx, it just doesn't automatically redirect. no big deal, and I don''t know about other browsers.

missytake:
using SEO tools to check for `https://delta.chat/download` doesn't make sense because they do already follow the redirect. Also I think now that it is not a big deal SEO-wise, and that SEO isn't that important for our page anyway, because we rank well for "delta chat" and never tried to grow via other keywords.

missytake:
If we merge this, why not already include .html files for /donate, /contribute, and all the other language-sensitive links?

dl.html
r10s: ah, if one just has to click the destination for "security reasons" in lynx, it seems better to stay with relative URLs i'd say, even if not fully standard. i also do not know about another browser, in all the years, having a similar behaviour of not supporting relative URLs here

r10s:
yip, i also would not think about SEO too much, knowing that the client-side redirects _are_ followed is enough. 

(different ranking because of the type of redirect is a discussion ongoing since forever, i just wanted to mention it :)

r10s:
> If we merge this, why not already include .html files for /donate, /contribute, and all the other language-sensitive links?

i wanted to see that as a proof-of-concept first - and do not mess up with precedence and caching issues :) -- and therefore created a redirect that is not in use server-side - [but is also useful](https://github.com/deltachat/deltachat-pages/pull/989#issuecomment-2462379105)

i added a commit to adapt the existing `index.html`, however. as anyway existent, that should not add extra confusion with sever-side redirectly

r10s:
~~ah, one thing that is missing is preserving hashes as `help#sth` redirecting to `LANG/help#sth`~~ EDIT: added

not sure if sth else is missing ...

missytake:
> If we merge this, why not already include .html files for /donate, /contribute, and all the other language-sensitive links?

note: right now https://staging.delta.chat/990/contribute redirects to https://staging.delta.chat/990/en/contribute (or even to /ru/ with my test chromium installation), but that is because of server-side staging.delta.chat redirects. To preserve those pages I think we need to add .html redirect files for:

- chatmail
- blog
- community-standards
- contribute
- donate
- download (not only dl)
- gdpr
- gdpr-website
- help
- imprint
- references
- serverguide
- user-voices
- verify-downloads

I don't think they will work by themselves as it stands now; we could of course touch the staging.delta.chat nginx config to test that

r10s:
this is what the server-side-catch-all-redirect provides probably, however, i do not think, all of them are really in use.

while we can add all, i'd start adding the ones _really_ in use, eg. for `user-voices` or `verify-download`, i have not seen the link in the wild - if we have 404 properly working again, it is okay if sth. is missing - that happend also before

so, i'd shrink the list to  things known to be in use:

chatmail, blog, contribute, gdpr, donate, download (we can also remove the dl shortcut then), gdpr, help, serverguide

missytake:
> this is what the server-side-catch-all-redirect provides probably, however, i do not think, all of them are really in use.
> 
> while we can add all, i'd start adding the ones _really_ in use, eg. for `user-voices` or `verify-download`, i have not seen the link in the wild - if we have 404 properly working again, it is okay if sth. is missing - that happend also before
> 
> so, i'd shrink the list to things known to be in use:
> 
> chatmail, blog, contribute, gdpr, donate, download (we can also remove the dl shortcut then), gdpr, help, serverguide

we will not notice when people run into 404 errors, and we will not remember to add a client-side redirect once we start sharing a language-agnostic permalink. we should just add this for every page, also when we add a new one.

r10s:
let's keep thinks simple for now and not getting  so much legacy stuff in again :)

the important things are all added (i stayed with "download" btw) , and i doubt, the other things are in use at all. also for new pages, there is no need to add the language-independent redirects "just in case" - this is really only needed when they get into use - and we will know that, as it is us, start spreading the link.

good think is, that together with the PR - eg. a blogpost - we can now create any redirect we want

with fixed 404 page, it is good enough

EDIT: closed accidentally

missytake:
> and we will know that, as it is us, start spreading the link.

I doubt "us" knowing anything - collectivizing knowledge is not trivial. Otherwise we wouldn't need documentation. At least I am used to "removing the language prefix from a delta.chat URL is a good practice and will just work when I share it because then it gets shown in the recipients browser language". This will not work anymore if we are not consistent here

r10s:
but you will immediate notice, that a link you are starting to spread is not working, right?

having tons of client-side redirect, one additional file for eg. every blog post - this is not nice, unless generated automatically, but, well, this again adds complexity, we want to get rid of here. also, as blog posts are not translated anyways - and it is not really visible from the URL if a page is a blogpost or not.

we should not add additional maintenance burden, here, let's get the bare thing working. 

it is just one drawback from the change, "catch all" redirects are not really doable nicely client-sided - and, tbh, we never wrote that it is working to unconditionally remove the language-folder.

r10s:
merging that in as the server side redirects are removed at https://github.com/deltachat/sysadmin-pyinfra/pull/106

r10s:
btw, "just" removing language-folder never worked, just got that on trying a blog entry: `https://delta.chat/de/de/de/de/de/de/de/de/de/de/de/de/de/de/de/de/de/de/de/de/2024-03-25-crypto-analysis-securejoin`

PR #992, opened 2024-11-08
show download options in help & offline help

r10s:
this is mainly useful when browsing the help offline, when maybe having never opened the official homepage before. or, in case that gets offline,
show some alternatives, before or after the fact.

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -1230,7 +1230,7 @@ Delta Chat is available for all major and some minor platforms:
 
 - If unavailable, use the **mirror** at <https://deltachat.github.io/deltachat-pages>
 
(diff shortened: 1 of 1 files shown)


en/help.md
```diff
@@ -1221,6 +1221,23 @@ Otherwise you might receive undecryptable messages from those group chats.
 - See [Standards used in Delta Chat]({% include standards-url %}).
 
 
+
+### Where can my friends find Delta Chat?
+
+Delta Chat is available for all major platforms:
```
hpk42:
```suggestion
Delta Chat is available for all major and some minor platforms:
```

hpk42 (approved):

PR #1005, opened 2024-11-22
fix appimage link

hpk42:


missytake (approved):
thanks! I wonder why it worked in the past... is the build artifact called differently now?

WofWca:
There were some changes to artifact names like a month ago. Could be related.

nicodh:
Yes we had a renaming with the mono repo build changes, but after some broken link reports I tested all links (at least I thought all) :-(

PR #1007, opened 2024-11-27
use 'result' instead of '_site' as site folder

r10s:
use 'result' instead of '_site' as site folder; this unifies nix and t-dance output

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,9 +5,6 @@ _site/
 .*.swp
 *.po
 
-# ignore private scripts and directories, eg. jekyll-build-local.prv.sh
-*.prv*
-
 # ignore tools/tranlations, we copy the files to the corresponding language directories
 tools/translations/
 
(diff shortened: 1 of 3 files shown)


missytake (commented):
The Github actions, including our nginx config on page, still use `_site` as output name for the jekyll build - should we change this as well?

r10s:
if nginx uses the tools/t-dance.sh script it needs to be changed, indeed.

otherwise, it does not matter

missytake:
it doesn't - still I wonder if we might run into inconsistencies. shouldn't be a bad problem though - worst case, the website stops updating until we also change `_site` to `result` in the github actions + nginx config. :shrug:

r10s:
t-dance seems to be used locally only, and the CI in this PR already works, imu, it uses "nix"- which was also the easiest ways for @adbenitez to build things locally -- this PR is mainly to make subsequent tools use the same output dir

anyways, so far things seem be be fine

adbenitez (approved):

missytake:
Can this be merged?

PR #1012, opened 2024-12-18
Bump desktop to 1.50.1

nicodh:
To be merged when the store releases are approved


r10s (approved):

missytake:
Windows Store already offers 1.50.1, Mac Store is still budging, but in this case we should already merge this to make Linux and Windows .exe versions available.

PR #1013, opened 2024-12-23
Update links and qr codes

ralphtheninja:


38c3.html
```diff
@@ -205,7 +205,7 @@ <h3 id="find-us">Find us</h3>
       <a href="https://chaos.social/@delta">https://chaos.social/@delta</a>
     </p>
 
-    <a href="https://37c3.c3nav.de/l/c:0:140.67:109.93/"
```
ralphtheninja, line 208: Gives a 500

ralphtheninja: Is the location correct still, or is this the location from last year?

ralphtheninja:
I've tested the invite link on Android. It takes me to the correct group.

38c3.html
```diff
@@ -155,7 +155,7 @@ <h2 id="test-delta-chat-without-registration">
           Optional: join our event group with this qr code (click or scan)
           <br />
           <a
-            href="OPENPGP4FPR:6FE1642916908F1AC9CC7557CC99CF5DDB92043C#a=groupsbot%40testrun.org&g=Delta%20Chat%20%40%2037c3&x=nsPqJMGqPC5&i=zSMGD0NKD0e&s=2-Y4FTj5xx1"
+            href="OPENPGP4FPR:6FE1642916908F1AC9CC7557CC99CF5DDB92043C#a=groupsbot%40testrun.org&g=Delta%20Chat%20%40%2038c3&x=GjWs3QctuRW__Bt8nBQUdeht&i=_AG0Px0PlWZ&s=LR9BivV4QtJ"
```
link2xt: Needs a new QR code as well.

38c3.html
hpk42: it's the location from last year i think.  not sure we have a new location already.

38c3.html
ralphtheninja: @link2xt What do you normally use for generating the qr codes? What's the exact text that should be in the qr code?

38c3.html
ralphtheninja: Pointing to the Komona area.

38c3.html
link2xt: Current one is correct, it is an https://i.delta.chat/... link inside.

link2xt (approved):

ralphtheninja:
This is what it looks like now

image

@link2xt Maybe we could even increase the max width to 400px now? (currently 300)

ralphtheninja:
image

Maybe ready to squash-merge this now? We can add more tweaks on top of this later.

PR #1015, opened 2024-12-24
Add cosmos to 38c3 page

link2xt:


link2xt:
Does not look good, a simple link would be better.

PR #1018, opened 2024-12-24
bring wording more in line with https://delta.chat

hpk42:
and don't talk so much about accounts :)


Simon-Laux (approved):

Simon-Laux:
for a better preview this should be rebased

PR #1019, opened 2024-12-25
several refinements, better links. 

hpk42:
Try to keep headings using the 38c3 font short.

also i think it makes sense to mention "super app" here and there.

### DIFF
--- a/38c3.html
+++ b/38c3.html
@@ -107,7 +107,7 @@
         alt="deltachat logo"
       />
       <p>
-        Delta Chat is a secure and decentralized super app offering
+        Delta Chat is a secure and decentralized super app offering ...
       </p>
     </div>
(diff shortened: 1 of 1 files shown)


38c3.html
```diff
@@ -107,42 +107,42 @@ <h2>Delta Chat @ 38c3</h2>
         alt="deltachat logo"
       />
       <p>
-        Delta Chat is a decentralized and secure messenger app
+        Delta Chat is a secure and decentralized super app offering
```
r10s: i'd add an ellipsis here as it is not obvious by typography that the sentence continues

```suggestion
        Delta Chat is a secure and decentralized super app offering ...
```

r10s (approved):

PR #1031, opened 2025-01-22
blog post about billionaires, webxdc app discovery and notifications

hpk42:


lk108:
Great post!

s/Lorimi/Lomiri (should I directly edit such typos?)

hpk42:
> Great post!

thanks!

> s/Lorimi/Lomiri (should I directly edit such typos?)

you could fix yourself, but i did in this case already.

PR #1038, opened 2025-01-27
Update and optimize Chinese translation

fengguowudi:
this PR is replaced by https://github.com/deltachat/deltachat-pages/pull/1039

for the future, translations are better done on transifex directly

### DIFF
--- a/zh_CN/contribute.md
+++ b/zh_CN/contribute.md
@@ -3,50 +3,49 @@ title: 贡献
 lang: zh_CN
 ---
 
-# 频道与仓库
+# 渠道和仓库
 
-在参与围绕 Delta Chat 的社区互动前，请阅读我们的[社区准则](community-standards)。
+为了围绕 Delta Chat 的社区互动，请阅读我们的 [社区标准](community-standards)。
 
-- [Fediverse](https://chaos.social/web/@delta) for announcements and feedback.
+- [联邦宇宙 (Fediverse)](https://chaos.social/web/@delta)：用于发布公告和接收反馈。
 
-- [Delta Chat 支持论坛](https://support.delta.chat) ：更广泛的功能讨论和获得支持。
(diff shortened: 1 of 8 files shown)


r10s:
thanks a lot for your effort!

however, translations are done via transifex (see [README](https://github.com/deltachat/deltachat-pages?tab=readme-ov-file#translate-the-website) and [Contribute](https://delta.chat/en/contribute)) - so, if we would merge this PR, things would get reverted soon :/

so, it would be very cool if you can do further changes on transifex (you can log in with your github handle there) - there is only a better way of discussing some string changes, that i, as speaking no word of chinese, cannot judge. non-changes, but just added translations, if course, usually do not require much discussions :)

r10s:
@fengguowudi if you give me your transifex handle, i can add you there, maybe i can also help on copy+pasting some strings over, esp. the FAQ of course are valueable, as also added to the app

r10s:
i copied the homepage over to transifex, will continue with other files so the effort is not wasted :), however, as said future translations should go directly to transifex

preview for the hompage is at https://staging.delta.chat/1039/zh_CN/ :

Screenshot 2025-01-27 at 15 50 03

r10s:
also copied help and the contribute page :) PR is at https://github.com/deltachat/deltachat-pages/pull/1039

again, thanks! great stuff, soon chinese website will be translated at 100% :)

r10s:
moved all related, added strings over to #1039, which also gets merged now, so that the changes are visible on https://delta.chat/zh_CN/ in some minutes ...

Screenshot 2025-01-27 at 23 52 09

fengguowudi:
applied for

r10s:
> applied for

great! i approved you there, you should also be able to edit existing translations, in case there are error. otherwise only 3 strings on https://delta.chat/zh_CN/chatmail are missing.

PR #1055, opened 2025-02-12
WIP: Remove mention of US hosting for Chika

ghost:
My chatmail server is no longer hosted in the so-called US and has moved to a Ukrainian data center.


ghost:
I added a badge as well to perhaps help people decide on their server choice. Is that ok?

missytake (approved):
Sure, works. But only 44% uptime? that's unfortunate :/ not sure how much sense it makes to display the uptime for one server but not for the others.

ghost:
Sorry for the noise the VPS that I moved to unfortunately isn't working anymore so I had to move back. And yeah makes more sense if all of them had badges!

PR #1057, opened 2025-02-21
feat: add required javascript for pagefind in desktop help pages

nicodh:
Adds javascript needed for https://github.com/deltachat/deltachat-desktop/pull/4691


r10s:
as this script is used also for iOS and android, which already have a search, we should add the `<script>` tags only conditionally (similar to `--add-top-links`)

`<title>` and `<span>` anchor can be added unconditionally - but is  `<title>` needed? it is not localised and seems not to be displayed at all on desktop

nicodh:
> as this script is used also for iOS and android, which already have a search, we should add the `<script>` tags only conditionally (similar to `--add-top-links`)

ah ok, didin't know that - of course I will add a condition for that
> 
> `<title>` and `<span>` anchor can be added unconditionally - but is `<title>` needed? it is not localised and seems not to be displayed at all on desktop

title can be removed since it's dynamically set (and localized) with Javascript onload

nicodh:
I added the requested changes and refactored the script a bit

r10s (approved):
lgtm, did not test

PR #1083, opened 2025-03-23
Update README.md

wakest:
cleaned up some English types and grammar

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,9 @@ Updating typos in the English sources
 **that shall not result in retranslations**
 is a bit tricky:
 
-The English source is used as the ID for the translations, so fixing a typo will change the ID and result in translations needing to be redone.
+The English source is used as the ID for the translations,
+so fixing a typo will change the ID
+and result in translations needing to be redone.
 
 Currently, this can be handled only manually:
 - fix the typo or grammar
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Formatting of blog posts and web pages
 In the various source text files that are used 
 for generating the static https://delta.chat website 
 we use [Semantic linefeeds](http://rhodesmill.org/brandon/2012/one-sentence-per-line/)
-to keep "diffs" for small changes. The gist of it is
+to keep "diffs" for changes small. The gist of it is
 that you put each "phrase" part of a sentence into a single line.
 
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,15 +2,15 @@ Delta Chat Website
...
+feel free to edit any file you like and send a pull request.
 
 Formatting of blog posts and web pages
 --------------------------------------
 
 In the various source text files that are used 
 for generating the static https://delta.chat website 
 we use [Semantic linefeeds](http://rhodesmill.org/brandon/2012/one-sentence-per-line/)
-to keep "diffs" for changes small. The gist of it is
+to keep "diffs" for small changes. The gist of it is
```
r10s: not sure about grammar, but imu this suggestion changes the meaning. 

it is about "having fewer changes in the differential outout", not about "small changes"

README.md
```diff
@@ -59,16 +59,16 @@ Testing changes
...
 --------------------------------------------------------------------------------
 
 - `/en` - Contains the source language files in Markdown. Can be edited directly.
 
-- `/de`, `/es`, `fr` … - Contain the translated files, use Transifex to edit them;
-  you **MUST NOT** edit any file in these directories,
-  they will be overwritten with the next pull from Transifex.
+- `/de`, `/es`, `fr` … - Contains the translated files, use Transifex to edit them;
+  you **MUST NOT** edit any files in these directories,
+  as they will be overwritten with the next pull from Transifex.
```
r10s, line 71: if it is more than one directory, so, plural, isn't "contain" correct? so, "these directories contain the translated files ..."

README.md
```diff
@@ -135,29 +135,27 @@ for more info see the comments in `./tools/t-dance.sh`.
...
 
-Updating typos in the english sources
+Updating typos in the English sources
 **that shall not result in retranslations**
 is a bit tricky:
 
-The english source is used as the ID for the translations -
-fixing a typo will change the ID an
-will result in translations needed to be redone.
+The English source is used as the ID for the translations, so fixing a typo will change the ID and result in translations needing to be redone.
```
r10s: let's stay with semantic lineends mentioning above :)
```suggestion
The English source is used as the ID for the translations,
so fixing a typo will change the ID
and result in translations needing to be redone.
```

r10s:
@wakest thanks a lot!  - a few comments above

README.md
r10s:
```suggestion
to keep "diffs" for changes small. The gist of it is
```

PR #1085, opened 2025-03-26
FAQ typofixing

link2xt:


missytake:
Keep in mind that this will trigger translations, and the whole paragraph needs to be re-translated then... maybe we can ignore spelling companies correctly and only fix the english typos?

r10s (approved):

r10s:
> Keep in mind that this will trigger translations, and the whole paragraph needs to be re-translated then... 

that's true :/

let's still merge that in, i will rescue the translations by some half-manual t-dance :)

PR #1088, opened 2025-03-27
update DC Community group link

adbenitez:


r10s:
let's do that over an "_include", sonwe can change that in the future for all translations at the same time and without forcing retranslations; i will add a commit to the pr

r10s (approved):

PR #1095, opened 2025-04-08
Remove outdated paragraph about registering with newsletters

missytake:
Since new chatmail relay addresses can't receive unencrypted messages anymore, this is outdated.


hpk42 (commented):
I think this needs an "update: since April 2025 chatmail relay servers can not be used for receiving unencrypted messages anymore so we removed this paragraph." in italics or so.

hpk42 (approved):

PR #1109, opened 2025-05-10
Update donation text

nicodh:
The current text is not very appealing. 

More translations would be useful.

We could/should link to this page also in release messages. 
Or send a dedicated message for donations like a week after first install?

### DIFF
--- a/de/donate.md
+++ b/de/donate.md
@@ -5,7 +5,13 @@ lang: de
 
 # Geld spenden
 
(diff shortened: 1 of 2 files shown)


missytake:
Great step forward :)

Translations only happen on transifex, based on the english version, so we don't need to write anything german here (and it actually gets overwritten by transifex).

hpk42 (approved):
good improvements but the german donate page can be removed from this PR.

PR #1112, opened 2025-05-12
[WIP] Add blog post about the statistics bot

Hocuri:
To accompany https://github.com/chatmail/core/pull/6851

TODO: Replace XXX with a link to the full list of collected metrics

### DIFF
--- a/_posts/2025-05-anonymous-usage-reporting.md
+++ b/_posts/2025-05-anonymous-usage-reporting.md
@@ -31,11 +31,11 @@ You can enable it at:
 
 We are interested in numbers like:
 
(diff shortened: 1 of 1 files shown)


hpk42:
a) the motivation is a little vague ("usable security") maybe also because the metrics are not listed 
b) there is no explanation or policy what happens with the data 
c) it should be mentioned early on that delta chat developers don't know numbers about failures-to-decrypt or other exceptional application states and that the "send statistics" function provides an opt-in mechanism to help us get some real-world data.

r10s:
ftr, meanwhile, this is partly covered by https://delta.chat/en/help#statssending

PR #1122, opened 2025-05-22
turn a link into a permalink in "Browser edition"

WofWca:
Otherwise it might break eventually.
But with a permalink people can always switch
to the current version of the file.


Simon-Laux:
though this way it will not be the newest version of the browser edition 🤔

WofWca:
True true. Well OK, let's close.

PR #1124, opened 2025-05-28
adapt broadcast lists behaviour to reality

r10s:
if we force re-translation because of https://github.com/deltachat/deltachat-pages/pull/1123 , we should also strike the other out-of-date claims.

once  this is merged, we can push strings to translators.

(in general, i do not think, it makes much sense to explain experimental features in details in the faq, it seems better to link to a forum entry or to a PR - or just say nothing and move things a bit more out of sight in the UI, with a warning. experimental things should mostly be used when ppl are know what they're doing by other sources. as of 2025, we do not really want to play avg users play around with that :)

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -1021,17 +1021,10 @@ another rather stable experiment we run on the side (sic!).
 
 ### What are Broadcast Lists and how can I use them?
 
-- With a Broadcast List you can send a message to many recipients at once; when
-  they reply to you, you get the reply in your direct 1:1 chat with them. The
-recipients can't see each other.
-- Technically, it is an E-Mail with many recipients in BCC.
-- You can turn on the feature in the "experimental features" section in the
-  advanced settings. Then you can create a Broadcast List from the "New Chat"
-dialog.
-- Messages sent to broadcast lists are not encrypted. Encryption would break
-  anonymity, because then all recipients would know who else received it
-  (Sending individual mails to everyone would be worse for rate limit and network
-  consumption reason).
+With a Broadcast List you can send a message to many recipients at once;
+the recipients cannot reply in that list.
+Broadcast lists are still highly experimental
+and will very probable be replaced by something else, stay tuned :)
 
 
 ### How can I share my location with my chat partners?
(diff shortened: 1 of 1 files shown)


en/help.md
```diff
@@ -1021,17 +1021,10 @@ another rather stable experiment we run on the side (sic!).
...
-  advanced settings. Then you can create a Broadcast List from the "New Chat"
-dialog.
-- Messages sent to broadcast lists are not encrypted. Encryption would break
-  anonymity, because then all recipients would know who else received it
-  (Sending individual mails to everyone would be worse for rate limit and network
-  consumption reason).
+With a Broadcast List you can send a message to many recipients at once;
+the recipients cannot reply in that list.
+Broadcast lists are still highly experimental
+and will very probable be replaced by something else, stay tuned :)
```
```suggestion
and will very probably be replaced by something else, stay tuned :)
```

missytake (commented):
Hm, I see how we should fix "users can reply to broadcast lists".

But removing all the other content... I don't see how that helps anyone? And I think it triggers a retranslation, or is that triggered anyway? Maybe we should wait until the new feature is ready...

overall I'm +- 0 on this change.

r10s:
> But removing all the other content... I don't see how that helps anyone?

it is wrong and misleading:

- "you get the reply in your direct 1:1 chat with them." -> ~~wrong, it creates own chats~~ EDIT:  ah, that is meant for replies, not for receiving. still, iirc this comes from a time where received was also in 1:1
- "technically, it is an E-Mail with many recipients in BCC." -> we do not want to talks about email, bcc etc. this is not helpful
- Messages sent to broadcast lists are not encrypted. -> wrong
- Explanation "why not encrypted" -> outdated
- the remaining things will be outdated soon as well

i think, we should just say as few as possible about experiments, it is too fast outdated and no one takes care. 

> or is that triggered anyway

yes, it will be triggered by #1123 which i did not pushed yet to transifex to not force retranslation of wrong / outdated things

missytake (approved):

missytake:
> yes, it will be triggered by #1123 which i did not pushed yet to transifex to not force retranslation of wrong / outdated things

Ah, we would need empty lines between the bullet points I guess :/ good to know for the future.

PR #1125, opened 2025-05-29
feat: improving es translation

josueggh:
Improving spanish tranlations

### DIFF
--- a/es/chatmail.md
+++ b/es/chatmail.md
@@ -1,57 +1,40 @@
 ---
-title: Chatmail for instant, interoperable and privacy-preserving messaging
+title: Chatmail para mensajería instantánea, interoperable y respetuosa con la privacidad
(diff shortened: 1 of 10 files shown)


missytake:
Thanks for translating! Unfortunately you are submitting it through the wrong channel... our translations happen on transifex, and if we merge this PR it would be overwritten soon :/

You can login to transifex with your github account, can you submit the changes there instead? https://www.transifex.com/delta-chat/delta-chat-pages/
That would be great!

josueggh:
I'll do , thanks for the update.

r10s:
thanks, then we can close this pr

josueggh:
@missytake I tried to login but it looks a business email is required.

r10s:
you can also log in with github

PR #1133, opened 2025-06-24
FAQ: Changing e-mail addresses is temporarily disabled

felurx:
https://github.com/chatmail/core/commit/4a2bfe03dab3c8a7f3297cb47719d0f3392c7201 disabled AEAP. To avoid the terrible disappointment / slight annoyance that comes with reading the FAQ, trying to change your address and it not actually working, I think there should be a note in the FAQ. Since the error message states that it should be back in a few months, I tried to mirror that.


missytake (approved):
Thanks, that's indeed sad.

PR #1138, opened 2025-07-10
avoid crash in timeout

adbenitez:
https://microg.org site is offline which causes the whole help generation script to crash checking URLs and not to actually generating the offline help


r10s (approved):

r10s:
microg is stil not reachable, btw

EDIT: there was a physical server damage on july, 5th, https://github.com/microg/GmsCore/issues/2965

PR #1154, opened 2025-07-30
FAQ: discourage parallel usage with other email apps

### LINKED ISSUE #1104: FAQ entry on that/why usage together with normal MUA on the same account is not recommended
we get enough people asking for it, I'd like to have an easy link to refer to

missytake:
closes #1104 

Any other reasons we can think of?

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -998,6 +998,22 @@ This is what Delta Chat does with these permissions:
   see [Provider Overview](https://providers.delta.chat)
 
 
+### Can I use Delta Chat in parallel with other E-Mail apps?
+
+Yes, but it is not recommended.
+
+Parallel usage with the same E-Mail address might lead to the following inconveniences:
+
+- Next to your Delta Chat notifications,
+  you might also get notifications for (encrypted, so unreadable) chat mail messages
+  in your other email app
+- If Delta Chat is not running
+  or "Advanced > Move automatically to the DeltaChat Folder" is disabled,
+  encrypted Delta Chat messages might clutter the Inbox in your other E-Mail apps.
+
+[Sieve rules](https://support.delta.chat/t/sieve-rule-to-move-deltachat-mails-to-deltachat-folder/288/10) can solve these problems.
+
+
 ### I want to manage my own server for Delta Chat. What do you recommend?
 
 - Most mail servers will work well. But what we personally recommend is a


en/help.md
```diff
@@ -998,6 +998,16 @@ This is what Delta Chat does with these permissions:
   see [Provider Overview](https://providers.delta.chat)
 
 
+### Can I use Delta Chat in parallel with other E-Mail apps?
+
+Yes, but it is not recommended.
+
+Parallel usage with the same E-Mail address might lead to the following inconveniences:
+
+- You might receive double notifications
```
Simon-Laux: elaborate: you also get notifications for the (encrypted) chat mail messages in your other email app.

en/help.md
```diff
@@ -998,6 +998,16 @@ This is what Delta Chat does with these permissions:
...
+- Depending on the configuration, encrypted Delta Chat messages might clutter the Inbox in your other E-Mail apps
```
Simon-Laux: say that you either need a filter rule in your other mail client, a server side filter rule or ensure that delta chat is always running and has the "move to DeltaChat Folder" option enabled.

en/help.md
```diff
@@ -998,6 +998,16 @@ This is what Delta Chat does with these permissions:
...
+
+
```
Simon-Laux, line 1016: hints on how you can fix this, but note that we don't support it officially:
- k9-mail / thunderbird on android has an option to hide all chat email messages (https://github.com/thunderbird/thunderbird-android/pull/5578)
- some email apps (like thunderbird) have a filter feature where you can setup a filter that moves the chat email messages to the DeltaChat IMAP folder.
- The best solution is a server side filte/sieve rule that moves the messages, but many email providers don't have that feature. (https://support.delta.chat/t/sieve-rule-to-move-deltachat-mails-to-deltachat-folder/288/10)

Keep also in mind that the quality of email services varies greatly and that most mainstream providers have limits that make for a bad experience (link to FAQ section about email provider choice and that you should use chatmail).

Simon-Laux (commented):
some improvement ideas

en/help.md
r10s: yeah, some more advices make sense!

however, this level of detail is better be placed in a forum entry (we have a section "Guides"), there it can also be maintained easier, ppl can comment etc.

as said, we want the whole email thing here very _short_, it is not the focus we have and works against other goals. it should not take too much room of the whole faq

Simon-Laux (approved):
Good Improvement already, we can always iterate in the future.

PR #1158, opened 2025-08-01
fix screenshots for 2.x

r10s:
adapted quickly using gimp, not perfect, but better than before

recreating from scratch would be another effort (requiring some creativity :) as it is not documented how borders were created, which screen sizes and devices were used etc., let alone the chat content (backup only helps a little, as all dates are outdated :) 

we are improving slightly, the [iOS screenshots](https://github.com/deltachat/deltachat-ios?tab=readme-ov-file#delta-chat-ios-client) (source in interface repo) could be adapted easier as the tooling was written to some README


hpk42 (approved):
Lgtm

r10s:
one second, there is a little fix pending, then i merge ...

PR #1167, opened 2025-08-19
remove unmaintained community experiment

r10s:
Delta Chat is for "family & friends", ppl one trusts.

the public community is currently somehow unmaintained, also getting here and there social issues by some ppl using outdating clients or doing harm on purpose. also, ppl are experimenting partly with weird setups and keys there.

apart from that, technically the experiment went nice so far, 150+ ppl could chat e2ee in a reliable way


missytake (approved):

r10s:
moderated alternative at https://github.com/deltachat/deltachat-pages/pull/1168

PR #1173, opened 2025-08-22
make clearer, that non-chatmail is supporting e2ee

r10s:
this small change may make it clearer that e2ee can be used by non-chatmail as well.

suggestion taken one-to-one from https://support.delta.chat/t/bring-back-avatars-redaction-modifying-member-list-to-unencrypted-chats/3977/38


Hocuri (approved):
:+1:

missytake (approved):

PR #1174, opened 2025-08-30
fix blog links in mirrors

r10s:
mirrors may be placed in a subdirectory,
this way, links continue to work;
we use the `../` approach also elsewhere, eg. on language selection, so this does not introduce smth new.

baseurl seems just not to be set,
and it is also not really used elsewhere.

may still be nice to hace mirrors on root,
however, with this PR this is less a prerequise.

closes https://github.com/deltachat/sysadmin/issues/197


missytake (approved):
Thank you :)

PR #1181, opened 2025-10-20
serverguide: weaken discouragement of mailcow

missytake:
I got the following feedback from a friend:

> Im $hackspace kam grade auf warum deltachat kein mailcow mehr empfiehlt (https://delta.chat/de/serverguide). Nehme an das ist nur eine Empfehlung für das chatmail relay setup und nicht gegen mailcow? Wird anscheinend missverstanden, also ist evtl eine umformulierung wert.

For some use cases, mailcow is still a good option, and we should not *discourage* setting up mailcow, but *encourage* chatmail relays.

### DIFF
--- a/en/serverguide.md
+++ b/en/serverguide.md
@@ -5,8 +5,8 @@ lang: en
 
 # How to Setup a Mail Server for Delta Chat
 
-> **Update:** This is outdated - we no longer recommend mailcow Servers,
-> but [Chatmail Relays](2023-12-13-chatmail) for chatting with Delta Chat.
+> **Update:** This is outdated - we now recommend
+> [Chatmail Relays](2023-12-13-chatmail) over mailcow servers for chatting with Delta Chat.
 > Read more [on GitHub](https://github.com/chatmail/relay) about how to set it up!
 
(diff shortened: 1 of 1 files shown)


Simon-Laux (approved):

en/serverguide.md
```diff
@@ -5,8 +5,8 @@ lang: en
 
 # How to Setup a Mail Server for Delta Chat
 
-> **Update:** This is outdated - we no longer recommend mailcow Servers,
-> but [Chatmail Relays](2023-12-13-chatmail) for chatting with Delta Chat.
+> **Update:** This is outdated - we now recommend
+> [Chatmail Relays](2023-12-13-chatmail) over mailcow servers for chatting with Delta Chat.
```
adbenitez: I think "instead" is clearer than "over"

missytake: Instead would come over too strong in my opinion - over means that mailcow is still a second good option.

Simon-Laux:
It could also mention the concrete reasons why chatmail is better/different:
1. faster delivery
2. reliable push notifications
3. forced message encryption

PR #1187, opened 2025-10-30
redirect /chatmail to chatmail.at

r10s:
for deleting en/chatmail: it was hardcoded in iOS at https://github.com/deltachat/deltachat-ios/pull/2860 - as we may not not want a redirect inside the local folders, i'd leave that for now and cleanup at some later point

### DIFF
--- a/chatmail.html
+++ b/chatmail.html
@@ -1,3 +1,3 @@
 <!DOCTYPE html>
-<meta http-equiv="refresh" content="1; url=en/chatmail">
+<meta http-equiv="refresh" content="1; url=https://chatmail.at/relays">
 <script src="redirect.js"></script>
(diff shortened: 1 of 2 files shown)


chatmail.html
```diff
@@ -1,3 +1,3 @@
 <!DOCTYPE html>
-<meta http-equiv="refresh" content="1; url=en/chatmail">
+<meta http-equiv="refresh" content="1; url=https://chatmail.at/relays">
 <script src="redirect.js"></script>
```
adbenitez: the script will not work, if you look at what `redirect.js` does, it expects a local/relative path, fix:

```suggestion
<meta http-equiv="refresh" content="1; url=https://chatmail.at/relays">
<script>window.location.href = "https://chatmail.at/relays"</script>
```

adbenitez (approved):

chatmail.html
r10s: indeed, that's much easier than adapting the script

r10s: i reverted https://github.com/deltachat/deltachat-pages/pull/1187/commits/cd92b9c9aa093411e34e321aa630107092f13c9e :)

PR #1191, opened 2025-11-06
simplify green dot and other docs

r10s:


en/help.md
```diff
@@ -475,48 +462,48 @@ This method is only recommended if "Add Second Device" as described above does n
...
 
+- An in-chat app can only exchange data within a Delta Chat chat, with its
+  copies on the devices of your chat partners. Other than that, it's completely
+  isolated from the Internet.
+
+- The privacy an in-chat app offers is the privacy of your chat - as long as you
+  trust the people you chat with, you can trust the in-chat app as well.
 
-### How can I create my own webxdc apps?
+- This also means: Just like for web links, do not open apps from untrusted contacts.
```
missytake, line 489: isn't the analogy more "attachments" than web links?

r10s: the comparison is about links: if you tap them, you may reveal the ip address to the sender, and the sender may trick you to enter secrets

an attachment may be similar i some regards, but we only want to have one item that is easy to explain and to understand for avg user. tons of text and relativations does not help here.

ftr, the recent PR were done in bb33 with 4 eyes or more

PR #1195, opened 2025-11-08
update metadata section

r10s:
- date is removed/randomized
- avg. user would understand by "address" smth more "official" as phone number, already say here that it is random (it is mentioned below, but this outstanding part is what ppl will remember)

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -727,9 +727,10 @@ Instead, all group metadata is end-to-end encrypted and stored on end-user devic
 
 Servers can therefore only see:
 
-- random sender and receiver addresses, by default
+- the sender and receiver addresses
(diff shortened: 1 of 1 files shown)


en/help.md
```diff
@@ -727,9 +727,7 @@ Instead, all group metadata is end-to-end encrypted and stored on end-user devic
 
 Servers can therefore only see:
 
-- the message date, 
-
-- sender and receiver addresses 
+- random sender and receiver addresses, by default
```
link2xt: "Servers can therefore only see the sender and receiver addresses and the message size. By default the addresses are randomly generated."

link2xt (approved):

PR #1199, opened 2025-11-12
provide reasoning and emphasize that a chatmail address/mailbox can only be used between chatmail clients

hpk42:
also streamlines the "what is delta chat" question to intertwine with chatmail clients.

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -570,11 +570,11 @@ You can find more information
 and give feedback in the [Forum](https://support.delta.chat).
 
 
-### Can I use a classic email address with Delta Chat?
+### Can I use a classic email account with Delta Chat?
 
-Yes, but only if the email address is used exclusively by [chatmail clients](https://chatmail.at/clients).
+Yes, but only if the email account is used exclusively by [chatmail clients](https://chatmail.at/clients).
 
-It is not supported to share a mailbox with non-chatmail apps or web-based mailers for the following reasons:
+It is not supported to share an email account with non-chatmail apps or web-based mailers for the following reasons:
 
 - Non-chatmail apps are largely not accomplishing automatic end-to-end email encryption for their users,
   while chatmail apps and relays pervasively enforce end-to-end encryption and security standards.
@@ -586,13 +586,13 @@ It is not supported to share a mailbox with non-chatmail apps or web-based maile
   would require considerable development and maintenance efforts,
   and complicate making chatmail-based messaging more resilient, reliable and fast.
 
-### How can I configure a chat profile with a classic email address? {#classic-email}
+### How can I configure a chat profile with a classic email account? {#classic-email}
 
-First off, **please do not use the same classic email address with non-chatmail apps**
+First off, **please do not use the same classic email account with non-chatmail apps**
 unless you are prepared to deal with encrypted messages in the inbox,
 double notifications, accidentally deleted emails or similar annoyances.
 
-You can configure a classic email address at **New Profile → Use Other Server → Use Classic Mail as Transport**.
+You can configure a classic email account at **New Profile → Use Other Server → Use Classic Mail as Transport**.
 Some classic email providers have limitations,
 see [Provider Overview](https://providers.delta.chat),
 and [Push notifications](#instant-delivery) will not be supported.
--- a/en/help.md
+++ b/en/help.md
@@ -15,7 +15,7 @@ available for mobile and desktop platforms.
   with secure and interoperable [chatmail relays](https://chatmail.at/relays)
   that offer instant message delivery and Push Notifications for iOS and Android devices.
 
-- Pervasive [multi-profile](#multiprofile) and
+- Pervasive [multi-profile](#multiple-accounts) and
   [multi-device](#multiclient) support on all platforms
   and between different [chatmail apps](https://chatmail.at/clients).
 
@@ -90,7 +90,7 @@ and can tap it to start chatting with the first contact.
 - For performance, images are optimized and sent at a smaller size by default, but you can send it as a "file" to preserve the original.
 
 
-### What are profiles? How can I switch between them? {#multiple-accounts} {#multiprofile}
+### What are profiles? How can I switch between them? {#multiple-accounts}
 
 A profile is **a name, a picture** and some additional information for encrypting messages.
 A profile lives on your device(s) only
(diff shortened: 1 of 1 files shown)


missytake (approved):

en/help.md
```diff
@@ -561,19 +564,33 @@ You can find more information
 and give feedback in the [Forum](https://support.delta.chat).
 
 
-### Can classic email be used as transport for a profile? {#classic-email}
+### Can I use a classic email address with Delta Chat?
```
adbenitez, line 568: to some level this is a regression cc @r10s

en/help.md
```diff
@@ -8,13 +8,16 @@ header: Frequently Asked Questions
 
 ## What is Delta Chat?
 
-Delta Chat is a reliable, decentralized and secure messaging app, 
+Delta Chat is a reliable, decentralized and secure instant messaging app,
 available for mobile and desktop platforms.
 
-- Anonymous chat profiles with fast, secure and interoperable [chatmail servers](https://delta.chat/chatmail)
-  which offer instant Push Notifications for iOS and Android devices.
+- Instant creation of private chat profiles
```
r10s: great to introduce "private chat profiles", it seems one of the key points. i'd even emphasise that to not be less outstanding as the links that are less interesting for average user (relay, chatmail, what's that?)

```suggestion
- Instant creation of **private chat profiles**
```

en/help.md
r10s: i think, it is fine. the "can i use as transport" is honestly not really a question user have.

but i would be more precise here, the question is about logging in to a classic email account, not eg. writing to a classic address.

```suggestion
### Can I use a classic email account with Delta Chat?
```

en/help.md
```diff
@@ -561,19 +564,33 @@ You can find more information
 and give feedback in the [Forum](https://support.delta.chat).
 
 
-### Can classic email be used as transport for a profile? {#classic-email}
+### Can I use a classic email address with Delta Chat?
 
-In general, this is possible at **New Profile → Use Other Server → Use Classic Mail as Transport**.
+Yes, but only if the email address is used exclusively by [chatmail clients](https://chatmail.at/clients).
+
+It is not supported to share a mailbox with non-chatmail apps or web-based mailers for the following reasons:
```
r10s: there are already so many terms, not sure if we should introduce another one with "mailbox"

```suggestion
It is not supported to share an account with non-chatmail apps or web-based mailers for the following reasons:
```

similar for "non-chatmail" - maybe better stay with "classic email (app)"

en/help.md
```diff
@@ -8,13 +8,16 @@ header: Frequently Asked Questions
...
-- Anonymous chat profiles with fast, secure and interoperable [chatmail servers](https://delta.chat/chatmail)
-  which offer instant Push Notifications for iOS and Android devices.
+- Instant creation of private chat profiles
+  with secure and interoperable [chatmail relays](https://chatmail.at/relays)
+  that offer instant message delivery and Push Notifications for iOS and Android devices.
 
-- Pervasive [multi-profile](#multiple-accounts) and multi-device support on all platforms.
+- Pervasive [multi-profile](#multiple-accounts) and
+  [multi-device](#multiclient) support on all platforms
+  and between different [chatmail clients](https://chatmail.at/clients).
```
r10s: this very first term should be as easy to understand as possible, an "app" is much more clear for avg user.

most user will also not understand "chatmail" at that stage, but that's probably kind of fine, an external link to smth technical, well :)

```suggestion
  and between different [chatmail apps](https://chatmail.at/clients).
```

en/help.md
```diff
@@ -561,19 +564,33 @@ You can find more information
...
+### How can I configure a chat profile with a classic email address? {#classic-email}
+
+First off, **please do not use the same classic email address with non-chatmail apps**
+unless you are prepared to deal with encrypted messages in the inbox,
+double notifications, accidentally deleted emails or similar annoyances.
+
+You can configure a classic email address at **New Profile → Use Other Server → Use Classic Mail as Transport**.
 Some classic email providers have limitations,
 see [Provider Overview](https://providers.delta.chat),
 and [Push notifications](#instant-delivery) will not be supported.
```
r10s: we should mention the fact that only INBOX is scanned, cmp. https://github.com/deltachat/deltachat-desktop/issues/5712 that we would probably like to point here

```suggestion
and [Push notifications](#instant-delivery) will not be supported.
Chatmail uses the default INBOX for message relay; ensure the provider does too. 
```

i would be really short on that, in general, i would give that classic email not more space than really needed (already unsure about two questions :) that we scan "spam" as well currently, is sort of a technical detail at that point. important is only that messages are not sorted to "SMART INBOX" or so

r10s (approved):
thanks a lot!

i did some suggestions that you may or may not pick up, no blockers.

Simon-Laux (approved):

r10s (approved):

en/help.md
```diff
@@ -8,13 +8,16 @@ header: Frequently Asked Questions
...
 
-Delta Chat is a reliable, decentralized and secure messaging app, 
+Delta Chat is a reliable, decentralized and secure instant messaging app,
 available for mobile and desktop platforms.
 
-- Anonymous chat profiles with fast, secure and interoperable [chatmail servers](https://delta.chat/chatmail)
-  which offer instant Push Notifications for iOS and Android devices.
+- Instant creation of **private chat profiles**
+  with secure and interoperable [chatmail relays](https://chatmail.at/relays)
+  that offer instant message delivery and Push Notifications for iOS and Android devices.
```
r10s: one can read that, as if instant message delivery is only available on iOS and Android.

or even as if the whole app is available for Android and iOS only.

```suggestion
  that offer instant message delivery and Push Notifications.
```

or 

```suggestion
  offering Push Notifications on iOS and Android and instant message delivery on all platforms.
```

i would tend to the first one, Push Notifications are less a thing on desktop anyways

en/help.md
```diff
@@ -86,7 +89,7 @@ and can tap it to start chatting with the first contact.
 - For performance, images are optimized and sent at a smaller size by default, but you can send it as a "file" to preserve the original.
 
 
-### What are profiles? How can I switch between them? {#multiple-accounts}
+### What are profiles? How can I switch between them? {#multiple-accounts} {#multiprofile}
```
r10s: having more than one anchor unfortunately does not render correctly.

<br><br><br><br>


as i am not aware of external links, however, it seems fine to use only `{#multiprofile}`

```suggestion
### What are profiles? How can I switch between them? {#multiprofile}
```

en/help.md
r10s: i added a comma for clarification

PR #1201, opened 2025-11-14
FAQ: Add description of groups and how/why clone

rakoo:


fr/help.md
```diff
@@ -213,6 +213,14 @@ They will most likely also not be decryptable anymore
 
 ## Groupes
 
+Un groupe est une liste de membres discutant ensemble. Il n'y a pas de rôles dans les groupes: tous les membres peuvent ajouter et supprimer d'autres membres, y compris le membre créateur.
```
Simon-Laux: AFAIK, translation needs to be done later and from transifex:

https://github.com/deltachat/deltachat-pages?tab=readme-ov-file#translate-the-website

rakoo: Oh ? Some of the paragraphs in the `fr` version don't exist in the `en` version, I thought it was done directly here. I'll delete it

rakoo:
I'm not sure what fails in the deploy, I think it's about a missing ssh key

fr/help.md
r10s: yip, we only create english translations here on github, translations are done later on transifex and then pulled regulary

en/help.md
```diff
@@ -275,6 +275,16 @@ or in the sidebar with a right click (in the desktop client).
...
 
+### What is a group ?
+
+A group is a list of members talking with each other. There are no roles in groups: every member has the possibility to add or remove anyone, including the creator.
+
+Groups can have an avatar and a subject. Just like members, those can be set by anyone in the group. Every member also has the possibility to set disappearing messages: messages older than the given duration starting from the moment the option is set are automatically deleted by DeltaChat. Each member can delete their own message from everyone's device but cannot delete messages of others.
+
+Because all members have the same rights, it is better to use groups among trusted peers.
+
+Groups exist only in the devices of its participants. The modification of groups happens by sending individual messages to each other members. Thanks to this the relay servers don't know who is part of a group, and don't even know what groups exist.
```
```suggestion
Groups exist only in the devices of its participants. The servers involved in transmitting the messages can not modify them. Servers can see, but typically do not store any membership information.
```

Unfortunately, we're not completely there yet that servers/relays can't know about groups :sweat_smile:

en/help.md
```diff
@@ -313,6 +323,18 @@ or in the sidebar with a right click (in the desktop client).
 - As an alternative, you can also "Mute" a group - doing so means you get all messages and 
   can still write, but are no longer notified of any new messages.
 
+### Cloning a group
+
+It can be useful to duplicate an existing group, for example to talk about something specific but different from what is happening in the existing group, with exactly the same people. To do so:
```
```suggestion
It can be useful to duplicate an existing group, for example to talk about something specific but different from what is happening in the existing group, or to exclude an abusive person without them noticing. To do so:
```

Thanks for documenting this feature :) While we're at it, it makes sense to spread this as the recommended workflow to avoid trolls.

en/help.md
```diff
@@ -313,6 +323,18 @@ or in the sidebar with a right click (in the desktop client).
...
+
+- On desktop, right-click on the group. On iOS/Android, long-press on the group
+
+- Select "Clone Chat"
+
+- Enter the new subject and optionally a new avatar
```
```suggestion
- Enter the new subject and optionally a new avatar or remove/add members
```

This is also possible when cloning groups :)

en/help.md
```diff
@@ -313,6 +323,18 @@ or in the sidebar with a right click (in the desktop client).
...
+
+The new created group now exists and it has no relationship with the previous one: members added/removed in one are not added nor removed in the other, subjects can be set independently, etc...
```
```suggestion
The new created group is created when you send the first message. It has no relationship with the previous one: members added/removed in one are not added nor removed in the other, subjects can be set independently, you start with an empty message history, etc...
```

And one last suggestion :)

missytake (approved):
Thanks for adding this missing piece of documentation! I have one correction and some small additions.

en/help.md
missytake: But maybe we can strike the whole paragraph - it goes deep into the details, and we actually want to keep the docs here as short as possible, and as useful as possible for the average first user. And it requires knowing about servers, which opens up many complex topics which we will just confuse most users.

rakoo: Yeah that paragraph can disappear, or it can go in another subsection, at the end of the "Groups" section. Actually it's a discussion on the technicalities (you know what they are) that started this conversation, so maybe a special separated section with its own details makes sense ?

rakoo: Also, I'm surprised servers actually see this, what a shame :(

missytake: Ah I'm misinformed myself - just learned that a stored message doesn't contain member lists and group ID anymore. Servers can of course try and record envelope metadata, that's not completely impossible, but it would require writing and running malicious code on the server.

But you see, we're getting more into intricate details than we need to for a simply FAQ entry. Let's strike the paragraph.

rakoo: Agreed

r10s:
pushed commits of this PR manually to main, after doing some streamlining wrt typos, wording, language etc.

thanks a lot again!

PR #1202, opened 2025-11-17
feat: add FreeBSD install command

nicodh:


r10s (approved):

Simon-Laux:
Doesn't this deserve it's own section like ubuntu touch? or is the user-agent of browsers on FreeBSD the same as in Linux, so not distinguishable?

CC @feld

PR #1204, opened 2025-11-19
"add second device" works with clipboard as well

r10s:
briefly mention the clipboard option,
no need to go into all details and click paths,
it is complex, if one does not have an idea how clipboard etc. works, ppl anyway need help.

doing transfer over clipboard is preferrable to transferring backup files - it is easier, and usually more secure.

it is still fine to have that at the pretty end,
other issues are still more important,
surely not all towers have a cam, but many do, and most notebooks as well. the bias in our circles may be different, of course :)

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -527,7 +527,12 @@ One device is not needed for the other to work.
 - Delta Chat is **already logged in** on the destination device?
   You can use multiple profiles per device, just [add another profile](#multiple-accounts)
 
-- If you still have problems or if you **cannot scan a QR code**
+- **No camera?** on the first device, click "Copy to Clipboard".
+  Do not close the screen.
+  Transfer the clipboard content to the second device (e.g. by another chat or file transfer),
+  and paste it there from "Add as Second Device / Paste from Clipboard"
+
+- If you still have problems
   try the **manual transfer** described below
 
 


en/help.md
```diff
@@ -527,7 +527,12 @@ One device is not needed for the other to work.
 - Delta Chat is **already logged in** on the destination device?
   You can use multiple profiles per device, just [add another profile](#multiple-accounts)
 
-- If you still have problems or if you **cannot scan a QR code**
+- **No camera?** on the first device, click "Copy to Clipboard".
+  Do not close the screen.
```
Simon-Laux, line 531: I believe on IOS you have about one minute to go back to it when sending the copied code with another app, before it gets frozen. (not sure if the second device setup continues as normal after being woken up from being frozen)

You have to switch apps on mobile[^1], so just "Do not close the screen" may be confusing on its own on mobile?
Footnotes


[^1]: unless you have some magic device sync that syncs your clipboard between devices, which I would not expect users to have ↩

r10s: good point, that's indeed tricky - you have only 20 seconds, in tests, sometimes more, but one can safely assume only that.

also on android, it is not that clear that the app lives long enough in background, btw (tho the situation is usually much better there)

a "share-to" will improve the situation, as the selection of the destination would not push the app to background, but still, user may check other apps etc.

need to think over a good wording

r10s:
k, closing this PR for now in favour of https://github.com/deltachat/deltachat-ios/issues/2905

it was not documented for years for android/desktop, so it can wait some more time also there

also, there is already a working (also complex, tho) alternative with "manual transfer"

PR #1213, opened 2025-11-26
fix i18n redirects

j4n:
Test if copying chatmail.html to fr/ and removing fr/chatmail.md will work with the pipeline.


j4n (commented):
Should work

r10s (approved):

r10s:
thanks a lot merged this PR as well as https://github.com/deltachat/deltachat-pages/pull/1212 - the latter to delete the pages also from translations

PR #1216, opened 2025-11-29
remove outdated entry about address-book access

adbenitez:
- [x] android: https://github.com/deltachat/deltachat-android/pull/4074
- [x] ios: https://github.com/deltachat/deltachat-ios/pull/2926

we probably should wait until the apps without the permission are released before merging this


r10s:
it is released now

PR #1226, opened 2025-12-11
what are relays?

r10s:
this PR adds a little section to answer the question "what are relays?" and give some high-level overview of the options, matching also the level of detail used elsewhere in the help.

this will also go to the offline help - tho advanced, it should at least give the non-nerd an idea what that is - and that they do not need to care about it

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -574,6 +574,34 @@ You can find more information
 and give feedback in the [Forum](https://support.delta.chat).
 
 
+### What are Relays?
+
+Relays are used to temporarily hold messages in case your devices is offline.
+Relays are cheap and dummy servers,
+that do not store data as group states, your name or avatar -
+all that exist only on your device.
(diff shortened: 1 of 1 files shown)


adbenitez:
it probably makes sense to point to this section from all other sites that link "chatmail relays" to the chatmail.at page, also there is still occurrences of the old link https://delta.chat/chatmail, better point to internal description and from them to chatmail.at

adbenitez:
also to revise and replace "chatmail servers" with "chatmail relays"

adbenitez (approved):

r10s:
yeah, i would also prefer internal linking, however, where this is most pressing, is the very first paragraph "What is Delta Chat?" - as this might be the very first thing ppl read about delta chat (installed from app store, by recommendation of friend&family), this is _by far_ too technical

however, this can be done in a subsequent PR, same as possible other adaptions

PR #1232, opened 2025-12-16
clarify the meaning of the ticks

### LINKED ISSUE #1230: FAQ entry for two ticks sounds like delivery receipts, but it is read receipts
https://delta.chat/en/help#what-do-the-ticks-shown-beside-outgoing-messages-mean says that "mean that at least one recipient’s device reported back to having received the message":
Image

Users read this as if Delta Chat has delivery receipts, similar to [Signal's "delivered" two empty checkmarks](https://support.signal.org/hc/en-us/articles/360007320751-How-do-I-know-if-my-message-was-delivered-or-read):
https://fedi.anarchy.moe/notice/B1G0wgwT63aXUGDnCy

FAQ should be more clear that Delta Chat does not do delivery receipts which are the source of problems described in https://arxiv.org/abs/2411.11194

r10s:
esp. remove the "delivery" confusion

moreover, say, what it means to groups and remove the "relativizations", as elsewhere, we want to have clearer, shorter descriptions :)

closes https://github.com/deltachat/deltachat-pages/issues/1230

cmp https://faq.whatsapp.com/665923838265756/?helpref=search&cms_platform=web
cmp https://support.signal.org/hc/en-us/articles/360007320751-How-do-I-know-if-my-message-was-delivered-or-read

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -185,8 +185,8 @@ and others will as well not always see that you are "online".
 
 In [groups](#groups) the second tick means that at least one member has reported back having read the message.
 
-Both you and your contacts may have disabled read receipts at **Settings → Chats → Read Receipts**,
-in this case you will not get the second tick.
+You will only get the second tick if both you and one of the recipients who read the message
+has the setting **Settings → Chats → Read Receipts** enabled.
 
 
 ### Correct typos and delete messages after sending {#edit}
--- a/en/help.md
+++ b/en/help.md
@@ -178,17 +178,15 @@ and others will as well not always see that you are "online".
 ### What do the ticks shown beside outgoing messages mean?
 
 - **One tick** <img style="vertical-align:middle; width:1.5em; margin:1px" src="../assets/help/tick1.png" alt=""/>
-  means that the message was sent successfully to your provider.
+  means that the message was sent successfully to the [relay](#relays).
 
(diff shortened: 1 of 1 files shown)


Simon-Laux (approved):

en/help.md
```diff
@@ -176,17 +176,15 @@ and others will as well not always see that you are "online".
...
-  reported back to having received the message.
+  indicate your contact has read the message.
 
-- Recipients may have disabled read-receipts,
-  so even if you see only one tick, the message may have been read.
+In [groups](#groups) the second tick means that at least one member has reported back having read the message.
 
-- The other way round, two ticks do not automatically mean
-  that a human has read or understood the message ;)
+Both you and your contacts may have disabled read receipts at **Settings → Chats → Read Receipts**,
```
link2xt: This may be misread as meaning that you will not get the second tick only if both have disabled the setting, especially bad if translator misreads it and translates it to have unambiguously wrong meaning.

link2xt: Something like this is unambiguous:
"You will only get the second tick if both you and one of the recipients who read the message has the setting **Settings → Chats → Read Receipts** enabled"

r10s: i applied your suggestion, thanks a lot!

link2xt (approved):

PR #1236, opened 2025-12-17
add paragraph about max. group size

### LINKED ISSUE #1235: Add FAQ entry for "How many members can participate in a single group?"
This is just a really frequently asked question, we had it again asked twice in the last week. I currently point to https://chaos.social/@delta/115697230575637850

Other ways the question was asked: "Can 5,000 users participate in a single group chat?", "How many users can participate in a Delta Chat group chat at the moment?", "I am wondering how good large groups are working in Delta Chat."

Other answers:
"no strict limit, but more than 150 is dis-recommended.  Please note there is no group-moderation. If the group goes haywire because of one abuser/troll, you might have to recreate a new group, and leave that abuser out to remove the spam from the group"

"there certainly is a limit, but it's not a hard-coded limit currently."

r10s:
closes https://github.com/deltachat/deltachat-pages/issues/1235

cmp https://en.wikipedia.org/wiki/Dunbar%27s_number

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -325,6 +325,16 @@ The new group is **fully independent** from the original,
 which continues to work as before.
 
 
+### How many members can participate in a single group?
+
+There is no strict technical limit,
+but more than 150 is not recommended.
+
+As groups get larger, they can become socially unstable and may need a hierarchy -
+where Delta Chat is a private messenger for chatting with [equal rights](#groups).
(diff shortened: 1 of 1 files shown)


link2xt (approved):

link2xt:
Nice that wikipedia page is translated into many languages (31). Not exactly related, but has many links from there to similar concepts.

r10s:
> Not exactly related

yeah, i also thought that and tended to remove it as a bit unrelated. but if you find it nice as well, i am happy to leave it :)

PR #1241, opened 2025-12-19
remove serverguide from active translations

r10s:
the existing translations stay as is "frozen" for now,
but as the file is legacy it does not make sense
to have translations as "todo" for new or existing languages -
esp. as translations of the big file is a lot of work (only help is larger!) -
and esp. as these technical document usually are not translated at all.


missytake (approved):
Fully agree.

PR #1247, opened 2025-12-26
try fix: libresilient: don't crash when CacheStorage API is not available.

Simon-Laux:
Try to fix lib resilient crash by not starting it when the CacheStorage API is not available.

This pr does not implement a fallback method for extra resilience and I haven't explored if it would event be possible to implement a fallback without the CacheStorage API.
It just doesn't load the libresilient code if the required CacheStorage API is unavaible and thus does not show the crash/error.

### DIFF
--- a/libresilient.js
+++ b/libresilient.js
@@ -8,8 +8,9 @@
 
 // check if the browser implements ServiceWorkers API
 if ('serviceWorker' in navigator) {
-    
-    if (navigator.serviceWorker.controller) {
+    if (!caches) {
+      console.error('LibResilient: CacheStorage API is not available.')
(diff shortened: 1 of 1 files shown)


link2xt (approved):

libresilient.js
```diff
@@ -8,8 +8,9 @@
 
 // check if the browser implements ServiceWorkers API
 if ('serviceWorker' in navigator) {
-    
-    if (navigator.serviceWorker.controller) {
+    if (!caches) {
```
link2xt: This should probably be `typeof caches === 'undefined'`. The error in lockdown mode is a `ReferenceError: Can't find variable: caches`.

PR #1248, opened 2025-12-26
Add openpgp event

link2xt:


r10s (approved):
maybe ppl learn that 2 hours learning sq takes longer than installing delta in 1 minute :)

link2xt:
Yes, i'm also not sure it's very interesting to add everything pgp-related.
More interesting is to add iroh when the event is on the hub.

PR #1249, opened 2025-12-26
Remove DECT phone number

link2xt:
We don't have it registered and will not setup a phone likely.


r10s (approved):
yeah, it was set up but  never really used in the past.

tho i just packed the DECT phones, but maybe someone else want to take them and experiment

PR #1252, opened 2026-01-12
Correct typo from 'Rekay' to 'Relay' in help.md

morta5:


r10s:
thanks for the hint!

as the filed edited in this PR is generated, it needs to be updated on [transifex](https://delta.chat/en/contribute) as well; i will do that now

PR #1254, opened 2026-01-28
add a page for removed "Move to DeltaChat folder"

r10s:
this PR adds a page that describes steps that should be done when a user has "Move to DeltaChat folder" enabled

the page should not be linked from anywhere but from the device message added  at https://github.com/chatmail/core/pull/7777

by the redirect-file, we have at least the option to translate it later, we'll see if that is worth the effort

(a similar page is needed for the "Only fetch from DeltaChat folder" option, which, however, is by far not that often used, as that was never the default and always required much more advanced user for server-side filtering)

(i was first thinking to write smth about doing the move in the classic client (it is working for me quite fine), but we should not recommend that officially, it is a call for further trouble and support effort. same if we keep one of the options in the one or other form (as discussed here and there :)

### DIFF
--- a/en/legacy-move.md
+++ b/en/legacy-move.md
@@ -23,9 +23,7 @@ To see how the change affects you,
 
 1. Open Delta Chat, go to **Settings → Advanced**
 
-2. **Disable "Move automatically to DeltaChat Folder"**,  
-   **disable "Only fetch from DeltaChat Folder"** and  
-   **set "Show Classic Emails" to "All"**.
+2. **Disable "Move automatically to DeltaChat Folder"**
 
 If you do not do this explicitly today,
 this will happen implicitly the next weeks.
@@ -43,18 +41,23 @@ the encrypted messages in the Inbox may be annoying.
 
 If so, you should change the profile as follows:
 
-1. Open Delta Chat, go to **Settings → Advanced → Relays**
+1. Open Delta Chat, go to **Settings → Advanced**
 
-2. Tap **Add Relay**
+2. **Disable "Only fetch from DeltaChat Folder"** and  
+   **set "Show Classic Emails" to "All"**
 
-3. Go to <https://chatmail.at/relays>, and select one of the relays shown there
+3. Go to **Settings → Advanced → Relays**
 
-4. **Scan the QR code** shown on the relay's website
+4. Tap **Add Relay**
+
+5. Go to <https://chatmail.at/relays>, and select one of the relays shown there
+
+6. **Scan the QR code** shown on the relay's website
 
     - **If you cannot scan,** right click or long tap the QR code,
       select "Copy Link" and paste that to the scanner window
 
-5. Tab the newly added relay so that it becomes **Default**
+7. Tab the newly added relay so that it becomes **Default**
 
 Then, continue chatting, **your contacts will learn the new relay** over time -
 until then, you are still reachable on the old email
--- a/en/legacy-move.md
+++ b/en/legacy-move.md
@@ -9,7 +9,7 @@ lang: en
 > This page is a service for users still using legacy options.
 >
 > If you did not get a warning,
-> go for for something else, [maybe play a game inside Delta Chat](help#webxdc) :)
+> go for something else, [maybe play a game inside Delta Chat](help#webxdc) :)
 
 
 ### Required Steps
--- a/en/legacy-move.md
+++ b/en/legacy-move.md
@@ -84,5 +84,5 @@ and there is no need to enter an email address - let alone a phone number.
 Finally removing the legacy "Shared Profile" options is the next logical step -
 it makes the app clearer and easier to support,
 freeing up developer mind space to to much more impactful things -
-as multi relay, channels calls.
+as multi relay, channels and calls.
 And, of course fixing bugs, making Delta Chat a truely stable experience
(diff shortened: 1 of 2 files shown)


link2xt:
For "only watch to delta chat folder" i am going to move it into transport configuration, next to IMAP port. So by default we will watch INBOX and with this setting can watch any other folder, then we don't need to tell these users to change anything and the only difference is that the code will select configured folder instead of INBOX.

en/legacy-move.md
```diff
@@ -0,0 +1,86 @@
...
+1. Open Delta Chat, go to **Settings → Advanced → Relays**
+
+2. Tap **Add Relay**
+
+3. **Scan the QR code** shown on <https://nine.testrun.org>
+
+    - **If you cannot scan,** right click or long tap [here](dcaccount:https://nine.testrun.org/new),
+      select "Copy Link" and paste that to the scanner window
+
+4. Tab the newly added relay so that it becomes **Default**
```
missytake:
```suggestion
3. Go to https://chatmail.at/relays, and select one of the relays shown there

4. **Scan the QR code** shown on the relay's website

    - **If you cannot scan,** right click or long tap the QR code,
      select "Copy Link" and paste that to the scanner window

5. Tab the newly added relay so that it becomes **Default**
```

I suggest that we recommend other relays than nine, to improve decentralization.

en/legacy-move.md
```diff
@@ -0,0 +1,88 @@
...
+which is discouraged since some time,
+the encrypted messages in the Inbox may be annoying.
+
+If so, you should switch to a dedicated chat profile as follows:
+
+1. Open Delta Chat, go to **Settings → Advanced → Relays**
+
+2. Tap **Add Relay**
+
+3. Go to https://chatmail.at/relays, and select one of the relays shown there
```
r10s:
```suggestion
3. Go to <https://chatmail.at/relays>, and select one of the relays shown there
```

en/legacy-move.md
```diff
@@ -0,0 +1,88 @@
...
+
+Therefore, by default, user get a dedicated profile since quite some time
+and there is no need to enter an email address - let alone a phone number.
+
+**Since then, the onboarding is one of the easiest around - you only need to enter a name 🎉**
+
+Finally removing the legacy "Shared Profile" options is the next logical step -
+it makes the app clearer and easier to support,
+freeing up developer mind space to to much more impactful things -
+as multi relay, channels calls.
```
Simon-Laux:
```suggestion
as multi relay, channels and calls.
```

en/legacy-move.md
```diff
@@ -0,0 +1,88 @@
...
+Also, while we keep on using email technically,
+this is nothing the average user should care about -
+similar to other messenger using https under the hood.
+
+Therefore, by default, user get a dedicated profile since quite some time
+and there is no need to enter an email address - let alone a phone number.
+
+**Since then, the onboarding is one of the easiest around - you only need to enter a name 🎉**
+
+Finally removing the legacy "Shared Profile" options is the next logical step -
```
Simon-Laux: nitpick: "Shared Profile" could be something different (like team accounts), maybe there is a better term for it, maybe sth. like "Mixed" or "Dual use".

maybe also mention what the options are that should be removed in parenthesis for additional clarity

r10s: yeah, you're right, this is not good wording.

r10s: i reworded a bit using terms already used elsewhere and avoid "Shared Profile"

en/legacy-move.md
```diff
@@ -0,0 +1,88 @@
...
+lang: en
+---
+
+
+## Use a Dedicated Chat Profile
+
+> This page is a service for users still using legacy options.
+>
+> If you did not get a warning,
+> go for for something else, [maybe play a game inside Delta Chat](help#webxdc) :)
```
link2xt:
```suggestion
> go for something else, [maybe play a game inside Delta Chat](help#webxdc) :)
```

en/legacy-move.md
```diff
@@ -0,0 +1,88 @@
...
+
+If you are using the email for chatting only,
+and not using another email app for that address beside Delta Chat,
+the following steps are not needed.
+
+**Only if you share the same email** for chatting and classic usage,
+which is discouraged since some time,
+the encrypted messages in the Inbox may be annoying.
+
+If so, you should switch to a dedicated chat profile as follows:
```
link2xt: Same as https://github.com/chatmail/core/pull/7777/commits/fb4c56fbfe4564a124c794fec43cd10e455ea215#r2765393078, we are not moving to a "dedicated chat profile", but to another relay

r10s: the result of the steps is a dedicated profile only for chatting, that does no longer share the address with other apps. this is done by adding a relay, and then removing one.

this is what was meant.

but yes, maybe confusion and both to complex.

maybe similar to https://github.com/chatmail/core/pull/7777 , it is not needed to say more than already said in the paragraph above.

```suggestion
If so:
```

however, i would be fine with smth as the following as well:

```suggestion
If so, you should change the profile as follows:
```

link2xt: > the result of the steps is a dedicated profile only for chatting, that does no longer share the address with other apps

Then it is converting existing profile to something else, not "switch". "Switch" from some profile to another means there are two profiles, the one you switch from and the one you switch to.

r10s: i added a commit

r10s:
@link2xt i added a commit to delay changing "Show Classic Email" until strictly needed; this will mitigate the risk that ppl receive unwanted mail they did not get before

adbenitez (approved):

PR #1258, opened 2026-02-02
FAQ: clarify why Deniability & Post-Compromise Security don't help

missytake:
After the FOSDEM talks I went into a bit of a rabbit hole why plausible deniability and post-compromise security are not as useful as most crypto nerds think they are. Not sure if explaining this in the FAQ makes sense, but I had to write it down anyway to cleanse my soul.

I tried to keep it short, maybe too short. I hope it's still comprehensive. Some interesting points are missing now:
- how PFS doesn't need to rotate with every message anymore if plausible deniability and post-compromise security are non-goals.
- how this helps with
  - reducing implementation complexity,
  - avoids unable-to-decrypt errors,
  - drops reliance on authenticated key exchange servers,
  - mitigates prekey-pogo attacks,
  - and drops the necessity to have one main device which needs to be online every once in a while.


r10s:
adding the only image in the whole faq  to this paragraph gives it an attention it might not deserve, we do not have that level of detail for much more important things. also, in general,  not sure if the faq is the correct place for that, maybe a forum post is a better place for that, one can link to that as well

or, we spare that for a security whitepaper, that is discussed here and there as well.

but thanks for diving into that rabbit hole :)

missytake:
Security Whitepaper! right, that's the right place. Looking forward to it.

PR #1259, opened 2026-02-05
skip 'show email' part from 'disable move to delta' instructions

r10s:
changing 'show email' will not be required to add relay on 1.41.0; reason is to not push users into potentially receiving spam until we have a 'show encrypted only' option

cc @link2xt

### DIFF
--- a/en/legacy-move.md
+++ b/en/legacy-move.md
@@ -52,7 +52,7 @@ If so, you should change the profile as follows:
     - **If you cannot scan,** right click or long tap the QR code,
       select "Copy Link" and paste that to the scanner window
 
-5. Tab the newly added relay so that it becomes **Default**
+5. Tap the newly added relay so that it becomes **Default**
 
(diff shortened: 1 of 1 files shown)


adbenitez (approved):

en/legacy-move.md
```diff
@@ -41,23 +41,18 @@ the encrypted messages in the Inbox may be annoying.
...
+3. Go to <https://chatmail.at/relays>, and select one of the relays shown there
 
-6. **Scan the QR code** shown on the relay's website
+4. **Scan the QR code** shown on the relay's website
 
     - **If you cannot scan,** right click or long tap the QR code,
       select "Copy Link" and paste that to the scanner window
 
-7. Tab the newly added relay so that it becomes **Default**
+5. Tab the newly added relay so that it becomes **Default**
```
link2xt:
```suggestion
5. Tap the newly added relay so that it becomes **Default**
```

PR #1263, opened 2026-02-15
simplify metadata paragraph

r10s:
this PR does not change facts, however, keeps them closer together.

i observed several times,
that the "random" aspect was overseen or not copy+pasted.

as being an important aspect,
eg. in comparison to messengers
using phone numbers or classic email addresses as anchors,
we should not hide that in a second paragraph that is not read at the end.

moreover, it is shorter :)


en/help.md
```diff
@@ -834,10 +834,8 @@ Instead, all group metadata is end-to-end encrypted and stored on end-user devic
 
 Servers can therefore only see:
 
-- the sender and receiver addresses
-- and the message size.
-
-By default, the addresses are randomly generated.
+- Sender and receiver addresses, randomly generated by default
```
missytake, line 837:
```suggestion
- Sender and receiver addresses, randomly generated by default
- Message date
```

As of now, we still haven't managed to fully eliminate the date from the stored messages; https://github.com/chatmail/relay/pull/849 is about to achieve that, so we could also leave it out here. But it points to the problem that we depend on the cooperation of servers to not store the correct date.

en/help.md
```diff
@@ -834,10 +834,8 @@ Instead, all group metadata is end-to-end encrypted and stored on end-user devic
 
 Servers can therefore only see:
```
missytake, line 835:
```suggestion
On the involved servers, you can therefore only find:
```

That's why I would be much more comfortable with this wording, as it makes clear that this is about *storing* metadata.

If we were talking about *seeing* metadata, we would need to include "IP addresses of users logging in", for example. To keep the IP addresses out of this, we need to presume cooperative/non-malicious servers as well. Servers could be forced to record logged-in users' IP addresses by hackers, or authorities in some legislations.

missytake (approved):
This PR improves things, still I'm not completely happy with this answer, as we imply very specific assumptions here.

I added two suggestions to be more precise:

PR #1264, opened 2026-02-16
Spelling and grammar update for dutch version of the home page

Amzd:


missytake:
Thanks for the suggestion, unfortunately we can't change the translations in this repository, but need to go through transifex: https://github.com/deltachat/deltachat-pages/?tab=readme-ov-file#translate-the-website

You can login with GitHub there, if that's not too much effort?

Amzd:
I realized that too. I am not part of the dutch team on transifex: https://app.transifex.com/delta-chat/teams/76181/nl/

r10s:
ah, cool, i just wanted to check that out, but i just noticed that @Amzd is now part of dutch team on transifex - and the strings are also already corrected - will pull them now

Amzd:
yep i fixed them on transifex and then ran out of steam, thanks for picking it up :)

PR #1262, opened 2026-02-13
fix: rewrite verify downloads for desktop

nicodh:
One part of resolving https://github.com/deltachat/deltachat-desktop/issues/6030

We need to adapt the signature.asc file that is generated and uploaded by desktop-builder before we can publish this.

The idea is to have the PGP key together with the sha5 checksums in one file so it can be verified in one command

```bash
gpg --decrypt signature.asc | shasum -a 512 --ignore-missing -c -
```
sha1 checksums will be removed

The file signature.asc will contain something like

```
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA512

b67e3e0fde06a1a98631a5a4a32221de879e28029542726bc8646418ba90db2b681b46130b9cb955e3b470a7a007c6d067da3a317cafc6d3d13dc28d242cf0be  DeltaChat-2.35.0-Portable.x64.exe
5a83e282dfa98908a0dadd19260405c8093dd2a3b0c451cc281f05034bc6900751b8b86cb1bfd68d5ff2f14373f31f2426213c7f615720853d509930fa4ef9b1  DeltaChat-2.35.0-Setup.x64.exe
34bebfba5d83a10a0ebb2ded889eeedd2b22f9fbf845fa23f7ef1997acb12991c2ef74a4025ed599dc38c7f52d7413fcf586e9fb90a9deebdd8f06caf504faab  DeltaChat-2.35.0-arm64.AppImage
etc.
-----BEGIN PGP SIGNATURE-----

wnUEARYKAB0WIQRjzR+BW6VgUYN2mZxibibIFpUTCAUCaUqrnAAKCRBibibIFpUT
CFWSAQDwBJc0jzOsyy/LKIFpiq1kPz9GwBttHA9QTotaz0qT3gEAwFTHuwdLQo7e
AvoE4W7qEwqKvzRSYYwWJ8Unadg2lQ8=
=M80U
-----END PGP SIGNATURE-----
```

PR #1265, opened 2026-02-21
improve desktop download verification

r10s:
this PR was created after trying to read and understand https://github.com/deltachat/deltachat-pages/pull/1262 , and comes with the following improvements:

- simple 1-2-3 steps, for ppl just want to get the verification done
- keep things together
- much shorter
- less "if" and "maybe" and distracting alternatives
- explain what the disadvantage is of not using gpg/rsop
- troubleshooting at the end
- visually less cluttered, matching appearance of the page elsewhere

in general, i did not check if the used method is "best practise", but sorted it only.

some things out of scope of this PR, but maybe can be targeted later

- i think, the name of `signature.asc` is wrong. it is no signature. it is a file listing checksums that is signed, i was a bit confused by that, `checksums.txt` would also make the whole flow much better to understand.

- it is unfortunate, that in the filenames and paths, the version is sometimes prefixed by `v` , sometimes not.

- why is the extension party `.asc.txt`? - in pgp-land, usually, it is only `.asc`

- the instructions inside `signature.asc` can be removed, they only will get outdated and broken

targets https://github.com/deltachat/deltachat-pages/pull/1262#issue-3936408354

### DIFF
--- a/en/verify-downloads.md
+++ b/en/verify-downloads.md
@@ -28,7 +28,7 @@ To print the SHA256 fingerprints of the APK signing certificate you can use eg.
    `<VERSION>` needs to be replaced by the version number, eg. `2.33.0`
 
    ```
-   wget https://download.delta.chat/desktop/v<VERSION>/signature.asc
+   wget https://download.delta.chat/desktop/v<VERSION>/signed-checksums.txt
    wget https://delta.chat/assets/deltachat_certificate.asc.txt
    gpg --import deltachat_certificate.asc.txt
    ```
@@ -38,7 +38,7 @@ To print the SHA256 fingerprints of the APK signing certificate you can use eg.
 3. **Verify and check results**
 
    ```
-   gpg --decrypt signature.asc | shasum -a 512 --ignore-missing -c -
+   gpg --decrypt signed-checksums.txt | shasum -a 512 --ignore-missing -c -
    ```
 
    Expected output:
@@ -55,6 +55,6 @@ To print the SHA256 fingerprints of the APK signing certificate you can use eg.
    The warning is normal as you have not explicitly trusted the key.
 
 If gpg is broken on your system, you can use  
-`cat signature.asc | rsop inline-verify deltachat_certificate.asc.txt` or  
-`cat signature.asc | grep deltachat | shasum -a 512 --ignore-missing -c -` -
+`cat signed-checksums.txt | rsop inline-verify deltachat_certificate.asc.txt` or  
+`cat signed-checksums.txt | grep deltachat | shasum -a 512 --ignore-missing -c -` -
 note, that the latter checks integrity but _not_ the developer's key.
--- a/en/verify-downloads.md
+++ b/en/verify-downloads.md
@@ -18,26 +18,43 @@ For Android, you can verify the signing certificate on the APK matches one of th
 To print the SHA256 fingerprints of the APK signing certificate you can use eg.  
 `keytool -printcert -jarfile <APK-file>`
 
+
 ## Desktop
 
-You can find detailed instructions for verification at `https://download.delta.chat/desktop/v<version>/signature.asc`
+1. Open your terminal and **change directory** to the file you want to verify, eg.  
+   `deltachat-desktop_<VERSION>_amd64.deb`
 
-The public key used for signing desktop releases is published below and on <https://keys.openpgp.org/search?q=deltachat-signing@merlinux.eu>.
+2. **Download signed checksums and import key;**
+   `<VERSION>` needs to be replaced by the version number, eg. `2.33.0`
 
-```
------BEGIN PGP PUBLIC KEY BLOCK-----
+   ```
+   wget https://download.delta.chat/desktop/v<VERSION>/signature.asc
+   wget https://delta.chat/assets/deltachat_certificate.asc.txt
+   gpg --import deltachat_certificate.asc.txt
+   ```
 
-xjMEaDSKLBYJKwYBBAHaRw8BAQdAbpU7t0wU34c3csvF60TBF+8NoH+xxew6vpG4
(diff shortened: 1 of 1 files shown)


missytake:
> some things out of scope of this PR, but maybe can be targeted later
> 
>     * i think, the name of `signature.asc` is wrong. it is no signature. it is a file listing checksums that is signed, i was a bit confused by that, `checksums.txt` would also make the whole flow much better to understand.

signed-checksums.txt maybe?

>     * the instructions inside `signature.asc` can be removed, they only will get outdated and broken

I agree, they cause warning in my output: https://github.com/deltachat/deltachat-pages/pull/1262/changes#r2807295501

missytake (approved):

nicodh:
> the instructions inside signature.asc can be removed, they only will get outdated and broken

As described in the comment of #1262 there are (or will be) no instructions anymore in that file.

I agree that renaming makes sense. The names were based on the existing names. 

I propose to call it **signed-sha512-checksums.txt** which would express both, the content and the fact that this is a signed file.

We need to check if the public key file extension can be renamed to .asc instead .asc.txt - maybe we need to adapt webserver configuration to allow files with that ending to be downloaded

nicodh:
We should also consider putting a 000-HOWTO-VERIFY.txt or similar into the root of the download page which just holds a snippet:

_How to verify Delta Chat Desktop downloads (PGP signature + SHA512 checksums):
https://delta.chat/verify-downloads_

nicodh:
Should we also mention the alternate ways using rsop and/or verify a download with the corresponding .sig file in the signatures/ dir?

r10s:
as discussed, it is better to keep the name short, @missytake's suggestion of `signed-checksums.txt` is a good compromise then - examples will be shorter, and we can change algos without changing the filename. while the filename still "says what it is".

for the `000-HOWTO-VERIFY.txt`, not sure if that is really important, but if we go for it, i would not add any technical details to the file. also the `000-` is a bit weird and raises questions where it should not. i know the idea is to have it top, but a simple `HOWTO-VERIFY.txt` is visible enough for the target group - esp. as we do not want to have tons of old releases, but only the last three or so

missytake:
> We need to check if the public key file extension can be renamed to .asc instead .asc.txt - maybe we need to adapt webserver configuration to allow files with that ending to be downloaded

I looked at it and it shouldn't be a problem.

r10s:
@nicodh feel free to push the new filenames directly to this PR once you know, what you're actually using. or ping me and i can do it

r10s:
another minor: the .asc files here and on keys.openpgp are not the same binary-wise, created a PR: https://github.com/deltachat/deltachat-pages/pull/1266

en/verify-downloads.md
```diff
@@ -18,26 +18,43 @@ For Android, you can verify the signing certificate on the APK matches one of th
...
 
-You can find detailed instructions for verification at `https://download.delta.chat/desktop/v<version>/signature.asc`
+1. Open your terminal and **change directory** to the file you want to verify, eg.  
+   `deltachat-desktop_<VERSION>_amd64.deb`
+
+2. **Download signed checksums and import key;**
+   `<VERSION>` needs to be replaced by the version number, eg. `2.33.0`
+
+   ```
+   wget https://download.delta.chat/desktop/v<VERSION>/signature.asc
```
nicodh:
```suggestion
   wget https://download.delta.chat/desktop/v<VERSION>/signed-checksums.txt
```

en/verify-downloads.md
```diff
@@ -18,26 +18,43 @@ For Android, you can verify the signing certificate on the APK matches one of th
...
+   wget https://delta.chat/assets/deltachat_certificate.asc.txt
+   gpg --import deltachat_certificate.asc.txt
+   ```
+
+   The key is also available at [keys.openpgp.org](https://keys.openpgp.org/search?q=deltachat-signing@merlinux.eu)
+
+3. **Verify and check results**
+
+   ```
+   gpg --decrypt signature.asc | shasum -a 512 --ignore-missing -c -
```
nicodh:
```suggestion
   gpg --decrypt signed-checksums.txt | shasum -a 512 --ignore-missing -c -
```

en/verify-downloads.md
```diff
@@ -18,26 +18,43 @@ For Android, you can verify the signing certificate on the APK matches one of th
...
-=isjO
------END PGP PUBLIC KEY BLOCK-----
-```
+   Make sure the fingerprint matches and that the file you want to verify is listed.
+   The warning is normal as you have not explicitly trusted the key.
 
-Download: [deltachat_certificate.asc.txt](../assets/deltachat_certificate.asc.txt)
\ No newline at end of file
+If gpg is broken on your system, you can use  
+`cat signature.asc | rsop inline-verify deltachat_certificate.asc.txt` or
```
nicodh:
```suggestion
`cat signed-checksums.txt | rsop inline-verify deltachat_certificate.asc.txt` or  
```

en/verify-downloads.md
```diff
@@ -18,26 +18,43 @@ For Android, you can verify the signing certificate on the APK matches one of th
...
------END PGP PUBLIC KEY BLOCK-----
-```
+   Make sure the fingerprint matches and that the file you want to verify is listed.
+   The warning is normal as you have not explicitly trusted the key.
 
-Download: [deltachat_certificate.asc.txt](../assets/deltachat_certificate.asc.txt)
\ No newline at end of file
+If gpg is broken on your system, you can use  
+`cat signature.asc | rsop inline-verify deltachat_certificate.asc.txt` or  
+`cat signature.asc | grep deltachat | shasum -a 512 --ignore-missing -c -` -
```
nicodh:
```suggestion
`cat signed-checksums.txt | grep deltachat | shasum -a 512 --ignore-missing -c -` -
```

nicodh (changes requested):
I just adapted the name to the current state

nicodh (approved):

PR #1266, opened 2026-02-23
fix: key on openpgp should binary-wise the same as here

r10s:
i tried a simple `cmp` and that failed.
sure, only because of the comments,
but one has to investigate deeper to check that
and at a first glance it looks bad.

unless there is a good reason, the files should be identical.

i would not care about removing the comments alltogether, however, changing the file on out site is easier - and maybe also makes sense to leave the "official" key untouched

Screenshot 2026-02-23 at 13 39 36

### DIFF
--- a/assets/deltachat_certificate.asc.txt
+++ b/assets/deltachat_certificate.asc.txt
@@ -1,4 +1,6 @@
 -----BEGIN PGP PUBLIC KEY BLOCK-----
+Comment: 63CD 1F81 5BA5 6051 8376  999C 626E 26C8 1695 1308
+Comment: deltachat-signing@merlinux.eu
 
 xjMEaDSKLBYJKwYBBAHaRw8BAQdAbpU7t0wU34c3csvF60TBF+8NoH+xxew6vpG4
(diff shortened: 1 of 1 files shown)


missytake (approved):
Thanks for looking this up! Can confirm. I wonder whether keys.openpgp.org added these comments automatically 🙄

nicodh (approved):
Makes sense!

r10s:
> I wonder whether keys.openpgp.org added these comments automatically 🙄

seems so. did a random lookup at https://keys.openpgp.org/search?q=security%40sequoia-pgp.org - and that indeed show the same lines.

if so, i find that questionable of keys.openpgp, for mentioned reasons. if one is not deep into pgp - or gpg is broken, having weird lock errors seem to be quite normal, at least on mac - using `cmp` is not far fetched.

adding random lines here does not improve already bad pgp ux, esp. for casual users, esp. as the downloaded files do not look like text files but more as a "solid something" 

but yeah, i think we merge this PR and call it a day. this at least fixes the issue for Delta Chat :)

PR #1269, opened 2026-02-26
Fixed typos and translated some passages

codiflow:


missytake:
Thanks for the fixes! Unfortunately we can't merge them here - we do our translations on transifex, if we merge them here they will be overwritten by the next `tx pull`.

Can you go over to https://www.transifex.com/delta-chat/delta-chat-pages/ and make the changes there? You can sign in with your GitHub account. I think you need to request to be added to the language or so, let me know if there is a hangup.

codiflow:
Sure think, will do 🙏

codiflow:
@missytake Unfortunately I only get a 404 when following your URL. The same happens when I'm logged into Transifex and follow the URL. I also don't find a way to "search" for Delta Chat or some other organisation.

Maybe you can add me by my Transifex username? It is `sweetgood`

Thanks 🙏

missytake:
Oh sorry for the broken link - does https://explore.transifex.com/delta-chat/delta-chat-pages/ work?

Unfortunately I couldn't find your username:

image

codiflow:
Thanks for the new link – this time it worked 😎

Seems like my Transifex account is tied to Wekan ... no idea why it is not possible to use this account with other projects. Will create a new one and try to join the organisation then.

Thanks for the help – will report back if I need anything.

r10s:
@codiflow thanks a LOT for the details review and corrections! this is really welcome.

i am often doing a good part of the german translations usually, but more as a side thing, i am a developer, and these days caring mostly about overall UI/UX, consistency etc. parts of your corrections are mistakes by the change from "sie" to "du" - but other parts, these are grammar rules that are not that deep in my mind :)

also you simplification of some phrases are great, doing shorter sentences here and there.

please let me know, if it works with your transifex account and what your handle is there, i would like to write you a PM there and invite you to our internal translators group. 

for now, for this PR i picked up all your suggestions and added them to the existing translations, this is also already available at https://delta.chat/de/help ([corresponding changes](https://github.com/deltachat/deltachat-pages/pull/1273/changes#diff-9ed2d0df9b95dfd1d4764652eecbd21eb90db462b3a19abe4316504830be4ffc))

codiflow:
@r10s Thank you too 🙏

I accepted one of the Transifex invitations and will translate more strings as soon as I find the time. I just didn't had enough time for this PR so I just stopped in the middle of the translations and just fixed some bigger issues 🤪

codiflow:
@r10s I took my time and went through all 306 strings in Transifex regarding the help page. Unfortunately I could only save "suggestions" so I think you might have to get through all of them again if you want them to be live 🤪🙏

PR #1271, opened 2026-02-27
docs: restructure and rewrite legacy-move.md

hpk42:
Shift from "Legacy" to "Dedicated vs. Shared": The term "legacy" has been removed in favor of describing the technical state: Shared addresses (shared with classic email clients) vs. Dedicated chat addresses. This is more descriptive and avoids the stigmatizing "legacy" label while accurately reflecting the technical evolution of the project.

Seamless Transition: Instead of suggesting users simply "stop" using their old address, the guide now provides a clear path to a split setup. Two Profiles, One App: Instructions are provided to create a dedicated chat profilewhile keeping the original email address as a separate, second profile for classic email usage. This ensures a "best of both worlds" experience: reliable, clutter-free chatting and full classic email access. While making the lifes for developers easier, and easing the introduction of other requested or desired features.

### DIFF
--- a/en/legacy-move.md
+++ b/en/legacy-move.md
@@ -4,83 +4,80 @@ lang: en
 ---
 
 
-## Use a Dedicated Chat Profile
-
-> This page is a service for users still using legacy options.
+> This page is for users who share their chat address with a classic email app or webmail provider.
 >
-> If you did not get a warning,
-> go for something else, [maybe play a game inside Delta Chat](help#webxdc) :)
+> If you didn't receive a warning leading you here, feel free to ignore this page and [maybe play a game](help#webxdc). :)
 
 
-### Required Steps
+## Upcoming Change: March 2026
 
-The option "Move automatically to DeltaChat Folder"
-as well as other legacy options
-will be removed in the next weeks.
+In **March 2026**, the option **"Move automatically to DeltaChat Folder"** will be removed.
+If you are not using other email apps to read your normal inbox, this change will likely not affect you.
 
-To see how the change affects you,
-**already today**, do the following steps:
+Otherwise, encrypted chat messages may appear in your other email apps.
+To prevent this, we recommend using two separate profiles within Delta Chat:
+1.  **A Dedicated Chat Profile**: For seamless chatting with your existing contacts.
+2.  **A Classic Email Profile (Optional)**: To read your normal inbox and communicate with non-chat contacts.
 
-1. Open Delta Chat, go to **Settings → Advanced**
+With this setup, none of your contacts need to change anything.
+Your encrypted chat contacts will simply continue chatting, and you can still read and reply to emails from Delta Chat.
+Here are the two steps we recommend to take.
 
-2. **Disable "Move automatically to DeltaChat Folder"**
+### Step 1: Move towards Dedicated Chat Profile
 
-If you do not do this explicitly today,
-this will happen implicitly the next weeks.
+1.  **Add a relay**: Go to **Settings → Advanced → Relays** and tap **Add Relay**.
 
+2.  **Scan a new QR code**: Visit <https://chatmail.at/relays>, select any relay, and scan its QR code.
 
-### Optional Steps
+3.  **Set as default**: Tap the new relay and select **Default**.
 
-If you are using the email address for chatting only,
-and not using another email app for that address beside Delta Chat,
-the following steps are not needed.
+Your contacts will automatically learn your new dedicated address when they see your messages, reactions, or read receipts.
 
-**Only if you share the same email address** for chatting and classic usage,
-which is discouraged since some time,
-the encrypted messages in the Inbox may be annoying.
+### Step 2: Keep using Delta Chat for your classic email (Optional)
 
-If so, you should change the profile as follows:
+To continue receiving normal emails on your old address:
 
-1. Open Delta Chat, go to **Settings → Advanced → Relays**
+- **Mobile**: Tap your avatar and select **Add Profile**.
 
-2. Tap **Add Relay**
+- **Desktop**: Click the **"+"** button in the sidebar.
 
-3. Go to <https://chatmail.at/relays>, and select one of the relays shown there
-
-4. **Scan the QR code** shown on the relay's website
-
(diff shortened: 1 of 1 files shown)


en/legacy-move.md
```diff
@@ -4,83 +4,80 @@ lang: en
...
+
+
+## Upcoming Change: March 2026
+
+In **March 2026**, the option **"Move automatically to DeltaChat Folder"** will be removed.
+If you are not using other email apps to read your normal inbox, this change will likely not affect you.
 
+Otherwise, encrypted chat messages may appear in your other email apps.
+To prevent this, we recommend using two separate profiles within Delta Chat:
+1.  **A Dedicated Chat Profile**: For seamless chatting with your existing contacts.
```
link2xt, line 19: What does saying "seamless chatting" mean? Is it different from just chatting?

en/legacy-move.md
```diff
@@ -4,83 +4,80 @@ lang: en
...
 >
-> If you did not get a warning,
-> go for something else, [maybe play a game inside Delta Chat](help#webxdc) :)
+> If you didn't receive a warning leading you here, feel free to ignore this page and [maybe play a game](help#webxdc). :)
+
+
+## Upcoming Change: March 2026
+
+In **March 2026**, the option **"Move automatically to DeltaChat Folder"** will be removed.
+If you are not using other email apps to read your normal inbox, this change will likely not affect you.
```
link2xt, line 15: At this point user has only one inbox, so why say "normal inbox"? There is also "normal inbox" below. Is there some other inbox?

en/legacy-move.md
```diff
@@ -4,83 +4,80 @@ lang: en
...
 
-### Required Steps
+With this setup, none of your contacts need to change anything.
+Your encrypted chat contacts will simply continue chatting, and you can still read and reply to emails from Delta Chat.
+Here are the two steps we recommend to take.
 
-The option "Move automatically to DeltaChat Folder"
-as well as other legacy options
-will be removed in the next weeks.
+### Step 1: Move towards Dedicated Chat Profile
```
link2xt, line 26: The title is unclear, for me "Move towards Dedicated Chat Profile" at first sounds like I have to move to another profile, but the profile here stays the same.

en/legacy-move.md
```diff
@@ -4,83 +4,80 @@ lang: en
...
+2.  **Scan a new QR code**: Visit <https://chatmail.at/relays>, select any relay, and scan its QR code.
 
-2. **Disable "Move automatically to DeltaChat Folder"**
+3.  **Set as default**: Tap the new relay and select **Default**.
 
-If you do not do this explicitly today,
-this will happen implicitly the next weeks.
+Your contacts will automatically learn your new dedicated address when they see your messages, reactions, or read receipts.
 
+### Step 2: Keep using Delta Chat for your classic email (Optional)
```
link2xt, line 36: Profile set up here will receive undecryptable messages until all contacts learn the new address. I think we really want to avoid this, so this should not be set up until all contacts learn the new address, but you never know when it happens.

r10s (approved):
in step 1, maybe explicitly say that "Move automatically to DeltaChat Folder" needs to be disabled

otherwise, lgtm, and it is indeed more helpful for ppl who want continue using delta for their mail.

link2xt:
> in step 1, maybe explicitly say that "Move automatically to DeltaChat Folder" needs to be disabled

The whole "test things today" should be above the steps IMO.

en/legacy-move.md
```diff
@@ -4,83 +4,80 @@ lang: en
...
-and there is no need to enter an email address - let alone a phone number.
+- **Complexity**: It hinders the development of major features like multi-server support.
 
-**Since then, the onboarding is one of the easiest around - you only need to enter a name 🎉**
+We recognize that several people who started using Delta Chat in 2024 
+or earlier appreciated using a single mailbox for "everything."
+The hundred thousand(s) of new Delta Chat users in 2025 and 2026 
+rarely know that they are using email addresses. 
+Instead of continuing to invest developer hours into what is now an edge case, 
+we are focusing on providing a seamless, reliable, and secure experience
```
link2xt, line 81: "providing experience", especially "seamless", is probably just going to annoy someone who needs to figure out the settings. Saying that it makes the app easier to support and frees developer time for more important things was better.

en/legacy-move.md
adbenitez: I guess **"[...] to read your existing email account"** (or inbox) would be a bit clearer

adbenitez (approved):
already a huge improvement!

en/legacy-move.md
```diff
@@ -4,83 +4,80 @@ lang: en
...
+> This page is for users who share their chat address with a classic email app or webmail provider.
 >
-> If you did not get a warning,
-> go for something else, [maybe play a game inside Delta Chat](help#webxdc) :)
+> If you didn't receive a warning leading you here, feel free to ignore this page and [maybe play a game](help#webxdc). :)
+
+
+## Upcoming Change: March 2026
+
+In **March 2026**, the option **"Move automatically to DeltaChat Folder"** will be removed.
```
link2xt, line 14: I would remove the date here, likely happens later and also depends on when you update.

link2xt:
Once this is finished would be nice to also add some footnote about dclogin or a pointer to https://support.delta.chat/t/how-to-migrate-your-profile-to-another-relay-version-2-33/4441 with detailed instructions for users who have an account on their own server and want to move it to another mailbox on their own server without having to use chatmail.

link2xt:
More users asking the same question: https://support.delta.chat/t/how-to-use-classical-e-mail-account-after-removal-of-delta-chat-folder/4834

en/legacy-move.md
hpk42: maybe "friction-less". the friction is described right in the sentence before.

hpk42:
apart from addressing the various comments and nit picks, there also is the issue that the upcoming multi-relay releases might render the current advise outdated.  I won't get to this the next days, so i am closing this effort for now.  If somebody else wants to take it up, be my guest.

r10s:
let's wait a few days until ["unpublished"](https://github.com/chatmail/core/issues/7980) is there

PR #1272, opened 2026-02-28
Update required Ubuntu Touch version

lk108:
Officially, DeltaTouch does not support Ubuntu Touch 16.04 anymore. From time to time, releases are still backported to UT 16.04, but the store does not accept uploads for this UT version anymore, so they are available from https://codeberg.org/lk108/deltatouch/releases only.

### DIFF
--- a/_includes/download-boxes.html
+++ b/_includes/download-boxes.html
@@ -189,7 +189,7 @@
             </div>
             <div class="descr">
                 {%if tx.download.minimum != ""%}{{ tx.download.minimum }}{%else%}{{ txEn.download.minimum }}{%endif%}:
-                Ubuntu Touch 16.04
+                Ubuntu Touch 20.04
             </div>
             <div class="descr">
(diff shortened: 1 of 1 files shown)


lk108:
I think the first part of the user agent to OS mapping condition for UT in [assets/js/download-page2.js](https://github.com/deltachat/deltachat-pages/blob/09e471d279baf0fab329506b122460e9abc40157/assets/js/download-page2.js#L16) relates to 16.04 only:
```
(uA.includes('Ubuntu; Mobile') || uA.includes('Ubuntu; Tablet')
```
so maybe this first part can be removed?

r10s:
> I think the first part of the user agent to OS mapping condition for UT [...] relates to 16.04 only:

it still seems better than to act as UT than android/ios/desktop in that case 🤷‍♂️ esp if it is not 100% clear, removal can do easily harm. i would change that only if there are complains or so

r10s (approved):

PR #1274, opened 2026-03-08
Use `contentinfo` ARIA role for footer

veyndan:
The accessibility tree of the homepage currently identifies all major landmarks as `generic`, giving no structural indication to the page. As a first step to giving semantic, accessible structure to the page, we can change the landmark ARIA role of the footer from `generic` to `contentinfo`. By using `<footer>` directly, [we get this accessibility information for free](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/footer#technical_summary).

Depending on the appetite for this change, I can create follow-up PRs to similarly improve the remaining major landmarks via the usage of semantic tags.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -76,7 +76,7 @@
 
 </div>
 
-<div id="footer">
+<footer>
 <p>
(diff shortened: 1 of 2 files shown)


missytake (approved):
Looks good, checked some of the compatible versions and they have been around for 10+ years. Accessibility is always good, thanks!

PR #1276, opened 2026-03-09
improve 'media' paragraph

r10s:
- move the 'media' paragraph a bit down. it is not the most important thing when starting with delta.

- reword the question from _if_ to _how_

- mention the three options we offer

cmp. https://github.com/deltachat/deltachat-android/pull/4259 for more details, discussion and reasoning


wchen342 (approved):
Looks good

PR #1279, opened 2026-03-09
Use `main` ARIA role for main content

veyndan:
The accessibility tree of all pages currently identifies the `<div id="content">` landmark as `generic`, giving no structural indication that this is the main content of the page. By using `<main>`, [we get the correct ARIA role of `main` for the main content for free](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/main#landmark).

This is a follow up to https://github.com/deltachat/deltachat-pages/pull/1274.

### DIFF
--- a/39c3.html
+++ b/39c3.html
@@ -110,7 +110,7 @@
         padding-inline: 0.4em;
         padding-bottom: 0.4em;
       }
-      #content {
+      main {
         max-width: 800px;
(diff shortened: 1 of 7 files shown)


missytake (approved):
Looks good :) also tested `create-local-help.py`, output looked the same as on main.

PR #1280, opened 2026-03-10
Use `banner` and `navigation` ARIA roles for nav bar

veyndan:
We're adding two landmarks here to represent the nav bar, with the first one sectioning out that it's the header landmark, while the second one is sectioning out the navigation landmark. It just so happens that the header landmark purely consists of the navigation landmark at the moment, but that doesn't always have to be the case.

This is a follow up to https://github.com/deltachat/deltachat-pages/pull/1274.

### DIFF
--- a/404.html
+++ b/404.html
@@ -11,11 +11,13 @@
 
 <body id="top">
 
-<div id="menu">
+<header>
+<nav>
 <ul>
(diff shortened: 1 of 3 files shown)


missytake (approved):
Cool :) TIL

PR #1281, opened 2026-03-10
Stop explicitly setting font size for website

veyndan:
The default font size for most browsers is 16px already, but some people (including myself) like to toggle the default font size in the browser settings. Explicitly setting the default font size to 16px in CSS overrides that preference.

### DIFF
--- a/37c3.html
+++ b/37c3.html
@@ -8,7 +8,6 @@
     <style>
       body {
         font-family: Verdana, Arial, Helvetica, sans-serif;
-        font-size: 15px;
         line-height: 1.4em;
         color: #000;
       }
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -34,7 +34,6 @@ nav li {
 
     a, a:visited {
         color: #fff;
-        font-size: 16px;
     }
 }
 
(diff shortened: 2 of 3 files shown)


37c3.html
```diff
@@ -8,7 +8,6 @@
     <style>
       body {
         font-family: Verdana, Arial, Helvetica, sans-serif;
-        font-size: 15px;
```
veyndan, line 11: This should be the only visible difference, where when visiting this page with the default browser font size untouched, the font will be 1px larger (15px to 16px).

r10s:
probably okay, usually, we have staging preview, but that requires to create pr in the deltachat-org - you do not have access to that, right?

if you are up for a call, we can maybe do that these days. 

in any case, i will see how to get a preview, but just now afk

veyndan:
> probably okay, usually, we have staging preview, but that requires to create pr in the deltachat-org - you do not have access to that, right?

Beyond https://github.com/deltachat, I have no idea what the deltachat-org is, so presumably no :sweat_smile:

> if you are up for a call, we can maybe do that these days.

Sure thing :slightly_smiling_face:

r10s (approved):

PR #1282, opened 2026-03-10
Remove commented out CSS

veyndan:
As the CSS isn't minified, these comments are sent to every website visitor, resulting in marginally slower load times. Emphasis on the word "marginally" :joy:

### DIFF
--- a/assets/css/comments.css
+++ b/assets/css/comments.css
@@ -5,10 +5,6 @@ img.avatar{
   float: left;
   border-radius: 30%;
   padding: 15px;
-  /*margin: .2rem .5rem;*/
-  /*margin-top: 1rem;*/
-  /*margin-right: 1rem;*/
-  /*margin-bottom: 1rem;*/
 }
 
 div.comment-content{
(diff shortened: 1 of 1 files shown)


r10s (approved):

r10s:
great! thanks a lot!

when it comes to simplificatio, what we wanted to do, btw, is to get rid of .scss, we never used a significant part of it, but it is another thing to know. also, in 2026 plain .css has improved compared to 7y ago.

also, we wanted to inprove the menu, having it more mobile friendly.

however, "we" as the app devs never really found time to do it - so, if you like to do, that's very welcome! seems you are deep into html :)

ps: what we explicily do not want currently is a dark mode. while easy to add, it increases costs over time and makes simple things harder, or look bad. for not much benefit. again, most ppl work on the site are no layouters and are struggling with one layout already

veyndan:
> in 2026 plain .css has improved compared to 7y ago.

Definitely true, though I guess if we want the site to be backwards compatible with no intermediary build tool, we might have to write CSS like it's 2021 (or many years ago), so niceties like nesting would be out the window.

> ps: what we explicily do not want currently is a dark mode. while easy to add, it increases costs over time and makes simple things harder, or look bad. for not much benefit. again, most ppl work on the site are no layouters and are struggling with one layout already

That's understandable, but on the accessibility front, it's quite a big win. Happy to not poke this for now though!

On a broader note, where would be a good place for me to raise ideas about the site, specifically with maintenance (i.e., not large changes in content). For example, some linter tool like Prettier would go a long way in making it a bit easier to work with codebase (and would also make it easier to review PRs with fewer erroneous styling changes). For something like that, would GitHub Issues be the place, filed under "Bug report" or would that be one for the Community forums?

PR #1285, opened 2026-03-13
Remove unused `desktoponly` and `nodesktop` classes

veyndan:
The last usage of `desktoponly` was in b9dbd20b871aa915648c17d828682604aedf5fe1 and the last usage of `nodesktop` was in 55b59fcda111ed3c8545f3eeec126eebe315fb0d.


missytake (approved):
Good catch :)

PR #1286, opened 2026-03-14
Center-align socials

veyndan:
A slight visual hiccup, where the socials were all off-center by 9px.

|Before|After|
|---|---|
|image|image|

### DIFF
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -211,15 +211,17 @@ main table th { background-color: #eee; font-weight: bold; }
 
 main hr { border: 1px solid #ddd; margin: 2em 0; }
 
+.socials {
+    display: inline-flex;
+    gap: 12px;
+}
 
 .feed {
-  margin-left: 9px;
   padding: 0 0 0 21px;
   background: url(feed-icon.png) no-repeat 0 50%;
 }
 
 .mastodon {
-  margin-left: 9px;
   padding: 0 0 0 21px;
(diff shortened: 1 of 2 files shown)


_sass/_base.scss
```diff
@@ -211,15 +211,17 @@ main table th { background-color: #eee; font-weight: bold; }
 
 main hr { border: 1px solid #ddd; margin: 2em 0; }
 
+.socials {
+    display: inline-flex;
+    gap: 12px;
+}
 
 .feed {
-  margin-left: 9px;
```
veyndan, line 216: Just removing this line would theoretically fix the problem. However, I went with a flexbox approach, because then, when adding or removing social links, we don't need to remember to add or remove the `margin-left` from the new or old social classes. By using `gap`, we can jsut let CSS figure out when the spacing should be applied, regardless of the number of elements.

_sass/_base.scss
```diff
@@ -211,15 +211,17 @@ main table th { background-color: #eee; font-weight: bold; }
 
 main hr { border: 1px solid #ddd; margin: 2em 0; }
 
+.socials {
+    display: inline-flex;
+    gap: 12px;
```
veyndan, line 216: Had to bump the spacing from 9px to something a bit higher, as previously the spacing was actually 9px plus the width of a space character. Pushing it to 12px makes it look visually close enough.

r10s (approved):
you have good eyes!

PR #1287, opened 2026-03-14
Remove magic numbers when defining social spacing

veyndan:
The way the spacing between the social images and their text work currently work is:

1. Social images have a fixed width of 14px.
2. We add a left padding of 21px to the text.
3. The resulting visual spacing is 21px - 14px = 7px.

This means that a change in the width of a social image would break the layout. It is also a bit convoluted to understand.

By just sticking the image in HTML and specifying a `gap`, where the `gap` aligns with the spacing we visually see, things become a lot easier to reason about. Changing the width of a social image still breaks the layout, but it does so in a less surprising way.

### DIFF
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -214,16 +214,12 @@ main hr { border: 1px solid #ddd; margin: 2em 0; }
 .socials {
     display: inline-flex;
     gap: 12px;
-}
 
-.feed {
-  padding: 0 0 0 21px;
-  background: url(feed-icon.png) no-repeat 0 50%;
-}
-
-.mastodon {
-  padding: 0 0 0 21px;
-  background: url(mastodon-icon.png) no-repeat 0 50%;
+    a {
+        display: flex;
+        align-items: center;
+        gap: 8px;
+    }
 }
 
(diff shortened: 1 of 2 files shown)


_sass/_base.scss
```diff
@@ -214,16 +214,12 @@ main hr { border: 1px solid #ddd; margin: 2em 0; }
...
-  background: url(feed-icon.png) no-repeat 0 50%;
-}
 
-.mastodon {
-  padding: 0 0 0 21px;
-  background: url(mastodon-icon.png) no-repeat 0 50%;
+    a {
+        display: flex;
+        align-items: center;
+        gap: 8px;
```
veyndan, line 221: I know social ~distancing~ spacing should be 2 meters, but I think 8px looks better /s

missytake (approved):

r10s:
a minor: this broke the icon on mirrors as https://deltachat.github.io/deltachat-pages - on mirrors it is not guaranteed, that the page is in the "root folder", therefore, we prefer relative paths, wherever possible

i did a fix at https://github.com/deltachat/deltachat-pages/pull/1324

PR #1289, opened 2026-03-16
Update Mastodon logo to latest

veyndan:
Mastodon [went purple](https://blog.joinmastodon.org/2022/06/mastodon-branding-updates/) back in 2022. I took the [new icon](https://joinmastodon.org/logos/logo-purple.svg) from [their branding page](https://joinmastodon.org/branding).

Another upside of this is that we're using an SVG, so the icon looks a lot less pixelated on high-density screens (or when the font size is increased as in #1288).

(These screenshots aren't a great representation in the actual quality improvement. On my desktop monitor, the difference in image quality is night and day.)

|Before|After|
|---|---|
|image|image|


assets/css/mastodon-icon.svg
veyndan, line 1: I optimised the icon through https://www.svgviewer.dev, which saved us a fair few bytes.

missytake (approved):
Great improvement!

PR #1294, opened 2026-03-16
Remove letter spacing in headings

veyndan:
This is a very subjective one, but when dropping the letter spacing, I found the titles quite a bit easier to read. I think it's common to not have letter spacing on headers, but I guess design shouldn't necessarily be driven by normalcy. I found the readability improved everywhere, but it particularly improved for the h3 headers (e.g., those on the FAQ).


r10s (approved):
i like it, but also open to a change here, after all the years :)

so, i am fine with that

PR #1298, opened 2026-03-18
Align nav bar color with desktop

veyndan:
The main motivation for changing the color of the nav bar is to increase the contrast between the text and background. To figure out a more accessible color, I used the [nav bar color that's already used on the desktop app](https://github.com/deltachat/deltachat-desktop/blob/e02c455d3c9c6b67493bb45e83944b83df414f4a/packages/frontend/themes/light.scss#L9). The benefit of this is having a bit of design coherence between the website and its apps.

### DIFF
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -13,8 +13,7 @@ nav {
     text-align: center;
     margin-bottom: 1em;
     padding: 8px 5px;
-    background-color: #566a73;
-    background-image: linear-gradient(120deg, #77888f, #364e59);
+    background-color: #415e6b;
 }
 	
 nav ul {


r10s (approved):
well, the gradient also picks up colors from the overall style.

but i would be fine with giving the more flat approach a try, some changes after some years :)

missytake:
Hm, I liked the gradient better, but let's see if others complain. Deciding about aesthetics is hard :D

r10s:
using it for some days, tbh, i also find it better before. also it fits better overall, as eg. also some buttons pick up the gradient.

the similarity between apps and website was also good enough before - eg. iOS also does not have the bar.

so i would also tend to revert the PR, and leave these broader changes up to a later point, if needed. the increased contrast - well, i would assume that visually impaired or blind ppl tweak the css anyways, so most important thing here is that we do not override these things by our css.

EDIT: i revered the corresponding commit

veyndan:
I'd imagine that the percentage of visually impaired people that tweak the CSS is not that different from those that aren't visually impaired (i.e., a very low percentage). Even if it was an order of magnitude more, we're still talking about an extreme minority of visually impaired people that would tweak their CSS.

It's admittedly a bit harder for people to override the CSS of the header when it's a gradient, as they'd need to know whether to override the `background-image` or the `background-color`, as it depends on the capabilities of their browser. Using just a `background-color` means that there can only be one thing to override.

I didn't want to choose a random colour, so chose a colour that's used elsewhere. Definitely up for using a different colour!

Obviously up to you both though! I'm not a designer, but I've worked with design systems quite a bit as part of my day job, and one of the foundations is having a unified colour palette. I was hoping to push us in that direction a bit, but totally understand if that's not something you want to do :)

r10s:
hey @veyndan i sent you a pm to your deltachat :)

PR #1301, opened 2026-03-18
Make `<h3>` distinguishable from `<h4>`

veyndan:
On [certain pages](https://delta.chat/en/gdpr#2-processing-when-using-the-delta-chat-e-mail-messenger), when a `<h3>` is followed by a `<h4>` it's impossible to visually distinguish which is which, removing all sense of visual hierarchy. By making `<h3>` slightly bigger than `<h4>`, we can form a visual hierarchy with the headings.


_sass/_base.scss
```diff
@@ -1,7 +1,7 @@
 /* reset
 *******************************************************************************/
 
-* {margin:0;padding:0} iframe,a img,fieldset,form,table {border:0} h6,h5,h4,h3,h2,h1,caption,th,td {font-size:100%;}
```
veyndan, line 4: The reason removing `font-size: 100%` from all the headers gives us what we want is because:

- `h1` and `h2` have an explicit font size lower down in the file, so this rule is always overridden for them.
- `h3`'s default font size is 1.17em. This is slightly bigger than 100% (i.e., 1em), which is the point of the PR.
- `h4`'s default font size is 1em. We were overriding its font size unnecessarily as it was essentially the same value.
- `h5` and `h6` are unused in the website.

missytake (approved):

PR #1303, opened 2026-03-24
update android to 2.47.0

adbenitez:


r10s (approved):

r10s:
as gplay is already rolling at 1%, and fdroid is tagged as well, we should merge this as well - we should always provide the newest release that is also available elsewhere

PR #1311, opened 2026-03-31
Recommend Lix instead of Determinate Systems installer

link2xt:
Determinate Systems is pushing its own fork of Nix that is not fully open source and tries to sell
their FlakeHub through it.


link2xt:
See https://github.com/deltachat/deltachat-android/pull/3937

r10s (approved):

PR #1313, opened 2026-03-31
Update help.md

duub:
Improve step by step group creation

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -281,7 +281,7 @@ Because all members have the same rights, groups work best among **trusted frien
 
 ### Creation of a group
 
-- Select **New chat** and then **New group** from the menu in the upper right corner or hit the corresponding button on Android/iOS.
+- At the chat list, on Android or Desktop tap the (+) button, on iOS tap the top right button of a paper and a pen and choose "Create a group".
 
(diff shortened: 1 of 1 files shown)


en/help.md
```diff
@@ -281,7 +281,7 @@ Because all members have the same rights, groups work best among **trusted frien
 
 ### Creation of a group
 
-- Select **New chat** and then **New group** from the menu in the upper right corner or hit the corresponding button on Android/iOS.
+- At the chat list, on Android or Desktop tap the (+) button, on iOS tap the top right button of a paper and a pen and choose "Create a group".
```
r10s, line 284: hm,  the change is vague as well. 

- there is no option named "Create a group"
- "pen and paper", not sure, iOS user identify the icon as that and know what that means.  here vague "corresponding icon" might be clearer, the icon is usually recognized if ppl are on iOS to "create smth" (as you on iOS?), so maybe the more vague thing is even better
- it sounds as if last advice "pen and paper and chose" applies to iOS only - where you have to do the same on android/desktop

still, thanks for the push, and i also think, it can be improved. there is also the idea to use the "+" button on iOS as well, that would make communication easier, and whatsapp uses that symbol as well

duub: Ok,  Also trying to explain the iOS process without one was a mistake, my bad

PR #1317, opened 2026-04-07
Make new imageless posts keep style of other new posts

veyndan:
We don't have this case at the moment, but if there's a new post without an image, we currently would show the post in the style of the old blog posts. This is visually a bit odd, as it seems misplaced (and is easy to miss among the other imaged posts).

Here's some theoretical screenshots, assuming that the image for the "Delta Chat V2: a major ..." blog post didn't have an image:

|Before|After|
|---|---|
|image|image|

### DIFF
--- a/_includes/blog.html
+++ b/_includes/blog.html
@@ -1,12 +1,14 @@
 {% assign index = 0 %}
 {% for post in site.posts %}
 {% assign index = index | plus:1 %}
-{% if post.image and index <= 12 %}
+{% if index <= 12 %}
 <div class="post-box">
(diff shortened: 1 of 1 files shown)


r10s (approved):
makes sense for code understanding and what the gist actually is, even tho i do not think we want no-image-blogposts at the moment

veyndan:
I suppose that ideally we'd want to require that the `post.image` is present, and actually fail the renderer if that isn't true. Otherwise, we get an `img` with a blank `src` which isn't great. I don't know much about Liquid the templating engine, so no idea if this is possible or not.

PR #1318, opened 2026-04-07
Simplify HTML generation logic for blog

veyndan:
This makes the logic a bit more declarative over imperative. Namely, we're not conditionalising the showing of the `h2` any more, and the `new_post_count` variable that's replacing the `index` variable is (in spirit) a constant.


missytake (approved):
Nice, much more pleasant to read.

PR #1321, opened 2026-04-09
Use unitless numbers for `line-height` values

veyndan:
Per the [advice of MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/line-height#prefer_unitless_numbers_for_line-height_values), we should "prefer unitless numbers for line-height values". The example given is quite informative, and explains why removing the `line-height` from the `h1` and `h2` causes the text to become squashed. Now we can let the headers inherit the `line-height`, causing the `line-height` to scale relative to its own font size instead of it's parents.

### DIFF
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -152,7 +152,7 @@ p.blogdate {
 
 body {
     font-family: Arial, Verdana, Helvetica, sans-serif;
-    line-height: 1.5em;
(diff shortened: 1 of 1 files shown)


r10s (approved):

r10s:
thanks a lot for taking care of these details!

PR #1328, opened 2026-04-13
Simplify download recommendation display logic

veyndan:
Removes an unnecessary DOM node, gives the DOM node a better name, and moves the display logic to CSS.


missytake (approved):
css is magic

PR #1332, opened 2026-04-15
avoid keeping relay, mention hiding relays

r10s:
delay deleting relays may result in unexpectedly deleted messages, if one is not aware that deleting in Delta Chat also deletes on the email address used a a relay. 

directly removing the relay seems to be the option with less friction therefore, esp. as the user sees the message in their classic app - even if they cannot decrypt it.

ftr, it is good that we get rid of all that confusion, by stop saying one shall share an email address with eg. thunderbird and delta ...


hpk42 (approved):
it's at least better than now :)

PR #1333, opened 2026-04-20
Remove setting of `th`/`td` rules as same as browser default

veyndan:


missytake (approved):

veyndan:
@missytake Thanks for approving! Just wondering if it was deliberate that this isn't merged as well? Not sure if we need to hold off on something or not 😄

missytake:
> @missytake Thanks for approving! Just wondering if it was deliberate that this isn't merged as well? Not sure if we need to hold off on something or not 😄

nah, I'm just forgetful every once in a while :sweat_smile:

PR #1336, opened 2026-04-21
Remove styles that are specifically for IE

veyndan:
IE is very out-of-date, and our usage of certain CSS is already incompatible with IE.


missytake (approved):
take that, microsoft

PR #1338, opened 2026-04-22
Rely on browser default for `list-style`

veyndan:
The browser default is slightly different to what we currently have, but I think the browser default is preferable. With nested sub-bullet points (e.g., in the contents of the FAQ), they are unfilled circles. Nested further, they become little squares.

I think this is preferable, as nested items are now distinguishable not just by there indentation, but also by the type of bullet point used.


hpk42 (approved):
lgtm

PR #1339, opened 2026-04-22
Remove unused `ol` styling

veyndan:
These styles are only applied if the class `"p"` is also applied to the `ol`. Since https://github.com/deltachat/deltachat-pages/pull/495, no `ol`'s have had a class `"p"` attached to them, so these styles are completely unused.


hpk42 (approved):
good find.

PR #1340, opened 2026-04-22
Remove styles in `_normalize.scss` if always overridden by `_base.scss`

veyndan:
At the top of our `_base.scss`, we always override these values, so setting them here is pretty useless.


hpk42 (approved):
lgtm

PR #1342, opened 2026-04-28
ci: setup zizmor

link2xt:
[Zizmor](https://docs.zizmor.sh/) is a linter for GitHub Actions

### DIFF
--- a/.github/workflows/jekyll-main.yml
+++ b/.github/workflows/jekyll-main.yml
@@ -8,17 +8,13 @@ on:
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
-# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
-permissions:
-  contents: read
-  pages: write
-  id-token: write
-
 # Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
 # However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
 concurrency:
   group: "pages"
   cancel-in-progress: false
+
+permissions: {}
 
 jobs:
   build:
@@ -26,7 +22,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-      - uses: DeterminateSystems/nix-installer-action@main
+        with:
+          persist-credentials: false
+      - uses: cachix/install-nix-action@2126ae7fc54c9df00dd18f7f18754393182c73cd # v31.9.1
       - name: Build the site with Jekyll
         run: nix build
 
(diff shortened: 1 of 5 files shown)


.github/workflows/jekyll-preview.yml
```diff
@@ -42,18 +54,22 @@ jobs:
...
           export RESPONSE=$(curl -L --header "authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" --url "$GITHUB_API_URL" --header "content-type: application/vnd.github+json" -H "X-GitHub-Api-Version: 2022-11-28")
           echo $RESPONSE > response
           grep -v '"Check out the page preview at https://staging.delta.chat/' response && echo "comment=true" >> $GITHUB_OUTPUT || true
+        env:
+          STEPS_PREPARE_OUTPUTS_PRID: ${{ steps.prepare.outputs.prid }}
       - name: "Post link to comments"
         if: steps.details.outputs.comment
         uses: actions/github-script@v6
+        env:
+          PR_ID: ${{ steps.prepare.outputs.prid }}
```
missytake, line 67: why is it `PR_ID` here and `STEPS_PREPARE_OUTPUTS_PRID` before?

.github/workflows/jekyll-preview.yml
```diff
@@ -2,21 +2,31 @@ name: Deploy Preview
...
     runs-on: ubuntu-latest
 
+    # Permission to post to PR details.
+    permissions:
+      pull-requests: write
+
     steps:
       - uses: actions/checkout@v3
+        with:
+          persist-credentials: false
```
missytake, line 18: TIL 😬 
https://yossarian.net/til/post/actions-checkout-can-leak-github-credentials/

.github/workflows/jekyll-main.yml
```diff
@@ -8,25 +8,23 @@ on:
...
-  id-token: write
-
 # Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
 # However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
 concurrency:
   group: "pages"
   cancel-in-progress: false
 
+permissions: {}
+
```
missytake, line 18: This just explicitly sets the `GITHUB_TOKEN` permissions to empty unless overridden in a later step?

missytake (approved):
Looks good, thanks for caring for this!

.github/workflows/jekyll-main.yml
link2xt: Yes, by default it has way too many permissions, you can use `gh` command to create releases, push commits and whatever. zizmor complains if job does not have permissions set explicitly, so better start with disabling all permissions on workflow level, then give the token necessary permissions on the job level.

link2xt: In this case only `deploy` job needs permissions, `build` job does not need anything. I think actions/upload-pages-artifact should be fine without permissions, if it breaks after merging we will see.

.github/workflows/jekyll-preview.yml
link2xt: `STEPS_PREPARE_OUTPUTS_PRID` is from `zizmor fix`, i think it just named it after the contents. Renamed to `PR_ID` now.

PR #1348, opened 2026-05-13
feat: update and clarify usage of ip addresses

r10s:
this updates the policy according to feedback we got the last months:

- we now have multiple relays
- be explicit about persistance and exposings
- more correct ip address precision
- remove defensive parts distracting from the gist and raising discussions
- replace lockdown mode hint by smth. more actionable

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -980,7 +980,7 @@ can not be identified easily.
 ### Who sees my IP Address?
 
 The used [relays](#relays) need to know your IP Address,
-as well as sometimes your contact's devices if you have a [call](#experiments)
+as well as sometimes your contact's devices if you have a [call](#calls)
 or use [apps](#webxdc) together.
 
 IP Addresses are needed for connectivity and efficiency.
--- a/en/help.md
+++ b/en/help.md
@@ -990,8 +990,8 @@ are not like an address you give to a delivery service,
 but typically less precise, often defining city or region only.
 
 If you see your IP Address as a risk,
-we recommend to use a VPN and block all traffic not using it.
-Per-app options leave gaps across your whole system.
+we recommend to use a VPN for the whole system.
+Per-app options leave gaps across your system.
 For example, tapping a link can expose IP Addresses to unknown parties, which is by far the larger risk.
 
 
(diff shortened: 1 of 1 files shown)


en/help.md
```diff
@@ -864,23 +864,20 @@ can not be identified easily.
 
 ### Who sees my IP Address?
 
-The used [relay](#relays) needs to know your IP Address,
+The used [relays](#relays) need to know your IP Address,
 as well as sometimes your contact's devices if you have a [call](#experiments)
```
adbenitez: I guess this will become outdated after your other PR gets merged so maybe wait for that first and update here?

en/help.md
```diff
@@ -864,23 +864,20 @@ can not be identified easily.
...
-Hunting down options in all apps on your system will leave gaps.
-For example, tapping a link exposes IP Addresses to unknown parties and is the by far larger risk here.
+Delta Chat neither persists nor exposes them.
+Note that IP Addresses
+are not like an address you give to a delivery service,
+but typically less precise, often defining city or region only.
+
+If you see your IP Address as a risk,
+we recommend to use a VPN and block all traffic not using it.
+Per-app options leave gaps across your whole system.
```
adbenitez:
```suggestion
we recommend to use a VPN for the whole system.
Per-app options leave gaps across your system.
```

r10s: comment on your first suggetion: 

the "block all traffic not using it" seems to be a thing, some VPN have explicit options for that. this is strongly related to "kill switch". not sure about the latter, but the precision does not hurt anyway

but your refinement seems to be even clearer

en/help.md
```diff
@@ -864,23 +864,20 @@ can not be identified easily.
...
+For example, tapping a link can expose IP Addresses to unknown parties, which is by far the larger risk.
```
adbenitez, line 995:
```suggestion
For example, every time you open a web link you expose your IP Addresses to potentially untrusted parties, which is by far a larger risk.
```

en/help.md
r10s: i merged #1349, rebased and adapted the link

adbenitez (approved):

PR #1349, opened 2026-05-13
feat: document channels & calls

### LINKED ISSUE #1343: add "channels" to faq
add the basic things about "channels" to the faq, keeping in mind that it should fit the level of detail used elsewhere and that it should be written for average  user, not for nerds

wording in UI is just "channels", not broadcast

### LINKED ISSUE #1344: add "calls" to faq
add the basic things about "calls" to the faq, keeping in mind that it should fit the level of detail used elsewhere and that it should be written for average  user, not for nerds

r10s:
initial documentation of channels and calls.
as usual, this part is meant for the average user, we do on purpose not detail about technical background, but just ui.

closes #1343 
closes #1344

### DIFF
--- a/en/help.md
+++ b/en/help.md
@@ -378,7 +378,7 @@ you can also create a [dedicated profile](#multiple-accounts) for joining a chan
 
 - Tap **New Chat** and choose **New Channel**.
 
-- Enter a **name**, optionally set **image** and **description**, and hit the **Create** button.
+- Enter a **name**, optionally set an **image** and **description**, and hit the **Create** button.
 
 - You can now send and manage messages as usual.
 
--- a/en/help.md
+++ b/en/help.md
@@ -357,8 +357,8 @@ Channels are a one-to-many tool for broadcasting messages.
   or tap the **invite link** you got from the channel owner.
 
 That's all!
-You will now receive all new messages from the channel owner
-and up to 10 older ones.
+You will receive a few of the messages from the channel history
+and, from that point on, all new messages from the channel.
 
 As all of Delta Chat, also Channels are private and decentralized,
 there is no public discovery.
--- a/en/help.md
+++ b/en/help.md
@@ -346,6 +346,116 @@ where Delta Chat is a private messenger for chatting with [equal rights](#groups
 See [Dunbar's number](https://en.wikipedia.org/wiki/Dunbar%27s_number) for more insights.
 
 
+## Channels {#channels}
+
+Channels are a one-to-many tool for broadcasting messages.
+
+
+### Subscribe to a channel
+
+- Scan the <img style="vertical-align:middle; height:1.3em; margin:1px" src="../assets/help/qr-icon.png" /> **QR code**
+  or tap the **invite link** you got from the channel owner.
+
+That's all!
+You will now receive all new messages from the channel owner
+and up to 10 older ones.
+
+As all of Delta Chat, also Channels are private and decentralized,
+there is no public discovery.
+
+Other channel subscribers will not see that you subscribed and cannot message you.
+The channel owner, however, can message you.
+They will also see that you read a message unless you have read receipts disabled.
+
+If you do not want to share your main profile,
+you can also create a [dedicated profile](#multiple-accounts) for joining a channel.
+
+
+### Create a channel
+
+- Tap **New Chat** and choose **New Channel**.
+
+- Enter a **name**, an optional **description**, and hit the **Create** button.
+
+- You can now send and manage messages as usual.
+
+- From the channel's profile, **share the QR code or invite link with others**.
+
+Subscribers will receive your messages,
+but they cannot reply to them.
+When subscribing, they will receive **up to 10 older messages**.
+
+You can see the **view count** beside each message.
+Note that this only counts subscribers who have read receipts enabled,
+so the real view count may be larger.
+
+
+### How many subscribers can a channel have?
+
+Channels are designed for much larger audiences than [groups](#groups).
+
+The practical limit depends on the used [relay](#relays),
+so there is no single fixed number that applies everywhere.
+
+For really large channels with several tens of thousands of subscribers,
+we recommend using a [dedicated profile](#multiple-accounts) for the channel
+and checking whether the relay is suitable.
+
+But don't be too hesitant: Delta Chat is designed to be relay-agnostic,
+so you can change your relay at any point easily -
+your existing subscribers will not even notice.
+You only have to update the invite link for new subscribers in that case.
+
+
+## Calls {#calls}
+
+Delta Chat supports one-to-one **audio calls** and **video calls**.
+
+
+### Place a call
+
+- In a one-to-one chat, tap the 📞 **call icon**.
+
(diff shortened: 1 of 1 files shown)


en/help.md
```diff
@@ -346,6 +346,116 @@ where Delta Chat is a private messenger for chatting with [equal rights](#groups
...
+
+
+### Subscribe to a channel
+
+- Scan the <img style="vertical-align:middle; height:1.3em; margin:1px" src="../assets/help/qr-icon.png" /> **QR code**
+  or tap the **invite link** you got from the channel owner.
+
+That's all!
+You will now receive all new messages from the channel owner
+and up to 10 older ones.
```
adbenitez: probably better to be more vague instead of giving precise numbers:

```suggestion
You will receive a few of the messages from the channel history
and, from that point on, all new messages from the channel.
```

en/help.md
```diff
@@ -346,6 +346,116 @@ where Delta Chat is a private messenger for chatting with [equal rights](#groups
...
 
+## Channels {#channels}
+
+Channels are a one-to-many tool for broadcasting messages.
+
+
+### Subscribe to a channel
+
+- Scan the <img style="vertical-align:middle; height:1.3em; margin:1px" src="../assets/help/qr-icon.png" /> **QR code**
+  or tap the **invite link** you got from the channel owner.
```
adbenitez, line 357: maybe worth mentioning that after clicking the link it might take a while because the channel owner needs to be online? something like "if after you tap a channel invite link it doesn't join you immediately, don't worry, when the channel owner comes online they will process your request to join"

en/help.md
```diff
@@ -346,6 +346,116 @@ where Delta Chat is a private messenger for chatting with [equal rights](#groups
...
+
+If you do not want to share your main profile,
+you can also create a [dedicated profile](#multiple-accounts) for joining a channel.
+
+
+### Create a channel
+
+- Tap **New Chat** and choose **New Channel**.
+
+- Enter a **name**, an optional **description**, and hit the **Create** button.
```
adbenitez: this is not mentioning avatar
```suggestion
- Enter a **name**, optionally set **image** and **description**, and hit the **Create** button.
```

en/help.md
r10s: yeah, makes sense. that probably changes faster than the help can be updated (as usual :)

en/help.md
```diff
@@ -346,6 +346,116 @@ where Delta Chat is a private messenger for chatting with [equal rights](#groups
...
+
+- You can now send and manage messages as usual.
+
+- From the channel's profile, **share the QR code or invite link with others**.
+
+Subscribers will receive your messages,
+but they cannot reply to them.
```
adbenitez: advertise the feature rather than making it look like a disadvantage, and technically they can reply to them using the "reply privately"
```suggestion
Subscribers will receive your messages,
but they cannot send messages in your channel.
```

en/help.md
```diff
@@ -346,6 +346,119 @@ where Delta Chat is a private messenger for chatting with [equal rights](#groups
...
+
+- Enter a **name**, optionally set **image** and **description**, and hit the **Create** button.
+
+- You can now send and manage messages as usual.
+
+- From the channel's profile, **share the QR code or invite link with others**.
+
+Subscribers will receive your messages,
+but they cannot reply to them.
+When subscribing, they will receive **up to 10 older messages**.
```
adbenitez: here also not sure it is good to give precise numbers, maybe better just say "a few of the latest messages of the channel history"

en/help.md
r10s: makes sense, i added a commit. the channel additionally says "waiting for the device of OWNER to replay", but it seems fine to repeat that point, as this is a huge difference to centralised services

en/help.md
```diff
@@ -346,6 +346,119 @@ where Delta Chat is a private messenger for chatting with [equal rights](#groups
...
+so there is no single fixed number that applies everywhere.
+
+For really large channels with several tens of thousands of subscribers,
+we recommend using a [dedicated profile](#multiple-accounts) for the channel
+and checking whether the relay is suitable.
+
+But don't be too hesitant: Delta Chat is designed to be relay-agnostic,
+so you can change your relay at any point easily -
+your existing subscribers will not even notice.
+You only have to update the invite link for new subscribers in that case.
```
adbenitez:
```suggestion
You only have to update the invite link you share with new subscribers in that case.
```

en/help.md
```diff
@@ -346,6 +346,119 @@ where Delta Chat is a private messenger for chatting with [equal rights](#groups
...
+
+But don't be too hesitant: Delta Chat is designed to be relay-agnostic,
+so you can change your relay at any point easily -
+your existing subscribers will not even notice.
+You only have to update the invite link you share with new subscribers in that case.
+
+
+## Calls {#calls}
+
+Delta Chat supports one-to-one **audio calls** and **video calls**.
```
adbenitez, line 415: in the case of Android we only support on android +8 IIRC (needs double checking) would be nice to mention this somewhere in this section so people can troubleshoot why calls don't work for them

en/help.md
```diff
@@ -346,6 +346,119 @@ where Delta Chat is a private messenger for chatting with [equal rights](#groups
...
+- If you do not answer, do not hear the ringing, or do not have your device at hand,
+  the call appears as a **missed call**.
+
+- **Only your accepted contacts** can make your device ring.
+  Contact requests will appear as usual and will not ring.
+
+- At **Settings → Notifications → Calls**,
+  you can disable the special call ringing screen completely.
+  If you do so, you only get notifications as for all other messages -
+  unless you've disabled them as well.
```
adbenitez: at least currently this is not true, you don't get any notification, which is also fine IMHO

```suggestion
  If you do so, you will not be disturbed by any ringing notification,
  you can still pick up the call by tapping the incoming call message bubble in its chat.
```

mmmh after drafting the suggestion, I think in iOS and maybe even desktop it is still not possible to pick the call that way??? it should tho

en/help.md
r10s: i added a line, assuming android 8

en/help.md
r10s: ah - i though you get some. 

anyway, the description would match both cases - main point is anyway to disable ringing

adbenitez (approved):

en/help.md
r10s: > I think in iOS and maybe even desktop it is still not possible to pick the call that way??? it should tho

yip, there are issues/pr, but that is no blocker for the release, the help might only be a little ahead in time then :)

tbh, the option is anyway only for few - i assume most are happy with the default, that your accepted contacts can ring you

en/help.md
adbenitez: I checked and it seems android 8 is the minimum so that is fine

en/help.md
r10s: github is broken somehow - it says the suggestion is applied, but it  actually is not

r10s: now it worked - we're maybe to quick working on the PR with two ppl :)

PR #1352, opened 2026-05-16
Localize table related rules

veyndan:
Essentially just moves the `<table>` related rules from the normalise section of the stylesheet to where other `<table>` related rules are.

Some of the `main table` selectors are now just `table`. No meaningful difference as the only `table`s that exist on the site are in a `main` anyway. IMO it was needlessly specific previously.

### DIFF
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -2,7 +2,6 @@
 *******************************************************************************/
 
 * {margin:0;padding:0}
-table {border-collapse:collapse} th {text-align:left;}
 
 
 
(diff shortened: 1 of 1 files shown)


link2xt:
Something went wrong? Not sure if it deployed the PR branch or main.

veyndan:
This is one of the worst PRs to figure that out as it's just moving around CSS rules so there shouldn't be any visible changes 😂 

The page preview is correct though, it's perhaps just a got a wrong branch name in it. I just validated it by diffing https://delta.chat/assets/css/styles.css and https://staging.delta.chat/main/assets/css/styles.css, and the latter does have the edits present in this PR.

missytake:
Nice, it was deployed to https://staging.delta.chat/1352/en/ successfully, #1367 seems to have worked. Only problem is that approved workflows can't post to the comments yet.

missytake (approved):

PR #1355, opened 2026-05-18
job offering is outdated

r10s:
remove job offering, this paragraph is outdated

PR #1356, opened 2026-05-18
Update contribute.md to remove job posting

Hocuri:
We're only looking for a Rust developer currently


Hocuri:
Race condition with https://github.com/deltachat/deltachat-pages/pull/1355

PR #1358, opened 2026-05-20
ci: allow forks to upload to staging.delta.chat after approval

missytake:
I followed this guide for this PR: https://dvc.org/blog/testing-external-contributions-using-github-actions-secrets/

The idea is, that when an external contributor (e.g. @veyndan) opens a PR from a fork:
- it will be assigned to the `approval_needed` environment, which has no environment secrets, but needs an approval by one of five people: https://github.com/deltachat/deltachat-pages/settings/environments/15576404735/edit
- if it gets that approval, it can use the build job, which has access to the staging.delta.chat environment, which has the deployment key.

PRs from branches within the repository (allowed for organization members) will directly get access to the staging.delta.chat environment, so they don't need approval.

I also used this opportunity to remove the TRANSIFEXRC secret, it's not used anyway.


.github/workflows/jekyll-preview.yml
```diff
@@ -1,11 +1,22 @@
 name: Deploy Preview
 
-on: [pull_request]
+on:
+  pull_request_target:  # zizmor: ignore[dangerous-triggers]
+    branches: main
+    types: [opened, synchronize, reopened]
```
link2xt: This is the default according to https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target

"By default, a workflow only runs when a pull_request_target event's activity type is opened, synchronize, or reopened."

link2xt (approved):

PR #1371, opened 2026-06-12
ci: use github.event.pull_request.number for URLs

missytake:
This should result in a PR ID in the URL, unfortunately we can only test it when it's on main.

### DIFF
--- a/.github/workflows/jekyll-preview.yml
+++ b/.github/workflows/jekyll-preview.yml
@@ -50,14 +50,14 @@ jobs:
         id: details
         run: |
           # URLs for API connection and uploads
-          export GITHUB_API_URL="https://api.github.com/repos/deltachat/deltachat-pages/statuses/${{ github.event.after }}"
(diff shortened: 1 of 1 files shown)


veyndan:
Do these previews above reflect the changes in this PR? If so, it seems like the new parameter that's used isn't set or is empty.

missytake:
> Do these previews above reflect the changes in this PR? If so, it seems like the new parameter that's used isn't set or is empty.

afaict the problem is that for `pull_request_target` workflows always the version on `main` is used: https://github.blog/changelog/2025-11-07-actions-pull_request_target-and-environment-branch-protections-changes/

.github/workflows/jekyll-preview.yml
hpk42: doesn't this also need to be changed?

missytake: oops, forgot one :see_no_evil:

PR #1375, opened 2026-06-16
fix: allow actions to post to commit status 

hpk42:
which should help prevent job failures for posting previews


missytake (approved):

missytake:
Let's see on main whether this works.

PR #1380, opened 2026-06-24
feat: add link to gentoo overlay

nicodh:
see https://github.com/deltachat/deltachat-desktop/issues/6503

We should add that also to  https://chatmail.at/clients


feld:
thank you! I will remember this repo in the future

WofWca (approved):
I don't know how trusted that repo is, but I guess we already have several list items marked "community-maintained", and in that this MR is in line.

PR #1382, opened 2026-07-07
ci: rrsync prepends the restricted upload path, we need to leave it out

missytake:
This is necessary because https://github.com/deltachat/sysadmin/issues/270 switches our server-side authentication to rrsync, and that uploads to a restricted path already; keeping it like this will result in a destination path like `/var/www/html/var/www/html`.

We need to re-run CI and merge after migrating page to www.

### DIFF
--- a/.github/workflows/jekyll-main.yml
+++ b/.github/workflows/jekyll-main.yml
@@ -35,7 +35,7 @@ jobs:
           mkdir -p "$HOME/.ssh"
           echo "${{ secrets.KEY }}" > "$HOME/.ssh/key"
           chmod 600 "$HOME/.ssh/key"
-          rsync -rILvh -e "ssh -i $HOME/.ssh/key -o StrictHostKeyChecking=no" $GITHUB_WORKSPACE/result/ "${{ secrets.USERNAME }}@delta.chat:/var/www/html/_site"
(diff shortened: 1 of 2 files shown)


.github/workflows/jekyll-main.yml
```diff
@@ -35,7 +35,7 @@ jobs:
           mkdir -p "$HOME/.ssh"
           echo "${{ secrets.KEY }}" > "$HOME/.ssh/key"
           chmod 600 "$HOME/.ssh/key"
-          rsync -rILvh -e "ssh -i $HOME/.ssh/key -o StrictHostKeyChecking=no" $GITHUB_WORKSPACE/result/ "${{ secrets.USERNAME }}@delta.chat:/var/www/html/_site"
+          rsync -rILvh -e "ssh -i $HOME/.ssh/key -o StrictHostKeyChecking=no" $GITHUB_WORKSPACE/result/* "${{ secrets.USERNAME }}@delta.chat:"
```
j4n:
```suggestion
          rsync -rILvh -e "ssh -i $HOME/.ssh/key -o StrictHostKeyChecking=no" $GITHUB_WORKSPACE/result/ "${{ secrets.USERNAME }}@delta.chat:"
```
just adding the trailing slash prevents creating a directory with rsync.

j4n (approved):
similarly, lgtm

PR #1389, opened 2026-07-27
update android 2.57.0

adbenitez:


j4n (approved):
from what I understand, looks good.

PR #1392, opened 2026-08-02
Don't suggest hiding relays in legacy-move.md anymore

lk108:
Hiding a relay is not offered anymore, so the corresponding text in `legacy-move.md` should be adapted.

The changes in this PR currently just remove the respective section, but maybe it should be kept and updated to say that hiding is offered for versions <= 2.53.0 (?) only, and later versions will take care of it automatically?

### DIFF
--- a/en/legacy-move.md
+++ b/en/legacy-move.md
@@ -58,10 +58,6 @@ If so, you should change the profile as follows:
 
 7. **Message your contacts;** their Delta Chat will learn the new relay automatically.
 
-In step 6. you are also offered to _hide instead of removing._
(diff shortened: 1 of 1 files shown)


lk108:
Maybe the warning `be prepared that **removing messages in Delta Chat will remove them on the email server.**` still applies and should remain?

r10s (approved):
it is good enough. we have a earning about "no shared usage" since some time on configuring profiles, and for older profiles, that warning was long issued. so, all in all the page gets kess important over time anyways.

thanks for taking care!

PR #1395, opened 2026-08-11
ci: update checkout action

missytake:
The allow-unsafe-pr-checkout input was introduced in https://github.com/actions/checkout/pull/2454, which is fine, because since https://github.com/deltachat/deltachat-pages/pull/1358, such a PR needs to be approved manually anyway.


j4n (approved):
lgtm

PR #1396, opened 2026-08-13
update android to 2.58.0

adbenitez:


missytake (approved):
If it's rolling on gplay, we can also publish it here I'd say.

PR #1398, opened 2026-08-18
Add farsi

r10s:
this PR adds "farsi" language. as this is the first rtl language, some adaptions were needed:

- add `dir`-attribute to <html> tag and to farsi item of the language selection footer

- on the homepage, remove wird div/block/aling=left in favour to a simple, default floating layout; re-pull translations as inline-html goes over transifex

- some CSS tweaks are needed, we do that in a subsequent PR as otherwise this will be hard to review in this large PR

ftr, sitting with @adbenitez here


missytake:
image

This looks very funny
(good enough for now I think?)

r10s:
> This looks very funny
> (good enough for now I think?)

there are some of these glitches, mostly related to CSS/HTML doing by far too much. i fixed the homepage.

for the blog at hand: the general layout with images on the right seems totally fine, that could even be a layout decision on an english page. what is weird, are the full stop signs as the exclamation marks.

but let's merge this PR as is, and improve CSS a bit in a subsequent PR

r10s:
some more rtl fixes at https://github.com/deltachat/deltachat-pages/pull/1399

PR #1393, opened 2026-08-08
Add full Persian (fa) website localization with RTL support

abbasdp:
Introduce Persian language pages, UI strings, and translated blog posts so Farsi speakers can use the site end-to-end. Wire fa into language selection and redirects, enable RTL layout and nav alignment for Persian only, and add a fa_posts collection for localized blog content.

PR #1400, opened 2026-08-18
Fix Persian footer label and include fa in local help

abbasdp:
Small follow-up after #1398 / https://delta.chat/fa

### Changes
- Footer language label: Arabic «الفارسية» → Persian «فارسی»
- Add `fa` to `tools/create-local-help.py` so offline/in-app help can include Persian
- Add missing space after the lock emoji on the Persian homepage (also pushed to Transifex)

### Notes
- `redirect.js` already includes `fa` on current `main`
- Kept the existing `dir="rtl"` on the footer link
- Tiny, reviewable PR per the “less complexity” guidance in #1393

### DIFF
--- a/_includes/footer-languages.html
+++ b/_includes/footer-languages.html
@@ -18,4 +18,4 @@
 · <a href="../ru/">Pусский</a>
 · <a href="../uk/">Українська</a>
 · <a href="../zh_CN/">简体中文</a>
-· <a href="../fa/" dir="rtl">الفارسية</a>
+· <a href="../fa/" dir="rtl">فارسی</a>
--- a/fa/index.md
+++ b/fa/index.md
@@ -11,7 +11,7 @@ lang: fa
 
 🥳 [مینی برنامه‌های تعاملی تحت وب](https://webxdc.org/) در چت برای بازی کردن و همکاری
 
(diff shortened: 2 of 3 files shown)


abbasdp:
Follow-up to #1398 / discussion in #1393.

Fixes the Arabic footer label «الفارسية» → Persian «فارسی», and includes `fa` in offline help generation.

r10s:
- the footer is already fixed at https://github.com/deltachat/deltachat-pages/pull/1399: EDIT: ah, it is about create-local-help.sh, yeah, let's get that in, tho that may require additional changes in the apps as well
- the string in index.html needs to be fixed on transifex: EDIT: ah, you already did, so that does not matter here
- the change for the language name is fine

abbasdp:
Thanks a lot for reviewing and merging, @r10s — and for fixing the other minor glitches as well ❤️

Really appreciate the guidance through the Transifex workflow and the RTL approach. Happy to help with any follow-up polish if useful.

r10s:
the work on translations is really a progress that always needs maintainance - we will probably adapt our homepage at some point, then new translations are required. same for new features in the app etc. help on that is really welcome!

you should be able to get a notification from transifex, once this happens - or just check the state every some weeks. esp. for the app one will anyway notice if some strings are suddenly english (again)

PR #1401, opened 2026-08-18
fix indentation of lists in rtl-mode by using browser defaults

r10s:
problem is that the indentation of lists is broken for rtl texts.

reasons is that we overwrite all browser default margins and then add our owns using `-left` - this fails for rtl languages.

instead of adding additional rules for rtl, this PR removes overwriting all browser defaults.

the lists are a bit more condensed esp. on the help pages, but that's totally fine, one can regard this even as an improvement.

reviewers: esp. check the english/non-rtl homepage if there are regressions.

successor of https://github.com/deltachat/deltachat-pages/pull/1398

### DIFF
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -1,12 +1,11 @@
-/* reset
-*******************************************************************************/
-
-* {margin:0;padding:0}
-
-
 
 /* rough layout: mobile first 
 *******************************************************************************/
+
+body, nav ul, nav ul li {
+    margin: 0;
+    padding: 0;
+}
 
 nav {
     text-align: center;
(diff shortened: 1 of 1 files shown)


missytake (approved):
Doesn't seem to solve the list indentation in #1400, but a good start :)

r10s:
> Doesn't seem to solve the list indentation in https://github.com/deltachat/deltachat-pages/pull/1400

where is a list indentation issue mentioned in #1400 ?

missytake:
> > Doesn't seem to solve the list indentation in #1400
> 
> where is a list indentation issue mentioned in #1400 ?

I just meant the missing space in the fa/index.md list item.

r10s:
k, however, that was just a typo on transifex :)

PR #1402, opened 2026-08-18
remove cosmos.delta.chat footer

r10s:
cosmos.delta.chat serves as an overview for devs and nerds about pages related to delta chat, and can stay as that.

however, a permanent footer is over the top meanwhile,
as the page addresses more and more non-nerds.
also, the linked pages partly also stopped using the footer or a little outdated.

by the removal, more weight is given to the main menu, which is more important for most user


link2xt (approved):
I also just checked, https://delta.chat/en/contribute has a link to cosmos already, somewhat in a place where i'd expect it to be.

PR #1403, opened 2026-08-18
Sync Persian footer: حمایت instead of اعانه

abbasdp:
### Problem
On https://delta.chat/fa the footer donate label was «اعانه», which is awkward/wrong in this context, and it was out of sync with the donate page heading («اهدای پول»). References also used «مرجع‌ها» on the page vs «مراجع» in the footer.

### Fix
- `donate` menu/footer string: **حمایت**
- Donate page title/heading/body: use **حمایت** consistently
- References page title: **مراجع** (matches footer)

Also updated the same strings on Transifex so the next translation pull stays consistent.

### DIFF
--- a/_data/lang/fa.yml
+++ b/_data/lang/fa.yml
@@ -16,7 +16,7 @@ fa:
   subscribe_to_feed: مشترک شدن در خوراک ما
 
   references: مراجع
-  donate: اعانه
+  donate: حمایت
   imprint: اطلاعات حقوقی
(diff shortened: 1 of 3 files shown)


r10s (approved):

r10s:
thanks a lot! ftr, if it is not urgent or a larger change, it is also sufficient to change only on transifex, we pull that regulary (faster if you ping me :)

PR #1404, opened 2026-08-18
Fix footer social icons alignment on RTL pages

abbasdp:
### Problem
On RTL pages (e.g. https://delta.chat/fa), the footer `.socials` row (RSS / Mastodon) gets flex-reversed, so icon/label order looks wrong.

### Fix
Minimal CSS only (no extra RTL rule soup):
- `direction: ltr` + `unicode-bidi: isolate` on `.socials` (labels are Latin brand names)
- center/align tweaks and a bit of vertical margin so it sits cleanly under the language list

Follows the “less complexity” guidance from #1393 / #1398.

### DIFF
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -201,16 +201,25 @@ table {
 }
 
 .socials {
+    /* Brand labels are Latin (RSS/Mastodon); keep icon→label order even on RTL pages */
     display: inline-flex;
+    direction: ltr;
+    unicode-bidi: isolate;
+    justify-content: center;
+    align-items: center;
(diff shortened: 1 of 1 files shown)


abbasdp:
Thanks a lot for the reviews and merges, @r10s ❤️

From my side everything looks good on https://delta.chat/fa now — footer labels, language name, and the overall layout.

About further RTL/right-align work: it may not be needed. With the current setup (especially after removing the hard left floats), nothing looks broken to me. Happy to leave CSS as-is unless you spot something specific.

r10s (approved):

PR #1411, opened 2026-08-23
update desktop version

missytake:
I uploaded the files here: https://download.delta.chat/desktop/v2.59.1/

[The github release](https://github.com/deltachat/deltachat-desktop/releases/tag/v2.59.1) did not contain the experimental/ and signatures/ directories... those are probably only generated in the local build process. Is it still fine? The instructions in https://delta.chat/en/verify-downloads can still be followed through :)


WofWca:
The lack of `experimental` is expected, see https://github.com/deltachat/deltachat-desktop/pull/6656. Regarding `signatures`: I don't think we ever uploaded such a dir to the GitHub release. Seems like the process is different there.

WofWca (approved):
OK, I followed the signature verification process for v2.59.1 and it works. Up until https://github.com/deltachat/deltachat-desktop/issues/6030 (https://github.com/deltachat/deltachat-pages/pull/1265) we used to recommend to use those per-file signatures that are missing instead of `signed-checksums.txt`.
It's not nice that they are missing just for this release, but it's not too sussy since the "new" verification method is still available.

Thanks for taking care of this!

PR #1412, opened 2026-08-25
improve landing page

adbenitez:
before-after


j4n (approved):
Great, definitively a step in the right direction :)

missytake:
Yeah, I like it as well - only downside I see is that there is no button under the text anymore. So if the text intrigues you, there is no immediate call to action.

assets/css/styles.scss
```diff
@@ -17,3 +17,15 @@
...
+.download-btn, .download-btn:hover, .download-btn:visited {
+	border-radius: 5px;
+	padding: 10px;
+	display: inline-block;
+	margin: 10px 0;
+
+	background: #364e59;
+	color: white !important;
+	font-weight: bold;
+    text-decoration: none;
```
WofWca: Formatting is off

```suggestion
	text-decoration: none;
```

WofWca (approved):
Looks much better to me.

Related forum post: https://support.delta.chat/t/website-overhaul-should-drop-outdated-look/5174?u=wofwca.

en/index.md
```diff
@@ -3,7 +3,20 @@ title: Delta Chat, decentralized secure messenger
 lang: en
 ---
 
-# Delta Chat is a decentralized and secure messenger app {#homepage-heading}
+# Decentralized & Secure Messaging {#homepage-heading}
+
+<div style="text-align: center">
+<a href="https://get.delta.chat">
```
WofWca: Though IDK about linking to here. I'd rather expect the image to enlarge, since it's pretty hard to make things out on this significantly downscaled image (I know it's like this on current main).

```suggestion
<a href="../assets/home/screenshots/desktop-and-mobile.png">
```

WofWca (commented):
I also tested on mobile.

And - are the old images not supposed to be removed by this MR? Are they still referenced?

en/index.md
adbenitez: nor WhatsApp, Signal or Telegram pages allow to enlarge the screenshots thumbnails, it is just to get some impression, no need to show much details there, user can always try the app, and that screenshots are also a bit bad so better not to let look closely XD, opening an image directly in browser (which might trigger download in some browsers) instead of some "image viewer" is also not nice, so I would say better not to do that, people can always pan the page to zoom a bit

adbenitez:
> And - are the old images not supposed to be removed by this MR? Are they still referenced?

for now I have left them because at least the desktop screenshots I couldn't find in the interface repo, the others could probably be removed, probably we should just move the desktop screenshot to interface repo

there are also super old images there besides that ones, so if we want to clean up there probably could be done in a separated PR

en/index.md
WofWca: Then it's better not to have this as a link at all. It's not intuitive that clicking a screenshot leads to the "download" page. For that we have the "download" button.

adbenitez: @WofWca ok, removed the link for now, thanks for the review!

WofWca (approved):
Still looks good.

PR #1418, opened 2026-09-15
fix: add language redirect for verify-downloads

missytake:
http://delta.chat/verify-downloads gives a 404 right now, should be a language redirect


r10s (approved):
just out of curiosity: is that linked used somewhere? in general, there are probably more pages without shortcut, as we add them only as needed.

but yeah, you seem to need it :)

missytake:
> just out of curiosity: is that linked used somewhere?
> 
> but yeah, you seem to need it :)

it's linked to in our non-public release docs, where I just changed the link to include `en` - so not really needed, true. I'll close this PR.

> in general, there are probably more pages without shortcut, as we add them only as needed.

But good to know, I'll continue to open PRs like this every time someone reports a 404 :D

